PR #5, opened 2025-09-15
Feature/vdbtensor removal

blackencino:
WIP:

Based on all the many conversations, here's the simplest (smallest amount of change) update for the NN modules vdbtensor removal, and also modification to support single-grids.

### DIFF
--- a/fvdb/nn/simple_unet.py
+++ b/fvdb/nn/simple_unet.py
@@ -22,8 +22,7 @@ Components:
 - SimpleUNetBasicBlock: Basic convolution-batchnorm-activation block
 - SimpleUNetConvBlock: Multi-layer residual block
 - SimpleUNetDown/Up: Resolution changing operations with channel adjustment
-- SimpleUNetSkipCat: Skip connection concatenation and channel fusion
-- SimpleUNetPad/Depad: Boundary handling for sparse convolutions
+- SimpleUNetPad/Unpad: Boundary handling for sparse convolutions
 - SimpleUNetDownUp: Recursive encoder-decoder structure
 - SimpleUNet: Main network combining all components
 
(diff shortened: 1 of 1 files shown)


fvdb/nn/modules.py
```diff
@@ -26,110 +31,230 @@ def _forward(self, *args, **kwargs):
...
+
+        # TODO(chorvath): If this is desired behavior, build into GridBatch directly.
+        new_coarse_data.jdata[torch.isinf(new_coarse_data.jdata)] = 0.0
+
+        return new_coarse_data, new_coarse_grid
 
 
 @fvnn_module
-class UpsamplingNearest(nn.Module):
+class MaxPoolSingle(_MaxPoolBase):
```
fwilliams: I think the modules should only operate batchwise (we only train networks with batches). This is similar to how Pytorch works.

fvdb/nn/modules.py
```diff
@@ -522,15 +893,25 @@ def convert_sync_batchnorm(cls, module: nn.Module, process_group: Any = None) ->
...
+# SyncBatchNorm is meaningless for batch size == 1 (Grid instead of GridBatch)
+
+
+# ------------------------------------------------------------------------------------------------
+
 # Non-linear Activations
 
 
 @fvnn_module
 class ElementwiseMixin:
```
fwilliams, line 384: keep for now, and let's get rid of these when we make the __torch_function__ changes

fvdb/nn/modules.py
```diff
@@ -618,3 +999,280 @@ class Dropout(ElementwiseMixin, nn.Dropout):
...
+# Sequential
+# torch.nn.Sequential is designed around layers which take a single input and produce a single output.
+# fvdb layers always take a JaggedTensor as the first argument, but sometimes take a GridBatch
+# as the second argument, and sometimes a GridBatch as the third argument.
+# Sometimes they return just a JaggedTensor, sometimes a tuple of (JaggedTensor, GridBatch),
+# and sometimes a tuple of (JaggedTensor, GridBatch, SparseConvPackInfo).
+
+
+@fvnn_module
+class Sequential(nn.Module):
```
fwilliams, line 487: I really feel we shouldn't roll our own Sequential here. Just make the user call the modules one after the other. This is a potential footgun since we don't always return grid/tensor pairs and the user now needs to keep track of that.

fvdb/nn/modules.py
```diff
@@ -249,22 +274,28 @@ def __init__(
...
         self.weight.data.uniform_(-std, std)
         if self.bias is not None:
             self.bias.data.uniform_(-std, std)
 
-    def _dispatch_conv(self, in_feature, in_grid, in_kmap, out_grid):
+    def _dispatch_conv(
+        self,
+        in_feature: JaggedTensor,
+        in_grid: GridBatch,
+        in_kmap: SparseConvPackInfo | None,
```
fwilliams: confirming the plan is to get rid of `SparseConvPackInfo` from the public API?

fvdb/convolution_plan.py
```diff
@@ -0,0 +1,777 @@
...
+    _pack_info: SparseConvPackInfoCpp
+    _channel_pairs: tuple[tuple[int, int], ...]
+    _transposed: bool
+    _expert_config: dict[str, Any]
+    _backend: ConvPackBackend
+
+    @classmethod
+    def from_grid_batch(
+        cls,
+        channel_pairs: tuple[tuple[int, int], ...],
```
fwilliams: Do we need to provide the channel pairs by default? Some backends constrain this and some backends don't have any constraints. Can we cache packinfo at runtime for different channel pairs in the plan?

fvdb/nn/modules.py
blackencino: yep.

fvdb/convolution_plan.py
blackencino: When you create a Conv layer in torch - of any variety - you have to specify the in and out channels. Since our backends are sensitive to channel count configurations, we need this information to do validation, otherwise we have the problem of "init succeeds, apply fails", which is super confusing. I think given the fact that the user pattern is already "channel counts known on construction", this pattern is not out of line with that.

blackencino: I fixed this so that by default all channel pairs are accepted. Selecting a specific backend will still require choosing the channel pairs that are limited to that backend, as it was before.

fvdb/nn/modules.py
blackencino: This is gone now. There's only ConvolutionPlan.

blackencino:
Superceded in other PR.

PR #14, opened 2025-09-16
Minimal conda environment with cuda

fwilliams:
This PR adds a minimal conda environment with system packages for cuda and compiler tools which can be used as a starting point for building fvdb (pip install -r env/build_requirements.txt). This acts as a good transition as we move over to pip-only for the release.


swahtz:
Not needed anymore.

PR #19, opened 2025-09-17
Fix PLY loading when there are no shN coefficients

fwilliams:
When a PLY file has no shN coefficients, we used to throw an exception. This case is now gracefully handles and we simply load an empty tensor.

### DIFF
--- a/src/fvdb/detail/io/GaussianPlyIO.cpp
+++ b/src/fvdb/detail/io/GaussianPlyIO.cpp
@@ -233,9 +233,6 @@ loadGaussianPly(const std::string &filename, torch::Device device) {
         shNData.reset();
     }
 
-    std::shared_ptr<PlyData> normalizationTransformData, cameraToWorldMatricesData,
-        projectionTypesData, projectionParametersData;
-
     // Read out metadata from comment strings
     auto [retMetadata, retTensorMetadata] = parsePlyMetadataComments(plyf);
 


src/fvdb/detail/io/GaussianPlyIO.cpp
```diff
@@ -222,8 +222,14 @@ loadGaussianPly(const std::string &filename, torch::Device device) {
     // Request SH coefficient data
     std::shared_ptr<PlyData> sh0Data =
```
blackencino, line 225: use auto with smart_ptrs whenever possible!

src/fvdb/detail/io/GaussianPlyIO.cpp
```diff
@@ -222,8 +222,14 @@ loadGaussianPly(const std::string &filename, torch::Device device) {
...
         plyf.request_properties_from_element("vertex", sh0PlyPropertyNames);
-    std::shared_ptr<PlyData> shNData =
-        plyf.request_properties_from_element("vertex", shNPlyPropertyNames);
+    std::shared_ptr<PlyData> shNData;
+    try {
+        shNData = plyf.request_properties_from_element("vertex", shNPlyPropertyNames);
+    } catch (std::invalid_argument &e) {
+        // If there are no SH N coefficients, tinyply will throw an exception. We can ignore this
+        // and just set shNData to nullptr
+        shNData = nullptr;
```
blackencino: shNData.reset() is the "blessed" way for smartptrs.

src/fvdb/detail/io/GaussianPlyIO.cpp
```diff
@@ -222,8 +222,14 @@ loadGaussianPly(const std::string &filename, torch::Device device) {
     // Request SH coefficient data
     std::shared_ptr<PlyData> sh0Data =
         plyf.request_properties_from_element("vertex", sh0PlyPropertyNames);
-    std::shared_ptr<PlyData> shNData =
-        plyf.request_properties_from_element("vertex", shNPlyPropertyNames);
+    std::shared_ptr<PlyData> shNData;
+    try {
+        shNData = plyf.request_properties_from_element("vertex", shNPlyPropertyNames);
+    } catch (std::invalid_argument &e) {
```
blackencino: do you need to be specific about which type of exception you catch? I'd just use std::exception, or even just ..., unless it matters to you that you only catch this one thing but allow everything else through. It is brittle for no reason.

PR #144, opened 2025-09-22
Add support for mGPU fused SSIM

matthewdcong:
This PR adds mGPU support for fused SSIM.

1. Adds the previously vendored single GPU fused SSIM kernels into the codebase. This enables the addition of the mGPU + Unified Memory variant, and fixes a small issue in the single GPU variant where the stream was not specified. The existing copyright of the vendored code is preserved with the OpenVDB copyright appended afterwards.
2. Defines the custom PyTorch operators `_fused_ssim` and `_fused_ssim_backward` and their appropriate schemas. Note that the names here are chosen (leading underscore, no trailing underscore) to be consistent with existing PyTorch fused operator schemas.
3. Provides/binds the aforementioned fused SSIM kernels to these operators for both the CUDA and PrivateUse1 backend.

As a result, PyTorch will dynamically dispatch to the CUDA or PrivateUse1 version of the operator (just like any other PyTorch operator) based on the selected devices. Furthermore, all the tests in test_metrics.py pass without modification.

### DIFF
--- a/src/fvdb/detail/ops/gsplat/FusedSSIM.cu
+++ b/src/fvdb/detail/ops/gsplat/FusedSSIM.cu
@@ -477,7 +477,8 @@ fusedSSIMBackwardKernel(int localToGlobalOffset,
 //   If train=false, derivative Tensors are empty.
 // ------------------------------------------
 std::tuple<torch::Tensor, torch::Tensor, torch::Tensor, torch::Tensor>
-fusedSSIMCUDA(double C1, double C2, torch::Tensor &img1, torch::Tensor &img2, bool train) {
+fusedSSIMCUDA(
+    double C1, double C2, const torch::Tensor &img1, const torch::Tensor &img2, bool train) {
     const at::cuda::OptionalCUDAGuard device_guard(device_of(img1));
     const auto stream = at::cuda::getCurrentCUDAStream(img1.device().index());
     int B             = img1.size(0);
@@ -509,8 +510,8 @@ fusedSSIMCUDA(double C1, double C2, torch::Tensor &img1, torch::Tensor &img2, bo
                                                 CH,
                                                 static_cast<float>(C1),
                                                 static_cast<float>(C2),
-                                                img1.contiguous().data_ptr<float>(),
-                                                img2.contiguous().data_ptr<float>(),
+                                                img1.contiguous().const_data_ptr<float>(),
+                                                img2.contiguous().const_data_ptr<float>(),
                                                 ssim_map.data_ptr<float>(),
                                                 train ? dm_dmu1.data_ptr<float>() : nullptr,
                                                 train ? dm_dsigma1_sq.data_ptr<float>() : nullptr,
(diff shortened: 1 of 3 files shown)


src/fvdb/detail/ops/gsplat/FusedSSIM.cu
```diff
@@ -0,0 +1,663 @@
...
+
+// For partial results after horizontal pass
+#define CONV_X BLOCK_X
+#define CONV_Y SHARED_Y
+
+// ------------------------------------------
+// Utility: Safe pixel fetch w/ zero padding
+// ------------------------------------------
+__device__ __forceinline__ float
+get_pix_value(const float *img, int b, int c, int y, int x, int CH, int H, int W) {
```
fwilliams: camelCase for consistency with rest of library

src/fvdb/detail/ops/gsplat/FusedSSIM.cu
```diff
@@ -0,0 +1,663 @@
...
+//    to dm_dmu1, dm_dsigma1_sq, dm_dsigma12
+// ------------------------------------------
+__global__ void
+fusedssimCUDA(int offset,
+              int H,
+              int W,
+              int CH,
+              float C1,
+              float C2,
+              const float *__restrict__ img1,
```
fwilliams: not for this PR, but open an issue to use tensor accessors here instead of pointers.

src/fvdb/detail/ops/gsplat/FusedSSIM.cu
```diff
@@ -0,0 +1,663 @@
...
+    int CH            = img1.size(1);
+    int H             = img1.size(2);
+    int W             = img1.size(3);
+
+    auto dL_dimg1 = torch::zeros_like(img1);
+
+    dim3 grid((W + BLOCK_X - 1) / BLOCK_X, (H + BLOCK_Y - 1) / BLOCK_Y, B);
+    dim3 block(BLOCK_X, BLOCK_Y);
+
+    fusedssim_backwardCUDA<<<grid, block, 0, stream>>>(0,
```
fwilliams: We should TORCH_CHECK_VALUE for float32 images and throw a meaningful error when this fails. data_ptr<float>() will throw an error but the message is something arcane.

We may also want to TORCH_CHECK_VALUE when the input images are not contiguous rather than just calling .contiguous().

matthewdcong: Done, thanks!

src/fvdb/detail/ops/gsplat/FusedSSIM.cu
matthewdcong: Fixed

src/fvdb/detail/ops/gsplat/FusedSSIM.cu
matthewdcong: Filed #147

PR #152, opened 2025-09-23
Docker workdir fixes

blackencino:
This change modifies the docker build so that it no longer uses the parent directory, which is a holdover from the 
previous OpenVDB repo.

It also uses a more tailored approach towards the mounting of local directories.

Builds within the container will not affect build directories in the local directory outside the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,13 +6,16 @@ RUN echo "Building fVDB container in $MODE mode"
 # used for cross-compilation in docker build
 ENV FORCE_CUDA=1
 
-WORKDIR /openvdb
+# Create and use a consistent workspace directory
+RUN mkdir -p /workspace
+WORKDIR /workspace
 
+# Conda setup
 # force this CUDA version to be used to build the docker container because `docker build` does not
 # expose the GPU to the docker build process for it to be detected
-ENV CONDA_OVERRIDE_CUDA=12.0
-# copy env/dev_environment.yml to /tmp/
-COPY fvdb/env/dev_environment.yml /tmp/
+# fVDB team NOTE!!! We've tried removing this ENV line and it doesn't work, container build fails.
+ENV CONDA_OVERRIDE_CUDA=12.8
+COPY env/dev_environment.yml /tmp/
 RUN  conda env create -f /tmp/dev_environment.yml
 
 RUN conda init
--- a/README.md
+++ b/README.md
@@ -35,34 +35,19 @@ During the project's initial development stages, it is necessary to [run the bui
 ## Building *f*VDB from Source
 
 ### Environment Management
-ƒVDB is a Python library implemented as a C++ Pytorch extension.  Of course you can build ƒVDB in whatever environment suits you, but we provide three paths to constructing reliable environments for building and running ƒVDB:  using [docker](#setting-up-a-docker-container), [conda](#setting-up-a-conda-environment), or a Python virtual environment.
+ƒVDB is a Python library implemented as a C++ Pytorch extension.  Of course you can build ƒVDB in whatever environment suits you, but we provide three paths to constructing reliable environments for building and running ƒVDB. These are separate options,
+choose only one. They're not intended to be used together.
+
+1. **RECOMMENDED** [conda](#option-1-setting-up-a-conda-environment-recommended)
+2. Using [docker](#option-2-setting-up-a-docker-container)
+3. Python virtual environment. [venv](#option-3-setting-up-a-python-virtual-environment)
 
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -95,7 +80,34 @@ conda activate fvdb
...
+```
+
+Additional information about the ƒVDB docker setup, with troubleshooting for common errors, can be found
+here: [`ƒVDB Docker`](docs/markdown/docker_readme.md)
+
+---
+
+#### **OPTION 3** Setting up a Python virtual environment
 
 After creating a Python virtual environment, proceed to install the exact version of PyTorch that corresponds to your CUDA version. Then, install the rest of the build requirements.
```
swahtz: I think we should change this phrase to 'Create a python virtual environment and then proceed to install…'. Starting with this 'After creating' phrase makes it sound like we've already done something in before we started Option 3

docker-compose.yml
```diff
@@ -2,12 +2,22 @@ version: '3'
...
+      dockerfile: Dockerfile
       args:
         MODE: dev
+    working_dir: /workspace
     volumes:
-      - ../:/openvdb
+      - type: bind
+        source: .
+        target: /workspace
+      # Keep build artifacts inside the container (not mirrored to host)
```
swahtz, line 14: Do you think there's ever a workflow where we want to retrieve build artifacts from the container? Say for distribution or sending to someone who is working in the same environment.  Is there any pathway for doing that?

README.md
swahtz:
```suggestion
Create a python virtual environment and then proceed to install the exact version of PyTorch that corresponds to your CUDA version. Finally, install the rest of the build requirements.
```

docker-compose.yml
blackencino: I'll update the docs to show how to build a wheel, and then use `docker cp` to copy the artifact out of the repo. Since we're not creating a manylinux build, it isn't totally portable, but that's not a thing we should try to design.

docker-compose.yml
```diff
@@ -2,12 +2,22 @@ version: '3'
...
+      context: ./
+      dockerfile: Dockerfile
       args:
         MODE: dev
+    working_dir: /workspace
     volumes:
-      - ../:/openvdb
+      - type: bind
+        source: .
+        target: /workspace
```
harrism, line 13: **❓ question:** ‏Why not call this `/fvdb`?

Dockerfile
```diff
@@ -6,13 +6,15 @@ RUN echo "Building fVDB container in $MODE mode"
...
 ENV FORCE_CUDA=1
 
-WORKDIR /openvdb
+# Create and use a consistent workspace directory
+RUN mkdir -p /workspace
+WORKDIR /workspace
 
 # force this CUDA version to be used to build the docker container because `docker build` does not
 # expose the GPU to the docker build process for it to be detected
 ENV CONDA_OVERRIDE_CUDA=12.0
```
harrism: **❓ question:** ‏Should this match the version in dev_environment.yml?

docker-compose.yml
blackencino: I like disambiguating the workspace from the named directories. Otherwise it just gets confusing - is that the "fvdb" subdir of the repo itself?

Dockerfile
blackencino: I've never really understood this, it seems like it should be matching. Or possibly removed?

README.md
```diff
@@ -99,9 +99,18 @@ When you are ready to build ƒVDB, run the following command within the docker c
 conda activate fvdb;
 cd /workspace;
 TORCH_CUDA_ARCH_LIST="7.5;8.0;8.6+PTX" \
-./build.sh install verbose
+./build.sh wheel verbose
```
harrism: For developers, `install` was correct here.

harrism: Or at least, that's what I use.

Dockerfile
harrism: I don't know the details (I didn't write it), but we have CUDA 12.8 in our dev_environment.yml so I would think you would want to match that.

README.md
blackencino: Fixed this, thank you for helping to clarify.

Dockerfile
blackencino: I tried removing it, and then `docker compose build ...` fails, for exactly the reasons that the comment suggested. So instead I set it to 12.8, and it's a thing we'll have to remember to update, unless we write a clever thing to extract it from the env/dev_environment.html

swahtz: Yes, sadly, there's no way to make `docker compose` (or build) expose the GPU and driver so you have to manually specify what the CUDA version would be that's supported by the driver.  And you have to do this before you build the conda environment

PR #164, opened 2025-09-24
ScopedTimer in python

blackencino:
A context manager for timing code blocks with optional CUDA synchronization.

    Examples:
        Basic usage:
        >>> with ScopedTimer(message="Elapsed time"):
        ...     # some code
        ...     pass
        >>> # Elapsed time: 1.2345 seconds

        With CUDA synchronization:
        >>> with ScopedTimer(message="GPU operation", cuda=True):
        ...     # GPU operations
        ...     result = torch.matmul(a, b)
        >>> # GPU operation: 1.2345 seconds

        Splitting timings:
        >>> with ScopedTimer(cuda=True) as timer:
        ...     # setup phase
        ...     kmap, _ = grid.sparse_conv_kernel_map(kernel_size=kernel.size(-1), stride=1)
        ...     setup_time = timer.split()
        ...     # computation phase
        ...     out_feature = kmap.sparse_conv_3d(in_feature, kernel)
        ...     compute_time = timer.split()
        >>> print(f"Setup: {setup_time:.4f}s, Compute: {compute_time:.4f}s")


tests/unit/test_utils_timer.py
```diff
@@ -0,0 +1,76 @@
...
+import sys
+import time
+import unittest
+
+import pytest
+import torch
+from fvdb.utils.timer import ScopedTimer
+
+
+class TestScopedTimer(unittest.TestCase):
```
fwilliams, line 15: Docstrings or nah because tests?

blackencino: nah because tests and this is both tiny and self-explanatory

fvdb/utils/timer.py
```diff
@@ -0,0 +1,86 @@
...
+        self.start_time: float | None = None
+        self.end_time: float | None = None
+        self.elapsed_time: float | None = None
+        self._last_split_time: float | None = None
+        self.message: str = message
+
+    def __enter__(self):
+        """Enter the context manager and start timing."""
+        if self.cuda:
+            torch.cuda.synchronize()
```
harrism: **💡 suggestion:** ‏Is this a whole device synchronize? (`cudaDeviceSynchronize()`) Either way, synchronizing every time you start and stop a timer is not a good way to time CUDA code because it can slow down the whole application.  Using `torch.cuda.Event` is lighter weight.  Keep in mind though that in multi-stream code with concurrent kernels timing accuracy may be off depending on the situation.  But I think that's better than the invasive approach of adding a lot of synchronization.

blackencino: This is just a cut & paste of the timing code that's already in the tests. Here's the warp timer, which has an option to use nvtx.
```
# timer utils
class ScopedTimer:
    indent = -1

    enabled = True

    def __init__(
        self,
        name: str,
        active: bool = True,
        print: bool = True,
        detailed: bool = False,
        dict: dict[str, list[float]] | None = None,
        use_nvtx: bool = False,
        color: int | str = "rapids",
        synchronize: bool = False,
        cuda_filter: int = 0,
        report_func: Callable[[list[TimingResult], str], None] | None = None,
        skip_tape: bool = False,
    ):
        """Context manager object for a timer

        Parameters:
            name: Name of timer
            active: Enables this timer
            print: At context manager exit, print elapsed time to ``sys.stdout``
            detailed: Collects additional profiling data using cProfile and calls ``print_stats()`` at context exit
            dict: A dictionary of lists to which the elapsed time will be appended using ``name`` as a key
            use_nvtx: If true, timing functionality is replaced by an NVTX range
            color: ARGB value (e.g. 0x00FFFF) or color name (e.g. 'cyan') associated with the NVTX range
            synchronize: Synchronize the CPU thread with any outstanding CUDA work to return accurate GPU timings
            cuda_filter: Filter flags for CUDA activity timing, e.g. ``warp.TIMING_KERNEL`` or ``warp.TIMING_ALL``
            report_func: A callback function to print the activity report.
              If ``None``,  :func:`wp.timing_print() <timing_print>` will be used.
            skip_tape: If true, the timer will not be recorded in the tape

        Attributes:
            extra_msg (str): Can be set to a string that will be added to the printout at context exit.
            elapsed (float): The duration of the ``with`` block used with this object
            timing_results (list[TimingResult]): The list of activity timing results, if collection was requested using ``cuda_filter``
        """
        self.name = name
        self.active = active and self.enabled
        self.print = print
        self.detailed = detailed
        self.dict = dict
        self.use_nvtx = use_nvtx
        self.color = color
        self.synchronize = synchronize
        self.skip_tape = skip_tape
        self.elapsed = 0.0
        self.cuda_filter = cuda_filter
        self.report_func = report_func or wp.timing_print
        self.extra_msg = ""  # Can be used to add to the message printed at manager exit

        if self.dict is not None:
            if name not in self.dict:
                self.dict[name] = []

    def __enter__(self):
        if not self.skip_tape and warp.context.runtime is not None and warp.context.runtime.tape is not None:
            warp.context.runtime.tape.record_scope_begin(self.name)
        if self.active:
            if self.synchronize:
                wp.synchronize()

            if self.cuda_filter:
                # begin CUDA activity collection, synchronizing if needed
                timing_begin(self.cuda_filter, synchronize=not self.synchronize)

            if self.detailed:
                self.cp = cProfile.Profile()
                self.cp.clear()
                self.cp.enable()

            if self.use_nvtx:
                import nvtx

                self.nvtx_range_id = nvtx.start_range(self.name, color=self.color)

            if self.print:
                ScopedTimer.indent += 1

                if warp.config.verbose:
                    indent = "    " * ScopedTimer.indent
                    print(f"{indent}{self.name} ...", flush=True)

            self.start = time.perf_counter_ns()

        return self

    def __exit__(self, exc_type, exc_value, traceback):
        if not self.skip_tape and warp.context.runtime is not None and warp.context.runtime.tape is not None:
            warp.context.runtime.tape.record_scope_end()
        if self.active:
            if self.synchronize:
                wp.synchronize()

            self.elapsed = (time.perf_counter_ns() - self.start) / 1000000.0

            if self.use_nvtx:
                import nvtx

                nvtx.end_range(self.nvtx_range_id)

            if self.detailed:
                self.cp.disable()
                self.cp.print_stats(sort="tottime")

            if self.cuda_filter:
                # end CUDA activity collection, synchronizing if needed
                self.timing_results = timing_end(synchronize=not self.synchronize)
            else:
                self.timing_results = []

            if self.dict is not None:
                self.dict[self.name].append(self.elapsed)

            if self.print:
                indent = "    " * ScopedTimer.indent

                if self.timing_results:
                    self.report_func(self.timing_results, indent=indent)
                    print()

                if self.extra_msg:
                    print(f"{indent}{self.name} took {self.elapsed:.2f} ms {self.extra_msg}")
                else:
                    print(f"{indent}{self.name} took {self.elapsed:.2f} ms")

                ScopedTimer.indent -= 1


```

blackencino: I switched to using torch.cuda.Event() instead.

PR #179, opened 2025-09-26
Fix viewer API and camera

fwilliams:
Had an issue with the camera due to sign issues with the viewer. Fix this and expose a proper orbit camera API


fvdb/viz/viewer.py
```diff
@@ -206,21 +367,25 @@ def set_camera_lookat(
...
-        self._impl.set_camera_origin(*camera_origin_vec3f)
-        self._impl.set_camera_view_direction(*view_direction)
+        # Check that the view direction is not parallel to the up direction
+        dot_product = np.dot(view_direction_vec3f, up_direction_vec3f)
+        if abs(dot_product) > 0.999:
+            raise ValueError("View direction and up direction cannot be parallel or anti-parallel.")
+
+        self._impl.set_camera_orbit_center(*lookat_point_vec3f)
+        self._impl.set_camera_view_direction(*view_direction_vec3f)
+        self._impl.set_camera_orbit_radius(-orbit_radius)
```
areidmeyer: This negative should not be needed with the latest NanoVDB Editor version.

areidmeyer (approved):
Looks okay. Just a sign flip that shouldn't be needed anymore.

fvdb/viz/viewer.py
fwilliams: fixed!

src/python/ViewerBinding.cpp
```diff
@@ -2,8 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-#include "TypeCasters.h"
```
matthewdcong, line 5: I know this is already merged but this line needs to stay there to resolve #168.

PR #184, opened 2025-09-29
Add rank 3 type checking and caster to Mat44f batch

fwilliams:
Adds support for rank3 type checking to `fvdb.types`. These types are implemented using the type `NumericMaxRank3`. I also added casters `to_Mat44fBatch` and `to_Mat44fBatchBroadcastable`


fvdb/types.py
```diff
@@ -261,6 +262,12 @@ def is_SequenceOfSequenceOfNumericScalarNative(x: Any) -> TypeGuard[Sequence[Seq
     return isinstance(x, Sequence) and all(is_SequenceOfNumericScalarNative(item) for item in x)
 
 
+def is_SequenceOfSequenceOfSequenceOfNumericScalarNative(
```
blackencino, line 265: There's probably a clever way to compress this GermanSuperNoun, but for now, we can leave it.

fwilliams: Ya indeed :)

PR #185, opened 2025-09-29
Expose ProjectionType as a bound class instead of a typecaster

matthewdcong:
We've been running into ODR issues with `GaussianSplat3d.ProjectionType`. See #167 , #168 , and #182. This PR proposes a long-term solution by adding Python bindings for the `ProjectionType` enum. This is preferable from a programmatic standpoint because it avoids the pitfalls of type casters (e.g. their need to be included consistently in all TUs, the allocation upon switching from Python to C++ and vice versa) and it promotes stricter cross-language type-safety. Furthermore, on the C++ side, `ProjectionType` is nested within `GaussianSplat3d` and it makes sense from a consistency standpoint to mimic that on the Python side as well. (Currently, `GaussianSplat3d` is a bound class while it's nested `ProjectionType` is a opaque implicit type caster.)

Note that this is breaking change from the Python standpoint. Existing instances of `"perspective"` or `"orthographic"` will need to be changed to their typed counterparts, but this is a relatively straightforward change.

### DIFF
--- a/fvdb/gaussian_splatting.py
+++ b/fvdb/gaussian_splatting.py
@@ -22,6 +22,7 @@ class GaussianSplat3d:
         Possible values include PERSPECTIVE or ORTHOGRAPHIC for perspective
         and orthographic proections respectively.
         """
+
         pass
 
     PLY_VERSION_STRING = "fvdb_ply 1.0.0"


fvdb/gaussian_splatting.py
```diff
@@ -15,6 +15,8 @@
 
 
 class GaussianSplat3d:
+    class ProjectionType(GaussianSplat3dCpp.ProjectionType):
```
harrism, line 18: **☑️ todo:** ‏Doc this class, including a list of supported types? (I think we just have PERSPECTIVE and ORTHOGRAPHIC so far.

fvdb/gaussian_splatting.py
```diff
@@ -788,8 +790,8 @@ def project_gaussians_for_depths(
             image_height (int): The height of the images to be rendered.
             near (float): The near clipping plane distance for the projection.
             far (float): The far clipping plane distance for the projection.
-            projection_type (str): The type of projection to use. Default is "perspective".
-                Other options could include "orthographic".
+            projection_type (ProjectionType): The type of projection to use. Default is PERSPECTIVE.
+                Other options could include ORTHOGRAPHIC.
```
harrism: **💡 suggestion:** ‏If we doc the ProjectionType class, can drop these "Other options could include" lines. "could include" is a bit wishy-washy.

harrism (approved):
Thanks @matthewdcong , looks good. Just a couple of doc requests.

fvdb/gaussian_splatting.py
matthewdcong: Good idea, done!

PR #188, opened 2025-09-30
Make editor sub-build and CPM cache more robust

harrism:
Some container sub-builds of `nanovdb-editor` wheels inside the fvdb build could fail silently because inside the container the build directory was not pre-created. This change creates the build working directory before attempting to build. Also attempts to make error diagnostic more useful.

Likewise, it creates the `CPM_SOURCE_CACHE` directory if it doesn't already exist, so the container build can use it, and tries to make more useful error output.

I was having trouble with the editor sub-build in the benchmark container.


phapalova (approved):
LGTM

PR #189, opened 2025-09-30
Don't keep reference to splats in viewer

fwilliams:
The GaussianSplat3dView keeps a reference to a `GaussianSplat3d` which is not needed for rendering and can lead to extra GPU memory usage.


areidmeyer (approved):
Approved

PR #190, opened 2025-09-30
CameraView - fix camera state values, docs

phapalova:


fvdb/viz/_viewer.py
```diff
@@ -74,12 +76,49 @@ def add_gaussian_splat_3d(
     def add_camera_view(
         self,
         name: str,
-        camera_to_world_matrices: torch.Tensor | None = None,
-        projection_matrices: torch.Tensor | None = None,
+        cam_to_world_matrices: Sequence[torch.Tensor | np.ndarray] | torch.Tensor | np.ndarray,
```
fwilliams: I added a `NumericMaxRank3` and `toMat44f` function in types.py which you can use in this case

fvdb/viz/_camera_view.py
```diff
@@ -84,26 +87,61 @@ def axis_thickness(self, thickness: float):
...
 
     @property
     def frustum_scale(self) -> float:
+        """
+        Get the scale factor applied to the frustum visualization.
+        """
         return self._view.frustum_scale
 
     @frustum_scale.setter
     def frustum_scale(self, scale: float):
```
fwilliams: The more general way (that we totally don't always follow) is to use `NumericScalarNative` instead of `float` here, which won't complain if you pass in e.g. np.float32

src/fvdb/detail/viewer/Viewer.cpp
```diff
@@ -290,49 +296,43 @@ Viewer::addCameraView(const std::string &name,
...
+        // Used for frustum visualization
+        it->second.mView.configs[i].near_plane = frustumNear;
+        it->second.mView.configs[i].far_plane  = frustumFar;
 
-        it->second.mView.configs[i].fov_angle_y  = fov;
-        it->second.mView.configs[i].aspect_ratio = std::max(aspectRatio, 0.f);
-        it->second.mView.configs[i].is_orthographic =
-            isOrthographic ? PNANOVDB_TRUE : PNANOVDB_FALSE;
+        // TODO (@phapalova) This will be properly set with images support
+        it->second.mView.configs[i].fov_angle_y  = DEFAULT_CAMERA_FOV_RADIANS;
```
fwilliams, line 334: We likely still want to support this without images by just taking in a w/h pair per camera. This can be in a follow up PR though. 

Image support is nice-to-have for the release but probably not top priority!

PR #191, opened 2025-10-01
Update conda envs to torch 2.8.0 and CUDA 12.9

swahtz:
Updates all the conda environments and tests to CUDA 12.9 and pytorch 2.8.0
Switched torchvision to the pip packages until a conda-forge one is available


env/dev_environment.yml
```diff
@@ -65,15 +65,15 @@ dependencies:
   - pyyaml
   - scikit-learn
   - srt
-  - torchvision
+  # - torchvision # torchvision is not available for pytorch 2.8.0 yet
   - tyro
   - utm
   - viser
   - pip:
     - point-cloud-utils
```
harrism, line 72: **💡 suggestion:** We can get torchvision via pip temporarily like this, I believe (note index-url from the pytorch.org configurator command line)


```suggestion
- pip:
  - --index-url https://download.pytorch.org/whl/cu129
  - torchvision
  - point-cloud-utils
```

swahtz: Good idea!  I'll use this but I'll just reorder where the index-url and torchvision items appear in the list so they don't override the index URL for the other pip packages.

harrism: Cool. I don't know if it will override them, since their wheels are not found there.

harrism (approved):

PR #197, opened 2025-10-01
[CI] Fix git issues

swahtz:
Fixes an issue @phapalova was seeing where sometimes the git user wasn't configured
Fix not having git installed on the pip/venv based tests before it was used.

NOTE:  These tests will fail because the job definitions still come from `main` during a `pull_request_target` event


blackencino (approved):
CI like the wind!

PR #204, opened 2025-10-02
Fix GridBatch Python/C++ function naming mismatches

swahtz:
Matched naming of C++ GridBatch functions to their python names and fixed corresponding function names in tests and in `_Cpp.pyi`
These functions include `coords_in_active_voxel`, `points_in_active_voxel`, `subdivided_grid` and `subdivide`


swahtz:
> Pretty straightforward, just don't understand how this was missed?

I assume it was a case of 'we're about to change a lot of this API, let's just make these 2 changes for the moment'.  And also tests weren't properly running

PR #213, opened 2025-10-03
Fuse log into projection

fwilliams:
Our previous kernels took in scales directly, but we store log(scales) to ensure the parameter is always valid. This change makes the projection kernel accept logScales instead, avoiding an extra operation (and derivative). 

This change requires us to update the gtests to use higher tolerances since the original test data was computed using scales as input rather than logScales. While the tolerances are larger, they are still a tiny fraction of the actual range of the data, so probably fine. 

I've also trained several splats in reality capture with this change. The only difference is a slight speedup.

Thanks @matthewdcong for the original code (I just fixed up the tests).

### DIFF
--- a/fvdb/gaussian_splatting.py
+++ b/fvdb/gaussian_splatting.py
@@ -441,7 +441,7 @@ class GaussianSplat3d:
         Note: To read the opacities, see the `opacities` property (which is read-only).
 
         Returns:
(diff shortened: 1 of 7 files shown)


matthewdcong (approved):
Looks good, thanks @fwilliams !

PR #216, opened 2025-10-03
Docs tests work in Dev environment

blackencino:
"fvdb" conda environment now has everything it needs to do docs testing.

The build.sh file now also has a "docstest" target which will run the documentation tests. You can run them manually like so:

```
pytest --markdown-docs ./docs
```

The example/test data was failing in some cases because if the directory `/tmp/fvdb_example_data` already existed, one of the functions which did git commands to clone that data would skip it, but then fail on subsequent attempts to check out a particular hash. We had duplicate functions for cloning test and example data, so I've combined them and added the required "fetch" if the repo already existed.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -12,6 +12,7 @@ usage() {
   echo "                   install    - Build and install the package (default)."
   echo "                   wheel      - Build the Python wheel."
   echo "                   ctest      - Run tests (requires tests to be built)."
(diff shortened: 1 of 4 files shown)


swahtz (approved):
Looks great.  Thanks for doing this and catching that issue with the example/test data fetching.

PR #217, opened 2025-10-03
Morton Order for supporting point transformer v3

TarzanZhao:
Point transformer v3 requires attention along the different serialization orders. This PR supports different z-order and transposed z-order. Next, I will implement hilbert order and transposed hilbert order to match with original ptv3 faithfully.

### DIFF
--- a/fvdb/grid.py
+++ b/fvdb/grid.py
@@ -1835,6 +1835,40 @@ class Grid:
     def ijk(self) -> torch.Tensor:
         return self._impl.ijk.jdata
 
+    def serialize_encode(self, order_type: str = "z") -> torch.Tensor:
+        """
+        Return the Morton codes for active voxels in this grid.
+        
+        Morton codes provide a space-filling curve that maps 3D coordinates to 1D integers,
+        preserving spatial locality. This is useful for serialization, sorting, and 
+        spatial data structures.
+
+        Returns:
+            torch.Tensor: A tensor of shape `[num_voxels, 1]` containing
+                the Morton codes for each active voxel.
+        """
+        return self._impl.serialize_encode(order_type).jdata
+
+    def permute(self, order_type: str = "z") -> torch.Tensor:
+        """
+        Get permutation indices to sort voxels by spatial order.
+        
+        This method computes Morton codes for all active voxels and returns the indices
+        that would sort them according to the specified ordering. This is useful for
+        spatially coherent data access patterns and cache optimization.
+
+        Args:
+            order_type (str): The type of spatial ordering to use:
+                - "morton": Morton Z-order curve (default, ascending)
+                - "ascending": Sort Morton codes in ascending order (same as "morton")
+                - "descending": Sort Morton codes in descending order
+
+        Returns:
+            torch.Tensor: A tensor of shape `[num_voxels, 1]` containing
+                the permutation indices. Use these indices to reorder voxel data for spatial coherence.
+        """
+        return self._impl.permute(order_type).jdata
+
     @property
     def num_bytes(self) -> int:
         return self._impl.total_bytes
--- a/fvdb/grid_batch.py
+++ b/fvdb/grid_batch.py
@@ -1956,6 +1956,48 @@ class GridBatch:
     def ijk(self) -> JaggedTensor:
         return self._impl.ijk
 
+    def serialize_encode(self, order_type: str = "z") -> JaggedTensor:
+        """
+        Return the Morton codes for active voxels in this grid batch.
+        
+        Morton codes provide a space-filling curve that maps 3D coordinates to 1D integers,
+        preserving spatial locality. This is useful for serialization, sorting, and 
+        spatial data structures.
+
+        Returns:
+            JaggedTensor: A JaggedTensor of shape `[num_grids, -1, 1]` containing
+                the Morton codes for each active voxel in the batch.
+        """
+        return self._impl.serialize_encode(order_type)
+
+    def permute(self, order_type: str = "z") -> JaggedTensor:
+        """
+        Get permutation indices to sort voxels by spatial order.
+        
+        This method computes Morton codes for all active voxels and returns the indices
+        that would sort them according to the specified ordering. This is useful for
+        spatially coherent data access patterns and cache optimization.
+
+        Args:
+            order_type (str): The type of spatial ordering to use:
+                - "z": Regular Z-order curve (xyz bit interleaving, default)
+                - "z-trans": Transposed Z-order curve (zyx bit interleaving)
+                - "morton": Alias for "z" (for backward compatibility)
+                - "ascending": Alias for "z"
+                - "descending": Sort Morton codes in descending order
+
+        Returns:
+            JaggedTensor: A JaggedTensor of shape `[num_grids, -1, 1]` containing
+                the permutation indices. Use these indices to reorder voxel data for spatial coherence.
+                
+        Example:
+            >>> z_indices = grid_batch.permute("z")  # Regular xyz z-order
+            >>> z_trans_indices = grid_batch.permute("z-trans")  # Transposed zyx z-order
+            >>> # Use indices to reorder some voxel data
+            >>> reordered_data = voxel_data.jdata[z_indices.jdata.squeeze(-1)]
+        """
+        return self._impl.permute(order_type)
+
     @property
     def jidx(self) -> torch.Tensor:
         if self.has_zero_grids:
--- a/src/fvdb/GridBatch.cpp
+++ b/src/fvdb/GridBatch.cpp
@@ -20,6 +20,9 @@
 
 // Ops headers
 #include <fvdb/detail/ops/ActiveGridGoords.h>
+#include <fvdb/detail/ops/SerializeEncode.h>
+#include <tuple>
+#include <vector>
 #include <fvdb/detail/ops/CoordsInGrid.h>
 #include <fvdb/detail/ops/CubesInGrid.h>
 #include <fvdb/detail/ops/GridEdgeNetwork.h>
@@ -1079,6 +1082,61 @@ GridBatch::ijk() const {
     return FVDB_DISPATCH_KERNEL(this->device(), [&]() {
         return fvdb::detail::ops::dispatchActiveGridCoords<DeviceTag>(*mImpl);
     });
+}
+
+JaggedTensor
+GridBatch::serialize_encode(const std::string &order_type) const {
+    c10::DeviceGuard guard(device());
+    return FVDB_DISPATCH_KERNEL(this->device(), [&]() {
+        return fvdb::detail::ops::dispatchSerializeEncode<DeviceTag>(*mImpl, order_type);
+    });
+}
+
+JaggedTensor
+GridBatch::permute(const std::string &order_type) const {
+    c10::DeviceGuard guard(device());
+    
+    // Parse order type and determine Morton code type and sort order
+    std::string morton_order = "z";  // Default z-order (xyz)
+    
+    if (order_type == "z") {
+        morton_order = "z";
+    } else if (order_type == "z-trans") {
+        morton_order = "z-trans";
+    } else {
+        TORCH_CHECK(false, "Invalid order_type: ", order_type, 
+                    ". Valid options are 'z', or 'z-trans'.");
+    }
+    
+    // Get Morton codes for sorting
+    JaggedTensor morton_codes = serialize_encode(morton_order);
+    
+    // Create output tensor for permutation indices
+    auto opts = torch::TensorOptions().dtype(torch::kInt64).device(device());
+    std::vector<int64_t> shape = {mImpl->totalVoxels()};
+    torch::Tensor permutation_indices = torch::empty(shape, opts);
+    
+    // Sort Morton codes and get permutation indices for each grid
+    int64_t offset = 0;
+    for (int64_t grid_idx = 0; grid_idx < mImpl->batchSize(); ++grid_idx) {
+        int64_t num_voxels = mImpl->numVoxelsAt(grid_idx);
+        if (num_voxels == 0) continue;
+        
+        // Extract Morton codes for this grid
+        torch::Tensor grid_morton_codes = morton_codes.jdata().narrow(0, offset, num_voxels);
+        
+        // Sort and get indices
+        auto sort_result = torch::sort(grid_morton_codes.squeeze(-1), /*dim=*/0);
+        torch::Tensor sorted_values = std::get<0>(sort_result);
+        torch::Tensor indices = std::get<1>(sort_result);
+        
+        // Store indices with offset
+        permutation_indices.narrow(0, offset, num_voxels) = indices + offset;
+        
+        offset += num_voxels;
+    }
+    
+    return mImpl->jaggedTensor(permutation_indices.unsqueeze(-1));
 }
 
 std::vector<JaggedTensor>
--- a/src/fvdb/detail/ops/SerializeEncode.cu
+++ b/src/fvdb/detail/ops/SerializeEncode.cu
@@ -1,0 +1,190 @@
+// Copyright Contributors to the OpenVDB Project
+// SPDX-License-Identifier: Apache-2.0
+//
+#include <fvdb/detail/GridBatchImpl.h>
+#include <fvdb/detail/ops/SerializeEncode.h>
+#include <fvdb/detail/utils/AccessorHelpers.cuh>
+#include <fvdb/detail/utils/ForEachCPU.h>
+#include <fvdb/detail/utils/cuda/ForEachCUDA.cuh>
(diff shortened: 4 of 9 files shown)


fvdb/grid.py
```diff
@@ -1835,6 +1835,40 @@ def has_zero_voxels(self) -> bool:
     def ijk(self) -> torch.Tensor:
         return self._impl.ijk.jdata
 
+    def serialize_encode(self, order_type: str = "z") -> torch.Tensor:
```
fwilliams: Defering to @blackencino on this, but perhaps we should have explicit methods for different orders.

E.g. `morton_codes_for_active_voxels`,`morton_transposed_codes_for_active_voxels`, `hilbert_codes_for_active_voxels`, `hilbert_transposed_code_for_active_voxels`, etc.

fvdb/grid.py
```diff
@@ -1835,6 +1835,40 @@ def has_zero_voxels(self) -> bool:
...
+        preserving spatial locality. This is useful for serialization, sorting, and 
+        spatial data structures.
+
+        Returns:
+            torch.Tensor: A tensor of shape `[num_voxels, 1]` containing
+                the Morton codes for each active voxel.
+        """
+        return self._impl.serialize_encode(order_type).jdata
+
+    def permute(self, order_type: str = "z") -> torch.Tensor:
```
fwilliams: similarly, we can have `morton_order_permutation_indices` or something less wordy. @blackencino you're better at this than me.

fvdb/grid.py
blackencino: Yes - I agree with @fwilliams. Though it does become a proliferation of methods and a certain amount of bloat, we've explicitly decided in other similar scenarios not to use "config" params to functions, either as strings or enums. For consistency with the rest of the API, use method names like:

```
encode_hilbert
encode_morton
encode_hilbert_zyx
encode_morton_zyx
```

At some point in the future, we may change our design around config enums (or strings), but for now this is the convention.

fvdb/grid.py
blackencino:
```
permutation_hilbert
permutation_morton
permutation_hilbert_zyx
permutation_morton_zyx
```

Same reasons as above, same "maybe we'll change this downstream" acknowledgement.

fvdb/grid_batch.py
```diff
@@ -1956,6 +1956,56 @@ def has_zero_grids(self) -> bool:
...
+                - "hilbert": Regular Hilbert curve (xyz)
+                - "hilbert-trans": Transposed Hilbert curve (zyx)
+
+        Returns:
+            JaggedTensor: A JaggedTensor of shape `[num_grids, -1, 1]` containing
+                the space-filling curve codes for each active voxel in the batch.
+        """
+        return self._impl.serialize_encode(order_type)
+
+    def permute(self, order_type: str = "z") -> JaggedTensor:
```
blackencino: If this is ultimately just "argsort(encode_{encoding})", we should just let that be done in torch directly.

src/fvdb/GridBatch.cpp
```diff
@@ -1081,6 +1084,65 @@ GridBatch::ijk() const {
...
+JaggedTensor
+GridBatch::serialize_encode(const std::string &order_type) const {
+    c10::DeviceGuard guard(device());
+    return FVDB_DISPATCH_KERNEL(this->device(), [&]() {
+        return fvdb::detail::ops::dispatchSerializeEncode<DeviceTag>(*mImpl, order_type);
+    });
+}
+
+JaggedTensor
+GridBatch::permute(const std::string &order_type) const {
```
blackencino: Don't use std::string for configuration selection in C++. Though sometimes the small-string optimization can avoid this, std::string often involves memory allocations and is just a needlessly expensive way to select configuration. On the C++ side, we have clear precedent for configuration via enum, which I think would be valid here.

src/fvdb/GridBatch.cpp
```diff
@@ -1081,6 +1084,65 @@ GridBatch::ijk() const {
...
+    c10::DeviceGuard guard(device());
+    
+    // Parse order type and determine space-filling curve type and sort order
+    std::string curve_order = "z";  // Default z-order (xyz)
+    
+    if (order_type == "z") {
```
blackencino: Enum instead of string, then use a switch statement rather than a cascading if/else chain.

src/fvdb/GridBatch.cpp
```diff
@@ -1081,6 +1084,65 @@ GridBatch::ijk() const {
...
+    int64_t offset = 0;
+    for (int64_t grid_idx = 0; grid_idx < mImpl->batchSize(); ++grid_idx) {
+        int64_t num_voxels = mImpl->numVoxelsAt(grid_idx);
+        if (num_voxels == 0) continue;
+        
+        // Extract space-filling curve codes for this grid
+        torch::Tensor grid_curve_codes = curve_codes.jdata().narrow(0, offset, num_voxels);
+        
+        // Sort and get indices
+        auto sort_result = torch::sort(grid_curve_codes.squeeze(-1), /*dim=*/0);
```
blackencino: This should just be done external to the C++ in torch via argsort.

src/fvdb/detail/ops/SerializeEncode.cu
```diff
@@ -0,0 +1,210 @@
...
+#include <vector>
+
+namespace fvdb {
+namespace detail {
+namespace ops {
+
+/// @brief Per-voxel callback which computes the space-filling curve code (Morton or Hilbert) for each active voxel in a batch of grids
+template <template <typename T, int32_t D> typename TorchAccessor>
+__hostdev__ inline void
+serializeEncodeVoxelCallback(int64_t batchIdx,
```
blackencino, line 27: Rather than performing the xyz-zyx transposition here, do it in the python code.

fvdb/grid.py
TarzanZhao: OK

fvdb/grid.py
TarzanZhao: OK

fvdb/grid.py
TarzanZhao: For the exposed API to users, we should avoid using 'config' parameters. Should we also avoid 'config' parameters in the internal CUDA implementations which users will not touch themselves?

fvdb/grid_batch.py
TarzanZhao: Good suggestions

TarzanZhao: I have moved permute from c++ to python side. 

However, I find it challenging to convert the coordinates from ijk to kji in Python because I'm using `const nanovdb::Coord &ijk = leaf.offsetToGlobalCoord(voxelIdx);` and `const nanovdb::Coord &bboxMin = bboxMinCoords[batchIdx];` within the cuda kernel to get each voxel's coordinate. These are internal states of gridBatch, which seems inappropriate to modify from outside (python side).

fvdb/grid.py
fwilliams: I think it's fine to have an enum internally if it reduces duplicated code.

fwilliams (dismissed):
A few comments:

I think the morton/hilbert/etc codes should be attributes on the grid in the same way that ijk is an attribute. *i.e* `grid.morton_codes` `grid.hilbert_codes` and variants for the transposed versions.

I also agree with Chris, that you can permute outside the grid. 

So you have something like

```
grid = Grid.from_xxx(...)
sidecar = grid.inject_from_ijk(ijk_values, data)

morton_pmt = torch.argsort(grid.morton_codes)
sidecar_morton = sidecar[morton_pmt]
```

fvdb/grid_batch.py
```diff
@@ -1956,6 +1956,139 @@ def has_zero_grids(self) -> bool:
...
+    def permute(self, curve_codes: JaggedTensor) -> JaggedTensor:
+        """
+        Get permutation indices to sort voxels by spatial order.
+
+        This method computes space-filling curve codes for all active voxels and returns the indices
+        that would sort them according to the specified ordering. This is useful for
+        spatially coherent data access patterns and cache optimization.
+
+        Args:
+            order_type (str): The type of spatial ordering to use:
```
blackencino: The documentation string doesn't match the implementation. Now that you're permuting the encoding directly, the docs should reflect that.

src/fvdb/detail/ops/SerializeEncode.cu
```diff
@@ -0,0 +1,212 @@
...
+        cudaMalloc(&bboxMinPtr, bboxMinCoords.size() * sizeof(nanovdb::Coord));
+        cudaMemcpy(bboxMinPtr,
+                   bboxMinCoords.data(),
+                   bboxMinCoords.size() * sizeof(nanovdb::Coord),
+                   cudaMemcpyHostToDevice);
+    } else {
+        bboxMinPtr = bboxMinCoords.data();
+    }
+
+    GetSerializeEncode<DeviceTag>(gridBatch, outMortonCodes, bboxMinPtr, order_type_int);
```
blackencino: We don't usually use the min bbox coord as an offset for morton or hilbert encoding. If this is necessary to match point transformer v3, then keep it for now. However, this would mean that the morton encoding will move around as the bounding box changes, which is not normally the desired behavior (same for hilbert). 

Because we are limited to -2^20, 2^20 in our coordinates, you can have the offset be (2^21-1). 

The downside of that offset is that the points which are close to the signed origin of (0,0,0) have a large encoding jump in morton (though not in hilbert). But by contrast, the encoding remains static regardless of the spread of points.  Choose which of these properties is more important, and just make a note of it here.

blackencino (changes requested):
I think this can go through with minimal documentation changes. I would like the offset, currently based on bounding box min, to be justified/explained with a comment in the code. Alternatively, you might choose to use the constant offset of (2^21-1), but either is ok, so long as it is explained. There's a documentation error in the permutation function that needs fixing.

src/fvdb/detail/ops/SerializeEncode.cu
TarzanZhao: I used the "min bbox coord" as an offset because current encoding requires to avoid any negative coordinates numbers in the morton order encoding. 

Thanks for pointing out! I will leave a note there.

fvdb/grid_batch.py
TarzanZhao: OK, I will fix this.

fwilliams:
Okay @blackencino and I discussed. Here's the conclusion:

1. Rename `encode_morton/hilbert/etc.` to `.morton/hilbert/etc.` These are methods which take in an optional base offset:
*.i.e.* 
```python

# Offset 
def hilbert(base_offset=None):
  ```
  if called with None, use the -bbmin as the offset, otherwise, use the passed in offset
  offset is added to the ijk values before they are transposed if they are transposed
  ```
2. Get rid of permute methods on the grid and in C++. Do this in Python by argsorting the space filling codes and indexing the sidecar.

File: fvdb/grid.py [hunk 10]
```diff
@@ -1835,6 +1835,148 @@ def has_zero_voxels(self) -> bool:
...
+
+        Returns:
+            torch.Tensor: A tensor of shape `[num_voxels, 1]` containing
+                the transposed Hilbert codes for each active voxel.
+        """
+        if offset is None:
+            offset = -torch.min(self.ijk, dim=0).values
+        return self._impl.hilbert_zyx(offset).jdata
+
+    def permute(self, curve_codes: torch.Tensor) -> torch.Tensor:
```
blackencino: We agreed to remove this function.

fvdb/grid.py
```diff
@@ -1835,6 +1835,148 @@ def has_zero_voxels(self) -> bool:
...
+        # Get the curve codes as a flat tensor
+        curve_data = curve_codes.squeeze(-1)  # Shape: [num_voxels]
+
+        # Sort and get indices
+        _, indices = torch.sort(curve_data, dim=0)
+
+        # Return indices as a tensor with shape [num_voxels, 1]
+        return indices.unsqueeze(-1)
+
+    def permutation_morton(self) -> torch.Tensor:
```
blackencino: Remove

fvdb/grid.py
```diff
@@ -1835,6 +1835,148 @@ def has_zero_voxels(self) -> bool:
...
+        """
+        Return permutation indices to sort voxels by Morton curve order.
+
+        Returns:
+            torch.Tensor: A tensor of shape `[num_voxels, 1]` containing
+                the permutation indices for Morton (Z-order) curve ordering.
+        """
+        return self.permute(self.morton())
+
+    def permutation_morton_zyx(self) -> torch.Tensor:
```
blackencino: Remove

fvdb/grid.py
```diff
@@ -1835,6 +1835,148 @@ def has_zero_voxels(self) -> bool:
...
+        """
+        Return permutation indices to sort voxels by transposed Morton curve order.
+
+        Returns:
+            torch.Tensor: A tensor of shape `[num_voxels, 1]` containing
+                the permutation indices for transposed Morton (Z-order) curve ordering.
+        """
+        return self.permute(self.morton_zyx())
+
+    def permutation_hilbert(self) -> torch.Tensor:
```
blackencino: Remove

fvdb/grid.py
```diff
@@ -1835,6 +1835,148 @@ def has_zero_voxels(self) -> bool:
...
+        """
+        Return permutation indices to sort voxels by Hilbert curve order.
+
+        Returns:
+            torch.Tensor: A tensor of shape `[num_voxels, 1]` containing
+                the permutation indices for Hilbert curve ordering.
+        """
+        return self.permute(self.hilbert())
+
+    def permutation_hilbert_zyx(self) -> torch.Tensor:
```
blackencino: Remove

fvdb/grid_batch.py
```diff
@@ -1956,6 +1956,176 @@ def has_zero_grids(self) -> bool:
...
+                the transposed Hilbert codes for each active voxel in the batch.
+        """
+        if offset is None:
+            offset = -torch.min(self.ijk.jdata, dim=0).values
+        else:
+            offset = to_Vec3i(offset)
+
+        return self._impl.hilbert_zyx(offset)
+
+    def permute(self, curve_codes: JaggedTensor) -> JaggedTensor:
```
blackencino: Remove

fvdb/grid_batch.py
```diff
@@ -1956,6 +1956,176 @@ def has_zero_grids(self) -> bool:
...
+
+            # Store indices with offset
+            permutation_indices[offset : offset + num_voxels] = indices + offset
+
+            offset += num_voxels
+
+        # Return as JaggedTensor with the same structure as the input
+        return self.jagged_like(permutation_indices.unsqueeze(-1))
+
+    def permutation_morton(self, offset: NumericMaxRank1 | None = None) -> JaggedTensor:
```
blackencino: Remove

fvdb/grid_batch.py
```diff
@@ -1956,6 +1956,176 @@ def has_zero_grids(self) -> bool:
...
+        """
+        Return permutation indices to sort voxels by Morton curve order.
+        """
+        return self.permute(self.morton(offset))
+
+    def permutation_morton_zyx(self, offset: NumericMaxRank1 | None = None) -> JaggedTensor:
```
blackencino: Remove

fvdb/grid_batch.py
```diff
@@ -1956,6 +1956,176 @@ def has_zero_grids(self) -> bool:
...
+        """
+        Return permutation indices to sort voxels by transposed Morton curve order.
+        """
+        return self.permute(self.morton_zyx(offset))
+
+    def permutation_hilbert(self, offset: NumericMaxRank1 | None = None) -> JaggedTensor:
```
blackencino: Remove

fvdb/grid_batch.py
```diff
@@ -1956,6 +1956,176 @@ def has_zero_grids(self) -> bool:
...
+        """
+        Return permutation indices to sort voxels by Hilbert curve order.
+        """
+        return self.permute(self.hilbert(offset))
+
+    def permutation_hilbert_zyx(self, offset: NumericMaxRank1 | None = None) -> JaggedTensor:
```
blackencino: Remove

src/fvdb/detail/ops/SerializeEncode.cu
blackencino: We can do this later.

blackencino (approved):
Approved, but please remove the permute methods

fvdb/grid_batch.py
TarzanZhao: OK, I will remove these functions and implement these permutation in fVDB instead.

fvdb/grid_batch.py
blackencino: I don't fully understand this comment, but it is easy to just permute to get kji from ijk: `kjis = ijks.permute(0, 3, 2, 1)`

PR #222, opened 2025-10-07
Publish wheels GHA workflow

harrism:
This PR introduces a new Github Actions workflow to build and package fVDB wheels and upload them to a package index. Currently this is WIP.

This workflow has two modes: pypi (actually currently only TestPyPI) and S3. 

**PyPI Mode**

 - Runs whenever a commit is pushed without "[s3]" in the commit message. 
 - Once this PR is merged, the idea is to update the workflow so it can be manually invoked from the Github UI or CLI (`workflow_dispatch` trigger), or invoked by publishing a release (`release`) trigger.
 - Creates a wheel with no local version naming, repairs it as a manylinux wheel
 - Uploads to test.pypi.org/fvdb-core. This only supports one CUDA and PyTorch version (currently 12.9 and 2.8, respectively) due to PyPI limitations. 
 
**S3 Mode**

 - Runs whenever a commit with "[s3]" in the commit message is pushed, or once merged, by using `workflow_dispatch` with input `publish_target` set to `s3`.
 - Builds a default linux x86_64 (not manylinux) wheel and modifies the version string to append the pytorch and cuda versions e.g. `fvdb_core-0.2.1+pt28.cu129....whl`.
 - Uploads the wheels to our S3 `fvdb-packages` bucket and builds the index. This index building was not working on my last push.

NOTE: The last time I ran this I got the following error -- I've been iterating a lot so maybe this caused me to exceed and EC2 instance quota?

```
Failed to start EC2 instance in any availability zone. Errors: Failed to start EC2 instance with configuration 1 in region us-east-2: You have requested more vCPU capacity than your current vCPU limit of 1152 allows for the instance bucket that the specified instance type belongs to. Please visit http://aws.amazon.com/contact-us/ec2-request to request an adjustment to this limit.
```

**Other changes in this PR**

 - Modifies cmake to clean up our install to not install `blosc`, `cudnn_frontend`, and `tinyply` (e.g. their headers) which would clutter up our wheels.
 - Strip any debug symbols from non-debug builds (I don't think there were any). 
 - Add metadata to pyproject.toml
 - Also added capabilities to rename_wheels.py script but I don't think this script is necessary anymore and should be deleted. Renaming wheels to be different than the name stored in their metadata is not allowed by pypi. Also the functionality of this script is much simpler in bash. [NOTE: @swahtz removed this, this was here for historical reasons]

Signed-off-by: Mark Harris [mharris@nvidia.com](mailto:mharris@nvidia.com)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,12 +1,6 @@
 name: Publish fVDB Pip Package
 
 on:
-  pull_request:
-    branches:
-      - '**'
-  push:
-    branches:
-      - main
   release:
     types: [published]
   workflow_dispatch:
@@ -84,7 +78,7 @@ jobs:
           mode: start
           github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
           ec2-image-id: ami-0e14a711dad782a70
-          ec2-instance-type: m6a.16xlarge
+          ec2-instance-type: m6a.8xlarge
           subnet-id: subnet-03f2320d6e6e0005b
           security-group-id: sg-0cd08bd89d6212223
           label: ec2-${{ matrix.python-version }}-pt${{ matrix.torch-version }}-cu${{ matrix.cuda-version }}-${{ github.run_id }}


CMakeLists.txt
```diff
@@ -88,17 +88,19 @@ target_compile_definitions(_Cpp PRIVATE
 # Configure the pybind11 module with PyTorch compatibility settings
 configure_torch_pybind11(_Cpp)
 
+# Strip symbols in non-Debug builds to reduce wheel size
```
matthewdcong, line 94: What is the default configuration for local builds at the moment? i.e. will this affect debugging during development?

swahtz: Our default for local builds is not to build with debug.  I don't see any '-g' flags in our generated build commands and you have to explicitly enable debug builds with `./build.sh debug`.  

So we shouldn't have any debug symbols in our local builds by default currently and it seems this behaviour would not be triggered if you were trying to make a build with debug symbols.  It seems to be out of any abundance of caution?

matthewdcong: In that case, I think we should make this an option then that's off by default and enabled when building releases. Otherwise when we're running local development and we get a crash, we'll have less information to go on.

swahtz: I added a `strip_symbols` argument to `build.sh` that's off by default and turned on when we do releases in the action.  Will only turn on `-s` if you provide that arg but it's ignored if doing a `debug` build

PR #223, opened 2025-10-07
Make submodules module private and clean up docs

fwilliams:
Improve docs on GaussianSplat3d rendering functions to describe alpha parameter properly.

Rename internal modules to start with underscores since you should not do e.g. `from fvdb.grid import Grid` but rather `from fvdb import Grid`.

### DIFF
--- a/fvdb/__init__.py
+++ b/fvdb/__init__.py
@@ -13,6 +13,29 @@ if torch.cuda.is_available():
     torch.cuda.init()
 
 
+def _parse_device_string(device_or_device_string: str | torch.device) -> torch.device:
+    """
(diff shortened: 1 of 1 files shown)


blackencino:
Also, these kinds of sweeping changes at the last minute just cause a lot of code rewrites for client code, again - not for any particularly good reason. There's nothing bad that's going to happen if people mix & match "from fvdb.grid import Grid" or whatever.. it can only good-happen.

harrism:
I would tend to agree this is last minute. We don't have to cram everything into this release. To understand how much work this causes, what are the required code rewrites resulting from this PR?

What is the Pythonic approach?

Issue #158, opened 2025-09-24
VIewerTest is seg faulting


phapalova:
Happens time to time...
During fVDB CUDA 12.8.1 Build and Test:
https://github.com/openvdb/fvdb/actions/runs/17965030824/job/51096644805?pr=155
During fVDB Unit Tests:
https://github.com/openvdb/fvdb/actions/runs/17972701290/job/51119955103

PR #224, opened 2025-10-08
Viewer - proper data cleanup

phapalova:
- updates to nanovdb_editor 0.0.3 interface
- fixes data cleanup when maps are overwritten
- added rgb_rgb_rgb_sh proprety

Note: the PR currently points to the nanovdb-editor's develop branch, I will merge it to the main and update the hash just before merging this PR.

### DIFF
--- a/src/cmake/get_nanovdb_editor.cmake
+++ b/src/cmake/get_nanovdb_editor.cmake
@@ -13,7 +13,7 @@
 option(NANOVDB_EDITOR_FORCE "Force rebuild of nanovdb_editor wheel" OFF)
 option(NANOVDB_EDITOR_SKIP "Skip nanovdb_editor wheel build" OFF)
 
-set(NANOVDB_EDITOR_TAG f860667e6cc2e93d62aa6f580ca34d776f4b9b79)
+set(NANOVDB_EDITOR_TAG 60e6dafbb5a4eb7b6693bb3885578b314d139dea)
 set(NANOVDB_EDITOR_VERSION 0.0.3)
 
 # If skip is set, get the latest tagged version to prevent unnecessary rebuilds each hash update


src/tests/ViewerTest.cpp
```diff
@@ -18,38 +18,78 @@ TEST(Viewer, ViewerTest) {
     fvdb::detail::viewer::Viewer viewer = fvdb::detail::viewer::Viewer("127.0.0.1", 8080, false);
 
 #ifdef LOCAL_TESTING
```
fwilliams, line 20: We should include the PLY in the test data that we grab, and then run this test in CI. The test data is here: `voxel-foundation/fvdb-test-data` (send a PR).

CMake uses [this file](https://github.com/openvdb/fvdb-core/blob/main/src/cmake/get_test_data.cmake) to get test data and you'll need to update the hash.

fvdb/viz/_gaussian_splat_3d_view.py
```diff
@@ -91,6 +113,46 @@ def sh_degree_to_use(self, degree: int):
         """
         self._view.sh_degree_to_use = degree
 
+    @property
+    def sh_stride_rgb_rgb_rgb(self) -> bool:
```
fwilliams: For the public API, we may want to have an enum rather than boolean parameters. It's a bit confusing that setting one boolean changes the value of another. 

Perhaps something like:

```python
from enum import Enum

class ShOrderingMode(str, Enum):
    RGB_RGB_RGB = "rgb_rgb_rgb"
    RRR_GGG_BBB = "rrr_ggg_bbb"

[...]

@property
sh_ordering_mode(self) -> ShOrderingMode:
  if self._view.rrr_ggg_bbb:
    return ShOrderingModel.RRR_GGG_BBB
  elif self._view.rgb_rgb_rgb:
    return ShOrderingModel.RGB_RGB_RGB
  else:
    raise RuntimeError("Invalid configuration for sh ordering. This should not happen. File a bug")
```

src/tests/ViewerTest.cpp
phapalova: Created issue for the tests: https://github.com/openvdb/fvdb-core/issues/158

PR #230, opened 2025-10-09
Add mGPU Gaussian splatting operators

matthewdcong:
This PR adds multiGPU support for Gaussian splatting.
- The existing single GPU implementation are mostly unchanged aside from the addition to chunk indexing offsets in order to re-use the code for multiGPU. This is demonstrated by the single GPU tests passing without modification.
- Switching from `atomicAdd` vs `atomicAdd_system` for the single GPU case adds approximately 200 microseconds per iteration so it's relatively negligible. Hence, we use `atomicAdd_system` for the single GPU case as well in order to maximize code re-use.
- The `deviceOffsetAndCount` function has been renamed to `deviceChunk`. Furthermore, an aligned variant, `deviceAlignedChunk`, has been introduced. In addition, several places in the mGPU U-Net operators that were computing the chunks manually have been changed to use the `deviceChunk` function for consistency.
- GaussianRasterize operators have switched from using an explicitly separated block size of `(C, H, W)` to an equivalent packed block size of `(C * H * W, 1, 1)` with the per dimension index computation being done within the kernel. By collapsing the block index to a linear index, we simplify load balancing/chunking across GPUs especially for the case where the number of GPUs is not a power of two.

### DIFF
--- a/src/fvdb/detail/ops/gsplat/GaussianTileIntersection.cu
+++ b/src/fvdb/detail/ops/gsplat/GaussianTileIntersection.cu
@@ -655,6 +655,8 @@ radixSortAsync(KeyT *keysIn,
                ValueT *valuesIn,
                ValueT *valuesOut,
                NumItemsT numItems,
+               int beginBit,
+               int endBit,
                OffsetT *mergeIntervals,
                const OffsetT *offsets,
                const CountT *counts,
@@ -683,8 +685,8 @@ radixSortAsync(KeyT *keysIn,
                                         deviceValuesIn,
                                         deviceValuesOut,
                                         counts[deviceId],
-                                        0,
-                                        sizeof(KeyT) * 8,
+                                        beginBit,
+                                        endBit,
                                         stream);
         C10_CUDA_CHECK(cudaMallocAsync(&deviceTempStorage, tempStorageBytes, stream));
         cub::DeviceRadixSort::SortPairs(deviceTempStorage,
@@ -694,8 +696,8 @@ radixSortAsync(KeyT *keysIn,
                                         deviceValuesIn,
                                         deviceValuesOut,
                                         counts[deviceId],
-                                        0,
-                                        sizeof(KeyT) * 8,
+                                        beginBit,
+                                        endBit,
                                         stream);
         C10_CUDA_CHECK(cudaFreeAsync(deviceTempStorage, stream));
         C10_CUDA_CHECK(cudaEventRecord(events[deviceId], stream));
@@ -1102,11 +1104,14 @@ gaussianTileIntersectionPrivateUse1Impl(
                 deviceChunk(total_intersections, deviceId);
         }
 
+        const int32_t num_bits = 32 + num_cam_id_bits + num_tile_id_bits;
         radixSortAsync(intersection_keys.data_ptr<int64_t>(),
                        keys_sorted.data_ptr<int64_t>(),
                        intersection_values.data_ptr<int32_t>(),
                        vals_sorted.data_ptr<int32_t>(),
                        total_intersections,
+                       0,
+                       num_bits,
                        merge_intervals.data_ptr<int64_t>(),
                        device_intersection_offsets.const_data_ptr<int64_t>(),
                        device_intersection_counts.const_data_ptr<int64_t>(),


src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
```diff
@@ -61,44 +66,48 @@ computeGradientState(const uint32_t C,
...
+                         T *__restrict__ outDLossDCovars,            // [N, 6] optional
+                         T *__restrict__ outDLossDQuats,             // [N, 4] optional
+                         T *__restrict__ outDLossDScales,            // [N, 3] optional
+                         T *__restrict__ outDLossDWorldToCamMatrices // [C, 4, 4] optional
 ) {
     // parallelize over C * N.
-    const uint32_t idx = cg::this_grid().thread_rank();
-    if (idx >= C * N || radii[idx] <= 0) {
+    uint32_t idx = cg::this_grid().thread_rank();
+    if (idx >= count)
```
fwilliams: Braces

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
```diff
@@ -227,7 +236,7 @@ projectionBackwardKernel(
             outDLossDMeans += gId * 3;
             GSPLAT_PRAGMA_UNROLL
             for (uint32_t i = 0; i < 3; i++) {
-                gpuAtomicAdd(outDLossDMeans + i, dLossDPoint[i]);
+                atomicAdd_system(outDLossDMeans + i, dLossDPoint[i]);
```
fwilliams, line 240: Nice cleanup! Been meaning to do this.

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
```diff
@@ -355,36 +366,172 @@ dispatchGaussianProjectionBackward<torch::kCUDA>(
...
+        dLossDQuats  = torch::zeros_like(quats);
+        dLossDScales = torch::zeros_like(logScales);
+    }
+    torch::Tensor dLossDWorldToCamMatrices;
+    if (worldToCamMatricesRequiresGrad) {
+        dLossDWorldToCamMatrices = torch::zeros_like(worldToCamMatrices);
+    }
+    if (C && N) {
+        for (const auto deviceId: c10::irange(c10::cuda::device_count())) {
+            C10_CUDA_CHECK(cudaSetDevice(deviceId));
```
fwilliams, line 525: Can we refactor shared code out of the kCuda and kPrivateUse1 specializations?

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
```diff
@@ -420,8 +604,18 @@ dispatchGaussianProjectionBackward<torch::kCUDA>(
             if (outNormalizedMaxRadiiAccum.has_value()) {
                 int32_t *outNormalizedMaxRadiiAccumPtr =
                     outNormalizedMaxRadiiAccum.value().data_ptr<int32_t>();
-                computeGradientState<float, true>
-                    <<<GET_BLOCKS(N, DEFAULT_BLOCK_DIM), DEFAULT_BLOCK_DIM, 0, stream>>>(
+
+                for (const auto deviceId: c10::irange(c10::cuda::device_count())) {
```
fwilliams, line 609: Same question here. Seems like the kCuda backend is just calling this same code without the loop and with deviceProblemOffset and deviceProblemSize set to 0 and C*N

src/fvdb/detail/ops/gsplat/GaussianProjectionForward.cu
```diff
@@ -329,17 +337,125 @@ dispatchGaussianProjectionForward<torch::kCUDA>(
...
+
+    if (N == 0 || C == 0) {
+        // Early exit if there are no gaussians or cameras
+        return std::make_tuple(outRadii, outMeans2d, outDepths, outConics, outCompensations);
+    }
+
+    using scalar_t = float;
+
+    for (const auto deviceId: c10::irange(c10::cuda::device_count())) {
+        C10_CUDA_CHECK(cudaSetDevice(deviceId));
```
fwilliams, line 392: Again wondering how much of this can be refactored into a shared function.

src/fvdb/detail/ops/gsplat/GaussianRasterize.cuh
```diff
@@ -64,6 +64,7 @@ template <typename ScalarType, size_t NUM_CHANNELS, bool IS_PACKED> struct Raste
     using ScalarAccessor                           = TorchRAcc64<ScalarType, NUM_OUTER_DIMS>;
     using VectorAccessor                           = TorchRAcc64<ScalarType, NUM_OUTER_DIMS + 1>;
 
+    uint32_t mBlockOffset;
```
fwilliams, line 70: I know we don't really have docs for other things but we should start :)
Can we add a comment explaining this member and that in the single GPU case it's 0?

matthewdcong: Good idea, done!

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
```diff
@@ -410,8 +557,45 @@ dispatchGaussianProjectionBackward<torch::kCUDA>(
...
+                        covars.has_value() ? dLossDCovars.data_ptr<float>() : nullptr,
+                        covars.has_value() ? nullptr : dLossDQuats.data_ptr<float>(),
+                        covars.has_value() ? nullptr : dLossDScales.data_ptr<float>(),
+                        worldToCamMatricesRequiresGrad ? dLossDWorldToCamMatrices.data_ptr<float>()
+                                                       : nullptr);
+            }
+            C10_CUDA_KERNEL_LAUNCH_CHECK();
+        }
+
+        for (const auto deviceId: c10::irange(c10::cuda::device_count())) {
```
matthewdcong, line 597: E.g. this will be replaced by device-side logic where we create and enqueue events on each stream after each kernel launch. This will be followed by stream 0 waiting on all of the events (fan-in), followed by all of the streams waiting on stream 0 (fan-out) in order to create a barrier. This guarantees that the whole mGPU calculation from the prior kernel is complete before the next mGPU set of kernels executes without involving the host as we do currently.

PR #244, opened 2025-10-15
Switch from test to release index-url for CUDA 13 PyTorch

matthewdcong:
Now that PyTorch 2.9.0 has been released, we can use the official cu130 index url instead of the test one.


swahtz (approved):
Great, thanks for switching that

PR #252, opened 2025-10-17
Cleanup so JaggedTensor tests are linter-clean

blackencino:
The unit test for JaggedTensor, `tests/unit/test_jagged_tensor.py` had many linter failures. This fixes them.

### DIFF
--- a/fvdb/types.py
+++ b/fvdb/types.py
@@ -47,8 +47,13 @@ Index = int | slice | type(Ellipsis) | None
 
 GridIdentifier = str | int | list[str] | list[int] | tuple[str, ...] | tuple[int, ...]
 
-LShapeSpec = Iterable[int] | Iterable[Iterable[int]]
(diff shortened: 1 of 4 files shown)


fvdb/types.py
```diff
@@ -47,8 +47,13 @@
 
 GridIdentifier = str | int | list[str] | list[int] | tuple[str, ...] | tuple[int, ...]
 
-LShapeSpec = Iterable[int] | Iterable[Iterable[int]]
-RShapeSpec = Iterable[int]
+LShapeRank1 = Sequence[int]
```
fwilliams, line 50: Feel free to make this way better btw

fwilliams (approved):

fvdb/types.py
blackencino: I'm not sure what you mean? But okay!

PR #262, opened 2025-10-21
Dense API changed to Channels Major/Minor

blackencino:
Based on a more nuanced understanding of the convolution API, it is more correct and less confusing to keep the spatial dimensions unscrambled in conversion to and from dense tensors. The only variation we may need is for the channels to be in the major axes vs. the minor axes. As a shorthand, the two basic orderings are:

(B, X, Y, Z, C*) -> Channels Minor (C minor)  
(B, C*, X, Y, Z)  -> Channels Major (C major)

C minor is the "default" and aligns more closely with what fVDB expects based on the layout of jagged feature tensors. 

C major is the ordering expected by the convolution API, and aligns with what torch.nn.Conv3d and torch.nn.functional.conv3d expect.

These replace the former "XYZC" and "CZYX" naming conventions.

### DIFF
--- a/fvdb/utils/tests/__init__.py
+++ b/fvdb/utils/tests/__init__.py
@@ -111,7 +111,7 @@ def _clone_git_repo(git_url: str, git_tag: str, repo_name: str) -> Tuple[Path, g
             raise ValueError(f"A path {repo_path} exists but is not a git repo")
     else:
         repo = git.repo.Repo.clone_from(git_url, repo_path)
-    repo.remotes.origin.fetch()
+    repo.remotes.origin.fetch(tags=True)
     repo.git.checkout(git_tag)
 
     return repo_path, repo


fvdb/_Cpp.pyi
```diff
@@ -372,8 +372,8 @@ class GridBatch:
     def ray_implicit_intersection(
         self, ray_origins: JaggedTensor, ray_directions: JaggedTensor, grid_scalars: JaggedTensor, eps: float = 0.0
     ) -> JaggedTensor: ...
-    def read_from_dense_xyzc(self, dense_data: torch.Tensor, dense_origins: Vec3i | None = None) -> JaggedTensor: ...
-    def read_from_dense_czyx(self, dense_data: torch.Tensor, dense_origins: Vec3i | None = None) -> JaggedTensor: ...
+    def read_from_dense_cminor(self, dense_data: torch.Tensor, dense_origins: Vec3i | None = None) -> JaggedTensor: ...
```
fwilliams, line 375: 🎵🎶

PR #263, opened 2025-10-21
Viewer - add device id

phapalova:
Adding a constructor argument for device_id to optionally run viewer on a different device than default 0


fwilliams (changes requested):
This will break the python viewer. We should update `init` in `_viewer_server.py` to take this argument. It can be -1 by default which we can map to `cuda_device_id = torch.cuda.current_device()`

fwilliams (approved):

PR #277, opened 2025-10-23
More API docs

fwilliams:


docs/api/utils.rst
```diff
@@ -1,4 +1,11 @@
 fvdb.utils
 ==================
 
-.. autofunction:: fvdb.utils.build_ext.FVDBExtension
+.. autofunction:: fvdb.utils.fvdbCudaExtension
```
swahtz, line 4: I'm confused, I don't see this in `fvdb.utils`, as far as I see this is still called `FVDBExtension`.

swahtz: Oh it's renamed farther below, nm

fvdb/grid.py
```diff
@@ -170,16 +201,15 @@ def from_dense_axis_aligned_bounds(
     @classmethod
     def from_grid_batch(cls, grid_batch: "GridBatch", index: int = 0) -> "Grid":
         """
-        Create a single grid from one grid in a grid batch.
-        If the given grid batch is empty, the returned grid will be empty
-        with the same empty voxel size and origin.
+        Create a single grid from one grid in a :class:`GridBatch`. If ``index`` exceeds the number
```
swahtz: **💡 suggestion:** ‏  Do we want to use the word 'create' here?  Are we just passing the reference to the grid in the batch?  If so and we're not creating a new grid, should we say 'Return a single member grid from a `GridBatch`.'

```suggestion
        Return a single member grid from a :class:`GridBatch`. If ``index`` exceeds the number
```

fvdb/grid.py
```diff
@@ -170,16 +201,15 @@ def from_dense_axis_aligned_bounds(
...
+        of grids in the batch (minus one), an error is raised.
 
         Args:
-            grid_batch (GridBatch): The grid batch to create the grid from.
-            index (int): The index of the grid to create. Defaults to 0.
+            grid_batch (GridBatch): The :class:`GridBatch` to create the grid from.
```
swahtz:
```suggestion
            grid_batch (GridBatch): The :class:`GridBatch` to return the grid from.
```

fvdb/grid.py
```diff
@@ -230,22 +257,19 @@ def from_mesh(
         device: DeviceIdentifier | None = None,
     ) -> "Grid":
         """
-        Create a grid from a triangle mesh.
+        Create a grid by voxelizing a triangle mesh.
```
swahtz: **💡 suggestion:** ‏ I want to try and clarify what we mean by voxelizing the mesh, some people have thought we mean fill the interior space.

```suggestion
        Create a grid by voxelizing a triangle mesh.  Voxels will be created along the manifold defined by the faces of the mesh.
```

fvdb/grid.py
```diff
@@ -230,22 +257,19 @@ def from_mesh(
...
-                Defaults to None, which inherits from mesh_vertices.
+            mesh_vertices (torch.Tensor): Vertices of the mesh. A :class:`torch.Tensor` with shape ``(num_vertices, 3)``.
+            mesh_faces (torch.Tensor): Faces of the mesh. A :class:`torch.Tensor` with shape ``(num_faces, 3)``.
+            voxel_size (NumericMaxRank1): Size of each voxel, broadcastable to shape ``(3,)``, floating dtype
+            origin (NumericMaxRank1): Origin of the grid. *i.e.* the world-space position of the center of the ``[0,0,0]`` voxel, broadcastable to shape ``(3,)``, floating dtype
+            device (DeviceIdentifier | None): Device to create the grid on. Defaults to ``None``, which inherits from ``mesh_vertices``.
 
         Returns:
-            Grid: A new Grid object.
+            Grid: A new :class:`Grid` object with voxels covering the mesh.
```
swahtz:
```suggestion
            grid: A new :class:`Grid` object with voxels covering the mesh.
```

fvdb/grid.py
```diff
@@ -195,20 +225,17 @@ def from_ijk(
...
-            device (DeviceIdentifier | None): Device to create the grid on.
-                Defaults to None, which inherits from ijk.
+            ijk (torch.Tensor): Voxel coordinates to populate. A :class:`torch.Tensor` with shape ``(num_voxels, 3)`` with integer coordinates.
+            voxel_size (NumericMaxRank1): Size of each voxel, broadcastable to shape ``(3,)``, floating dtype
+            origin (NumericMaxRank1): Origin of the grid. *i.e.* the world-space position of the center of the ``[0,0,0]`` voxel, broadcastable to shape ``(3,)``, floating dtype
+            device (DeviceIdentifier | None): Device to create the grid on. Defaults to None, which inherits from ``ijk``.
 
         Returns:
-            Grid: A new Grid object.
+            Grid: A new :class:`Grid` object.
```
swahtz:
```suggestion
            grid: A new :class:`Grid` object.
```

fvdb/grid.py
```diff
@@ -267,20 +291,16 @@ def from_nearest_voxels_to_points(
         device: DeviceIdentifier | None = None,
     ) -> "Grid":
         """
-        Create a grid from the nearest voxels to a set of points.
+        Create a grid by adding the eight nearest voxels to every point in a point cloud.
```
swahtz, line 408: **💭 thought (non-blocking):** ‏ Not for this review, more of an issue we can create, but perhaps we can just add a `pad` argument to `from_points` which would accomplish the same thing and also be customizable.  `pad=1` is this function's equivalent, `pad=2` gives you another boundary layer beyond pad=1, and so on

fvdb/grid.py
```diff
@@ -267,20 +291,16 @@ def from_nearest_voxels_to_points(
...
-            device (DeviceIdentifier | None): Device to create the grid on.
-                Defaults to None, which inherits from points.
+            points (torch.Tensor): Points to populate the grid from. A :class:`torch.Tensor` with shape ``(num_points, 3)``.
+            voxel_size (NumericMaxRank1): Size of each voxel, broadcastable to shape ``(3,)``, floating dtype
+            origin (NumericMaxRank1): Origin of the grid, broadcastable to shape ``(3,)``, floating dtype
+            device (DeviceIdentifier | None): Device to create the grid on. Defaults to ``None``, which inherits from ``points``.
 
         Returns:
-            Grid: A new Grid object.
+            Grid: A new :class:`Grid` object.
```
swahtz:
```suggestion
            grid: A new :class:`Grid` object.
```

fvdb/grid.py
```diff
@@ -267,20 +291,16 @@ def from_nearest_voxels_to_points(
...
-            voxel_size (NumericMaxRank1): Size of each voxel,
-                broadcastable to shape (3,), floating dtype
-            origin (NumericMaxRank1): Origin of the grid,
-                broadcastable to shape (3,), floating dtype
-            device (DeviceIdentifier | None): Device to create the grid on.
-                Defaults to None, which inherits from points.
+            points (torch.Tensor): Points to populate the grid from. A :class:`torch.Tensor` with shape ``(num_points, 3)``.
+            voxel_size (NumericMaxRank1): Size of each voxel, broadcastable to shape ``(3,)``, floating dtype
+            origin (NumericMaxRank1): Origin of the grid, broadcastable to shape ``(3,)``, floating dtype
+            device (DeviceIdentifier | None): Device to create the grid on. Defaults to ``None``, which inherits from ``points``.
```
swahtz:
```suggestion
            device (DeviceIdentifier | None): Device to create the grid on. Defaults to ``None``, which inherits the device of ``points``.
```

fvdb/grid.py
```diff
@@ -230,22 +257,19 @@ def from_mesh(
...
-                broadcastable to shape (3,), floating dtype
-            origin (NumericMaxRank1): Origin of the grid,
-                broadcastable to shape (3,), floating dtype
-            device (DeviceIdentifier | None): Device to create the grid on.
-                Defaults to None, which inherits from mesh_vertices.
+            mesh_vertices (torch.Tensor): Vertices of the mesh. A :class:`torch.Tensor` with shape ``(num_vertices, 3)``.
+            mesh_faces (torch.Tensor): Faces of the mesh. A :class:`torch.Tensor` with shape ``(num_faces, 3)``.
+            voxel_size (NumericMaxRank1): Size of each voxel, broadcastable to shape ``(3,)``, floating dtype
+            origin (NumericMaxRank1): Origin of the grid. *i.e.* the world-space position of the center of the ``[0,0,0]`` voxel, broadcastable to shape ``(3,)``, floating dtype
+            device (DeviceIdentifier | None): Device to create the grid on. Defaults to ``None``, which inherits from ``mesh_vertices``.
```
swahtz:
```suggestion
            device (DeviceIdentifier | None): Device to create the grid on. Defaults to ``None``, which inherits the device of ``mesh_vertices``.
```

fvdb/grid.py
```diff
@@ -195,20 +225,17 @@ def from_ijk(
...
-            voxel_size (NumericMaxRank1): Size of each voxel,
-                broadcastable to shape (3,), floating dtype
-            origin (NumericMaxRank1): Origin of the grid,
-                broadcastable to shape (3,), floating dtype
-            device (DeviceIdentifier | None): Device to create the grid on.
-                Defaults to None, which inherits from ijk.
+            ijk (torch.Tensor): Voxel coordinates to populate. A :class:`torch.Tensor` with shape ``(num_voxels, 3)`` with integer coordinates.
+            voxel_size (NumericMaxRank1): Size of each voxel, broadcastable to shape ``(3,)``, floating dtype
+            origin (NumericMaxRank1): Origin of the grid. *i.e.* the world-space position of the center of the ``[0,0,0]`` voxel, broadcastable to shape ``(3,)``, floating dtype
+            device (DeviceIdentifier | None): Device to create the grid on. Defaults to None, which inherits from ``ijk``.
```
swahtz:
```suggestion
            device (DeviceIdentifier | None): Device to create the grid on. Defaults to None, which inherits the device of ``ijk``.
```

fvdb/grid.py
```diff
@@ -304,17 +324,13 @@ def from_points(
...
-            voxel_size (NumericMaxRank1): Size of each voxel,
-                broadcastable to shape (3,), floating dtype
-            origin (NumericMaxRank1): Origin of the grid,
-                broadcastable to shape (3,), floating dtype
-            device (DeviceIdentifier | None): Device to create the grid on.
-                Defaults to None, which inherits from points.
+            points (torch.Tensor): Points to populate the grid from. A :class:`torch.Tensor` with shape ``(num_points, 3)``.
+            voxel_size (NumericMaxRank1): Size of each voxel, broadcastable to shape ``(3,)``, floating dtype
+            origin (NumericMaxRank1): Origin of the grid, broadcastable to shape ``(3,)``, floating dtype
+            device (DeviceIdentifier | None): Device to create the grid on. Defaults to ``None``, which inherits from ``points``.
```
swahtz:
```suggestion
            device (DeviceIdentifier | None): Device to create the grid on. Defaults to ``None``, which inherits the device of ``points``.
```

fvdb/grid.py
```diff
@@ -304,17 +324,13 @@ def from_points(
...
 
         Returns:
-            Grid: A new Grid object.
+            Grid: A new :class:`Grid` object.
```
swahtz:
```suggestion
            grid: A new :class:`Grid` object.
```

fvdb/grid.py
```diff
@@ -334,25 +350,24 @@ def from_zero_voxels(
...
-            origin (NumericMaxRank1): Origin of the grid,
-                broadcastable to shape (3,), floating dtype
-                Defaults to 0.
+            device: The device to create the Grid on. Can be a string (e.g., "cuda", "cpu") or a :class:`torch.device` object. Defaults to ``"cpu"``.
+            voxel_size (NumericMaxRank1): Size of each voxel, broadcastable to shape ``(3,)``, floating dtype. Defaults to ``1``.
+            origin (NumericMaxRank1): Origin of the grid, broadcastable to shape ``(3,)``, floating dtype. Defaults to ``0``.
 
         Returns:
-            Grid: A new Grid object with zero voxels.
+            Grid: A new :class:`Grid` object with zero voxels.
```
swahtz:
```suggestion
            grid: A new :class:`Grid` object with zero voxels.
```

fvdb/grid.py
```diff
@@ -372,25 +387,33 @@ def avg_pool(
         coarse_grid: "Grid | None" = None,
     ) -> tuple[torch.Tensor, "Grid"]:
         """
-        Downsample grid data using average pooling.
+        Apply average pooling to the given data associated with this :class:`Grid` returned as data associated
```
swahtz, line 504:
```suggestion
        Apply average pooling to the given data associated with this :class:`Grid`.  Data returned is associated
```

fvdb/grid.py
```diff
@@ -372,25 +387,33 @@ def avg_pool(
...
+        with the given ``coarse_grid`` or a newly created coarse :class:`Grid`.
```
swahtz, line 505: **💭 thought (non-blocking):** ‏ Do we need to mention any checking that happens to the provided `coarse_grid` or what is needed for the `coarse_grid` to be correct for this operation?

fvdb/grid.py
```diff
@@ -469,26 +491,27 @@ def contiguous(self) -> "Grid":
...
         Returns:
-            Grid: A new Grid with contiguous memory layout.
+            grid (Grid): The same :class:`Grid` object.
         """
         return Grid(impl=self._impl.contiguous())
 
     def conv_grid(self, kernel_size: NumericMaxRank1, stride: NumericMaxRank1 = 1) -> "Grid":
         """
-        Return a grid representing the convolution of this grid with a given kernel.
+        Return a :class:`Grid` representing the active voxels at the output of a convolution applied this :class:`Grid` with a given kernel.
```
swahtz:
```suggestion
        Return a :class:`Grid` representing the active voxels at the output of a convolution applied to this :class:`Grid` with a given kernel.
```

fvdb/grid.py
```diff
@@ -469,26 +491,27 @@ def contiguous(self) -> "Grid":
...
 
         Args:
-            kernel_size (NumericMaxRank1): The size of the kernel to convolve with,
-                broadcastable to shape (3,), integer dtype
+            kernel_size (NumericMaxRank1): The size of the kernel to convolve with, broadcastable to shape ``(3,)``, integer dtype
             stride (NumericMaxRank1): The stride to use when convolving,
-                broadcastable to shape (3,), integer dtype
+                broadcastable to shape ``(3,)``, integer dtype
```
swahtz, line 625:
```suggestion
                broadcastable to shape ``(3,)``.  Integer dtype
```

fvdb/grid.py
```diff
@@ -469,26 +491,27 @@ def contiguous(self) -> "Grid":
...
 
     def conv_grid(self, kernel_size: NumericMaxRank1, stride: NumericMaxRank1 = 1) -> "Grid":
         """
-        Return a grid representing the convolution of this grid with a given kernel.
+        Return a :class:`Grid` representing the active voxels at the output of a convolution applied this :class:`Grid` with a given kernel.
 
         Args:
-            kernel_size (NumericMaxRank1): The size of the kernel to convolve with,
-                broadcastable to shape (3,), integer dtype
+            kernel_size (NumericMaxRank1): The size of the kernel to convolve with, broadcastable to shape ``(3,)``, integer dtype
```
swahtz, line 623:
```suggestion
            kernel_size (NumericMaxRank1): The 3D convolution kernel's total size, necessarily broadcastable to shape ``(3,)``.  Integer dtype.
```

fvdb/grid.py
swahtz: Wait, is stride a parameter broadcastable to shape (3,)?  This isn't correct, yes?

fvdb/grid.py
```diff
@@ -1044,44 +1104,45 @@ def is_contiguous(self) -> bool:
         """
         Check if the grid data is stored contiguously in memory.
 
+        .. note::
+
+            This method always returns ``True`` for grids with a single voxel. It exists
```
swahtz:
```suggestion
            This method always returns ``True`` for a single :class:`Grid`. It exists
```

fvdb/grid.py
```diff
@@ -1251,71 +1322,87 @@ def rays_intersect_voxels(
...
         return self._impl.read_from_dense_cmajor(dense_data.unsqueeze(0), dense_origin).jdata
 
     def sample_bezier(self, points: torch.Tensor, voxel_data: torch.Tensor) -> torch.Tensor:
         """
-        Sample voxel features at arbitrary points using Bézier interpolation.
+        Sample data in a :class:`torch.Tensor` associated with this :class:`Grid` at world-space
+        points using Bézier interpolation.
 
-        Interpolates voxel data at continuous world-space positions using cubic Bézier
+        This method interpolates data at voxel centers in continuous world-space positions using cubic Bézier
```
swahtz:
```suggestion
        This method interpolates voxel-centered data in continuous world-space positions using cubic Bézier
```

fvdb/grid.py
```diff
@@ -1325,22 +1412,40 @@ def sample_bezier_with_grad(
         self, points: torch.Tensor, voxel_data: torch.Tensor
     ) -> tuple[torch.Tensor, torch.Tensor]:
         """
-        Sample voxel features and their gradients using Bézier interpolation.
+        Sample data in a :class:`torch.Tensor` associated with this :class:`Grid` at world-space
+        points using Bézier interpolation, and return the spatial gradients at those points.
```
swahtz:
```suggestion
        points using Bézier interpolation, and return the samples with their spatial gradients at those points.
```

fvdb/grid.py
```diff
@@ -1325,22 +1412,40 @@ def sample_bezier_with_grad(
...
+
+        This method interpolates data at voxel centers in continuous world-space positions using cubic Bézier
```
swahtz:
```suggestion
        This method interpolates voxel-centered data in continuous world-space positions using cubic Bézier
```

fvdb/grid.py
```diff
@@ -1350,20 +1455,36 @@ def sample_bezier_with_grad(
 
     def sample_trilinear(self, points: torch.Tensor, voxel_data: torch.Tensor) -> torch.Tensor:
         """
-        Sample voxel features at arbitrary points using trilinear interpolation.
+        Sample data in a :class:`torch.Tensor` associated with this :class:`Grid` at world-space
+        points using trilinear interpolation.
+
+        This method interpolates data at voxel centers in continuous world-space positions using trilinear
```
swahtz:
```suggestion
        This method interpolates voxel-centered data in continuous world-space positions using trilinear
```

fvdb/grid.py
```diff
@@ -1374,22 +1495,39 @@ def sample_trilinear_with_grad(
         self, points: torch.Tensor, voxel_data: torch.Tensor
     ) -> tuple[torch.Tensor, torch.Tensor]:
         """
-        Sample voxel features and their gradients using trilinear interpolation.
+        Sample data in a :class:`torch.Tensor` associated with this :class:`Grid` at world-space
+        points using trilinear interpolation, and return the spatial gradients at those points.
+
+        This method interpolates data at voxel centers in continuous world-space positions using trilinear
```
swahtz:
```suggestion
        This method interpolates voxel-centered data in continuous world-space positions using trilinear
```

fvdb/grid.py
```diff
@@ -1800,60 +1953,173 @@ def write_to_dense_cmajor(
...
+        Returns:
+            voxel_size (torch.Tensor): A ``(3,)``-shaped tensor representing the size of each voxel.
+        """
         return self._impl.voxel_size_at(0)
 
     @property
     def world_to_grid_matrix(self) -> torch.Tensor:
+        """
+        The world-to-grid transformation matrix for this :class:`Grid`, which
+        transforms world space coordinates to voxel space coordinates.
```
swahtz, line 2354: **❓ question:** ‏ Do we want to use the word 'indices' when referring to voxel space coordinates?  If I remember correctly, this function won't return fractional voxel space indices, you'll get integer values, so should we refer to them as indices to help with the distinction?

fvdb/grid.py
fwilliams: There is a distinction here between this method and what a from_points with padding would do. 

Here the padding happens by adding the nearest voxels to a point. So if you have a voxel size of `(1, 1, 1)` and an origin of `(0, 0, 0)`, then a point `(0.75, 0.75, 0.1)` will add the voxels in the range `(0, 0, 0)` -> `(1, 1, -1)`.
If instead you added a point `(0.75, 0.75, 0.75)` you would add from `(0, 0, 0)` -> `(1, 1, 1)`

blackencino (approved):
LGTM!

fvdb/grid.py
fwilliams: there is no checking! you can pass anything in and this function will inject the data into the tensor associated with the coarse grid. If the voxel being injected to is not there, then nothing gets copied. This gives you the freedom to set up your coarse domain however you see fit. In the common case, you can use this to create the grid you want (and I'll add a reference to coarsened_grid).

fvdb/grid.py
fwilliams: should be yeah, the scalar (1,) is broadcastable to (1, ,1, 1)

fvdb/grid.py
```diff
@@ -579,60 +701,68 @@ def cubes_intersect_grid(
...
+                the primal grid bounds. Default is ``False``.
 
         Returns:
-            Grid: A new Grid representing the dual grid.
+            grid (Grid): A new :class:`Grid` representing the dual grid.
         """
         return Grid(impl=self._impl.dual_grid(exclude_border))
 
-    def grid_to_world(self, ijk: torch.Tensor) -> torch.Tensor:
+    def voxel_to_world(self, ijk: torch.Tensor) -> torch.Tensor:
```
swahtz, line 738: Why change the name of this function to use `voxel` instead of `grid`?  If we're up for changing the name of it, shouldn't we be calling this function `index_to_world` since it goes from index space to world space and this is the terminology used:

https://www.openvdb.org/documentation/doxygen/overview.html#secSpaceAndTrans

fvdb/grid.py
swahtz: This sentence I find very confusing because of the '_at_ voxel centers _in_ continuous world-space positions'. How about

This method uses Bézier interpolation to interpolate data values at arbitrary continuous positions in world space, based on values defined at voxel centers.

fvdb/grid.py
swahtz: Same as above

fvdb/grid.py
swahtz: Same as above

fvdb/grid.py
swahtz: Same as above

fvdb/grid.py
fwilliams: I like this

fvdb/grid.py
swahtz: Do you really mean 'single voxel' here?

swahtz (changes requested):
Some minor things and a few little corrections but I think my only big question is why rename `grid_to_world` to `voxel_to_world` as well as other places where we use `grid space` in this way?  

Strictly it should be 'index' space but I know we mostly use the word 'index' to refer to the offset into the sidecar because of 'IndexGrid'.

But regardless of that, why change it, what's the advantage?  I don't think 'voxel space' is a more clear reference to 'index space' than 'grid space' is.

fvdb/grid.py
fwilliams: the docs at the top of the class explain this. Index is used to refer to the linear offsets in a tensor associated with the grid. This also conforms to our use of ijk_to_index and ijk_to_inv_index

fvdb/grid.py
fwilliams: changed the docs for this

PR #278, opened 2025-10-23
Switch SH accessors from 32-bit to 64-bit

matthewdcong:
As the number of Gaussian splats becomes large (55M+ in my case with degree 3) , the 32-bit accessor becomes insufficient and throws the following error in the SH evaluation code. Switch to 64-bit accessors to address the issue.

`RuntimeError: numel needs to be smaller than int32_t max; otherwise, please use packed_accessor64`

2^32 / (55M * sizeof(float) * 16 SH DOFs) is very close to 1


swahtz (approved):
lgtm!

PR #281, opened 2025-10-23
GridBatch API Docs Updates

swahtz:
This is a large update to the docstrings in `grid_batch.py` based largely on docstring changes in `grid.py`.  It includes a couple of minor fixes to docstrings `grid.py` and it removes all the docstrings written into the `GridBatchBindings.cpp` because none of those get exposed to the end user in the current organization.

### DIFF
--- a/fvdb/grid_batch.py
+++ b/fvdb/grid_batch.py
@@ -139,7 +139,7 @@ class GridBatch:
 
     """
 
-    # Class variable
+    #: :meta private: # NOTE: This is here for sphinx to not complain that the attribute is double defined in the class and in the class documentation.
     max_grids_per_batch: int = GridBatchCpp.max_grids_per_batch
 
     def __init__(self, *, impl: GridBatchCpp):


fvdb/grid.py
```diff
@@ -66,7 +66,7 @@ class Grid:
     which a grid can be used to index into. This also allows multiple grids to share the same data
     storage if desired.
 
-    When using a :class:`Grid`, voxel coordinates, there are three important coordinate systems to be aware of:
+    When using a :class:`Grid`'s voxel coordinates, there are three important coordinate systems to be aware of:
```
fwilliams, line 70: check that this doesn't throw a warning when renedrering docs with sphinx-build. I had to escale the quote if I recall correctly

fvdb/grid_batch.py
```diff
@@ -2,39 +2,39 @@
...
-- from_empty: Create one or more grids with zero voxels.
-- from_dense: Create from dense grid dimensions
-- from_ijk: Create from explicit voxel coordinates
-- from_mesh: Create from triangle meshes
-- from_points: Create from point clouds
-- from_nearest_voxels_to_points: Create from nearest voxels to points
-
-Module-level functions for loading and saving grid batches:
-- load_gridbatch/save_gridbatch: Load and save grid batches to/from .nvdb files
+- :meth:`GridBatch.from_zero_grids()`: for an empty grid batch where grid-count = 0.
```
fwilliams, line 14: rst list likely won't render without a newline before it here

fvdb/grid_batch.py
```diff
@@ -61,43 +59,89 @@ class GridBatch:
...
-        - GridBatch.from_points() for point clouds
-        - GridBatch.from_zero_grids() for zero grids
-        - GridBatch.from_zero_voxels() for one or more empty grids (zero voxels)
-
-        The GridBatch constructor is for internal use only, always use the classmethods.
+    sampling, convolution, pooling, dilation, union, etc. It also provides more advanced features
+    such as marching cubes, TSDF fusion, and fast ray marching.
+
+    A :class:`GridBatch` can be thought of as a collection of :class:`Grid` instances and,
+    like the :class:`Grid`, does not collect the sparse voxel grids' data but only collects
```
fwilliams, line 70: we use collection as a noun to refer to a batch of grids, then use collect as a verb to mean contain voxels which is a bit confusing. Maybe we can use the word batch rather than collection since this is a well understood concept.

fvdb/grid_batch.py
```diff
@@ -61,43 +59,89 @@ class GridBatch:
...
+
+    When using a :class:`GridBatch`'s voxel coordinates, there are three important coordinate systems
+    to be aware of:
+
+    - **World Space**: The continuous 3D coordinate system in which each grid in the batch exists.
+    - **Voxel Space**: The discrete voxel index system of each grid in the batch, where each voxel is identified by its integer indices (i, j, k).
+    - **Index Space**: The linear indexing of active voxels in each grid's internal storage.
+
+
+    At its core, a :class:`Grid` uses a very fast mapping from voxel space into index space to perform operations on a :class:`torch.Tensor` of
```
fwilliams: Grid -> GridBatch and tensor -> JaggedTensor

fvdb/grid_batch.py
```diff
@@ -61,43 +59,89 @@ class GridBatch:
...
+
+    A :class:`GridBatch` can be thought of as a collection of :class:`Grid` instances and,
+    like the :class:`Grid`, does not collect the sparse voxel grids' data but only collects
+    their structure (or topology). Voxel data (e.g., features, colors, densities) for the
+    collection of grids is stored separately as an :class:`JaggedTensor` associated with
+    the :class:`GridBatch`. This separation allows for flexibility in the type and number of
+    channels of data with which a grid can be used to index into. This also allows multiple grids to
+    share the same data storage if desired.
+
+    When using a :class:`GridBatch`'s voxel coordinates, there are three important coordinate systems
```
fwilliams: voxel coordinates is confusing in this sentence since we define voxel space below. Just say, When using GridBatch, there are three important coordinate systems to be aware of.

fvdb/grid_batch.py
```diff
@@ -117,36 +161,56 @@ def from_dense(
         device: DeviceIdentifier | None = None,
     ) -> "GridBatch":
         """
-        A dense grid has a voxel for every coordinate in an axis-aligned box of Vec3,
-        which can in turn be mapped to a world-space box.
+        Create a batch of dense grids.
+
+        A dense grid has a voxel for every coordinate in an axis-aligned box.
+
+        For each grid in the batch, the dense grid is defined by:
```
fwilliams, line 171: Newline so list will render correctly

fvdb/grid_batch.py
```diff
@@ -207,20 +323,33 @@ def from_ijk(
...
         Args:
-            ijk (torch.Tensor): Voxel coordinates to populate.
-                Shape: (num_voxels, 3) with integer coordinates.
-            voxel_size (NumericMaxRank1): Size of each voxel,
-                broadcastable to shape (3,), floating dtype
-            origin (NumericMaxRank1): Origin of the grid,
-                broadcastable to shape (3,), floating dtype
-            device (DeviceIdentifier | None): Device to create the grid on.
-                Defaults to None, which inherits from ijk.
+            ijk (JaggedTensor): Per-grid voxel coordinates to populate. Shape: ``(batch_size, num_voxels, 3)``
```
fwilliams: for jagged tensor notation, label the variable axis as something variable so like `(batch_size, num_voxels_for_grid_b, 3)`

fvdb/grid_batch.py
```diff
@@ -417,20 +584,32 @@ def from_points(
         device: DeviceIdentifier | None = None,
     ) -> "GridBatch":
         """
-        Create a grid batch from a point cloud.
+        Create a batch of grids from point clouds.
```
fwilliams: "a batch of point clouds"?

fvdb/grid_batch.py
```diff
@@ -417,20 +584,32 @@ def from_points(
...
         Args:
-            points (JaggedTensor): Points to populate the grid from, per-grid
-                Shape: (batch_size, num_points, 3).
-            voxel_sizes (NumericMaxRank2): Size of each voxel, per-grid
-                broadcastable to shape (batch_size, 3), floating dtype
-            origins (NumericMaxRank2): Origin of the grid, per-grid
-                broadcastable to shape (num_grids, 3), floating dtype
-            device (DeviceIdentifier | None): Device to create the grid on.
-                Defaults to None, which inherits from points.
+            points (JaggedTensor): Per-grid point positions to populate the grid from. A :class:`fvdb.JaggedTensor` with shape ``(batch_size, num_points, 3)``.
```
fwilliams: num_points_b?

fvdb/grid_batch.py
```diff
@@ -756,8 +957,8 @@ def dual_bbox_at(self, bi: int) -> torch.Tensor:
             bi (int): The batch index of the grid.
 
         Returns:
-            torch.Tensor: A tensor of shape (2, 3) containing the minimum and maximum
-                coordinates of the dual bounding box in index space.
+            dual_bbox (torch.Tensor): A tensor of shape ``(2, 3)`` containing the minimum and maximum
```
fwilliams, line 968: in this docstring, probably

```

.. seealso::

    :meth:`dual_grid` to copute the actual dual grid

File: fvdb/grid_batch.py [hunk 11]
```diff
@@ -862,26 +1064,36 @@ def inject_from(
         default_value: float | int | bool = 0,
     ) -> JaggedTensor:
         """
-        Inject data from the source grid to this grid.
-        This method copies sidecar data for voxels in the source grid to a sidecar corresponding to voxels in this grid.
+        Inject data associated with the source grid batch to a :class:`fvdb.JaggedTensor`
```
fwilliams, line 1075: Here we use inject, and in inject_from/to_dense we changed it write. Let's keep inject in the docs above. We'll have one definition of what it means to inject which will make this a well defined verb in the future

fvdb/grid_batch.py
```diff
@@ -1180,26 +1385,32 @@ def max_pool(
...
+        :class:`GridBatch` and the given ``coarse_grid``.
+
+        .. note::
+
+            If you pass ``coarse_grid = None``, the returned coarse grid batch will have its
+            voxel sizes multiplied by the ``pool_factor`` and origins adjusted accordingly.
+
+        .. note::
+
+            This method supports backpropagation through the pooling operation.
```
fwilliams, line 1419: haven't been checking but we should make sure we have this everywhere relevant

fvdb/grid.py
swahtz: All good, no warning

fvdb/grid_batch.py
swahtz: `grid.py` is missing this newline too, I'll insert it in both.  But actually neither of these sections in either files is rendered into the docs when building.  I assume that's probably because it's not part of the class itself?

fvdb/grid_batch.py
swahtz: Yeah I generally don't like using a word in its definition so went with 'collection'.  I'll switch it out for mini-batch since that's the original nomenclature, clarifies it a bit.

fvdb/grid_batch.py
swahtz: Ah copy paste issue

fvdb/grid_batch.py
swahtz: Done

fvdb/grid_batch.py
swahtz: I checked and there were 2 missing this note, all fixed.

PR #286, opened 2025-10-24
Fix PyTorch index url for package

matthewdcong:
The default PyTorch 2.8.0 package uses CUDA 12.8. We need to specify the PyTorch CUDA 12.9 index URL in order to pick up the right version.

This also switches the dev dockerfile to CUDA 12.9 in order to match the package.


swahtz (approved):
lgtm

harrism (approved):

PR #307, opened 2025-10-27
shared memory types in IntegrateTSDF kernel

swahtz:
Host code computes the dynamic shared-memory size with `nanovdb::math::Mat3<scalar_t>` and `Mat4<scalar_t>`, but inside the kernel those matrices are instantiated with `ScalarType = OpType<scalar_t>::type`. For the c10::Half dispatch this means the kernel stores Mat3<float>/Mat4<float> (36/64 bytes each) while the launch only reserves enough space for Mat3<half>/Mat4<half> (18/32 bytes). On Blackwell that mismatch shows up as the out-of-bounds shared write.


matthewdcong:
Looks good, thanks @swahtz

PR #313, opened 2025-10-30
Rename _Cpp python extension library binary to _fvdb_cpp, include pybind11 headers first

### LINKED ISSUE #130: Rename fvdb bindings `_Cpp` binary/module, and make sure pybind11.h header is always first include.
Suggestions from @rwgk:

1. `_Cpp` could in theory clash with other `_Cpp` extensions. This is because the C symbols imported by Python are not fully qualified.

```
rwgk-win11.localdomain:~/fvdb $ nm ./fvdb/build/cp312-cp312-linux_x86_64-Release/_Cpp.so | grep PyInit__Cpp
0000000000051fb0 T PyInit__Cpp
000000000001a628 t PyInit__Cpp.cold
rwgk-win11.localdomain:~/fvdb $
```

I recommend making the name longer and more unique.

2. https://pybind11.readthedocs.io/en/stable/basics.html#header-and-namespace-conventions — pybind11/pybind11.h includes Python.h, as such it must be [the first file included in any source file or header for the same reasons as Python.h](https://docs.python.org/3/extending/extending.html#a-simple-example).

Both are totally unrelated to the ABI issue (almost certainly at least), but something I'd take care of at an opportune moment.

harrism:
Fixes #130 

Renames the python extension library target to _fvdb_cpp from _Cpp to avoid collisions with other libraries.
Always includes pybind11 headers first (and updates clang-format ordering rule).

### DIFF
--- a/src/python/TypeCasters.h
+++ b/src/python/TypeCasters.h
@@ -4,12 +4,12 @@
 #ifndef PYTHON_TYPECASTERS_H
 #define PYTHON_TYPECASTERS_H
 
+#include <pybind11/numpy.h>
+#include <pybind11/stl.h>
+
 #include <fvdb/Types.h>
 
 #include <torch/extension.h>
-
-#include <pybind11/numpy.h>
-#include <pybind11/stl.h>
 
 namespace pybind11 {
 namespace detail {
--- a/.clang-format
+++ b/.clang-format
@@ -117,6 +117,8 @@ AllowShortLoopsOnASingleLine: false
 IncludeBlocks: Regroup
 IncludeIsMainRegex: "$"
 IncludeCategories:
+  - Regex:           '^<pybind11/' # pybind11 includes
+    Priority:        0
   - Regex:           '^"' # quoted includes
     Priority:        1
   - Regex:           '^<fvdb/' # fvdb includes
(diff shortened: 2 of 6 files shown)


src/python/TypeCasters.h
```diff
@@ -4,13 +4,13 @@
 #ifndef PYTHON_TYPECASTERS_H
 #define PYTHON_TYPECASTERS_H
 
+#include <pybind11/numpy.h>
```
matthewdcong, line 7: Thanks Mark! I think this is a good change but I think we can make it more maintainable long-term by removing the pybind11 includes from the .cpp files entirely. I think these two lines in TypeCasters.h will transitively provide the necessary pybind11 includes.

stl.h -> pybind11.h -> typing.h -> cast.h
numpy.h -> pytypes.h

harrism: Transitive includes are generally considered bad practice and IWYU is generally considered good practice. So I'm of two minds about your suggestion. On one hand it could reduce maintenance. On the other it could introduce future problems.

harrism: Also if I remove the IWYU includes from the .cpp files, I get clangd warnings about TypeCasters.h include not being directly used.

matthewdcong: Thanks for giving it a try, we can keep it as is.

matthewdcong (approved):

PR #315, opened 2025-11-03
Fix radix sort prefetch for disjoint sorts

matthewdcong:
When merging sorted arrays across GPUs, it's possible that all of the elements on the first device are less than all of the elements on the second device. This means that the second GPU needs to prefetch no elements from the first device and the first GPU needs to prefetch no elements from the second device. In this case, `rightCount` and `leftCount` are zero for the left and right device respectively and we need to check for that before calling `memPrefetchAsync` to avoid a `cudaErrorInvalidValue` failure.


sifakis (approved):
Looks perfectly reasonable

PR #319, opened 2025-11-04
Make nightly workflows not run on forks of openvdb/fvdb-core

harrism:
Nightly workflows were running on my fork, but failing after 24 hours of waiting because my fork can't get runners. This PR adds a check to prevent the nightly workflow from running on a fork even if the fork has actions enabled.


matthewdcong (approved):
Looks good to me, thanks Mark

PR #325, opened 2025-11-08
[Bug Fix] Improve binary search handling in JIdxForJOffsets

iYuqinL:
Refactor binary search logic to handle cases with duplicate offsets correctly.

The original jidx from joffsets binary search can not handle the case that when joffsets has same values.
For example, joffsets is [0, 10, 10, 40], when idx = 10, the original code will wrongly set jidx to -1.

### DIFF
--- a/src/tests/JaggedTensorTest.cpp
+++ b/src/tests/JaggedTensorTest.cpp
@@ -101,6 +101,80 @@ TEST_F(JaggedTensorTest, LSizesConstructor) {
     EXPECT_EQ(computed_lsizes[2], 1);
 }
 
+// Test jidx_from_joffsets, JaggedTensor on CPU
+TEST_F(JaggedTensorTest, JIdxFromJOffsetsCPU) {
+    auto device = torch::kCPU;
+
+    { // case 1: all tensor are not empty
+        std::vector<int64_t> lsizes = {3, 2, 1};
+        auto data                   = torch::tensor({1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 6.0f},
+                                  torch::TensorOptions().dtype(torch::kFloat32).device(device));
+
+        JaggedTensor jt(lsizes, data);
+        auto jidx     = jt.jidx_from_joffsets(jt.joffsets(), data.size(0));
+        auto jidx_cpu = jidx.cpu().to(torch::kInt64);
+
+        std::vector<int64_t> expected_vec = {0, 0, 0, 1, 1, 2};
+        auto expected                     = torch::tensor(
+            expected_vec, torch::TensorOptions().dtype(torch::kInt64).device(torch::kCPU));
+
+        EXPECT_TRUE(torch::equal(jidx_cpu, expected));
+    }
+
+    { // case 2: lsizes with zeros
+        std::vector<int64_t> lsizes2 = {0, 3, 0, 0, 2, 1, 0};
+        auto data2                   = torch::tensor({1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 6.0f},
+                                   torch::TensorOptions().dtype(torch::kFloat32).device(device));
+
+        JaggedTensor jt2(lsizes2, data2);
+        auto jidx2     = jt2.jidx_from_joffsets(jt2.joffsets(), data2.size(0));
+        auto jidx2_cpu = jidx2.cpu().to(torch::kInt64);
+
+        std::vector<int64_t> expected_vec2 = {1, 1, 1, 4, 4, 5};
+        auto expected2                     = torch::tensor(
+            expected_vec2, torch::TensorOptions().dtype(torch::kInt64).device(torch::kCPU));
+
+        EXPECT_TRUE(torch::equal(jidx2_cpu, expected2));
+    }
+}
+
+// Test jidx_from_joffsets, JaggedTensor on CUDA
+TEST_F(JaggedTensorTest, JIdxFromJOffsetsCUDA) {
+    auto device = torch::kCUDA;
+
+    { // case 1: all tensor are not empty
+        std::vector<int64_t> lsizes = {3, 2, 1};
+        auto data                   = torch::tensor({1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 6.0f},
+                                  torch::TensorOptions().dtype(torch::kFloat32).device(device));
+
+        JaggedTensor jt(lsizes, data);
+        auto jidx     = jt.jidx_from_joffsets(jt.joffsets(), data.size(0));
+        auto jidx_cpu = jidx.cpu().to(torch::kInt64);
+
+        std::vector<int64_t> expected_vec = {0, 0, 0, 1, 1, 2};
+        auto expected                     = torch::tensor(
+            expected_vec, torch::TensorOptions().dtype(torch::kInt64).device(torch::kCPU));
+
+        EXPECT_TRUE(torch::equal(jidx_cpu, expected));
+    }
+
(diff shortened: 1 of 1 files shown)


src/fvdb/detail/ops/JIdxForJOffsets.cu
```diff
@@ -28,21 +28,36 @@ jIdxForJOffsets(TorchRAcc32<fvdb::JOffsetsType, 1> offsets,
...
+        // if (idx >= offsets[mid] && idx < offsets[mid + 1]) {
+        //     outJIdx[idx] = mid;
+        //     return;
+        // }
+        // the above check can not handle the following case:
+        // offsets = [0, 10, 10, 40], idx=10, mid=1.
+        ///////////////////////////////////////////////////////////////
+
+        // the bin-search should consider when there are some value in offsets
+        if (offsets[mid] <= idx){
```
fwilliams: A few small formatting fixes: 
```C++
if (offsets[mid] <= idx) { // space after {
```

src/fvdb/detail/ops/JIdxForJOffsets.cu
```diff
@@ -28,21 +28,36 @@ jIdxForJOffsets(TorchRAcc32<fvdb::JOffsetsType, 1> offsets,
...
+            // if offsets[mid] <= idx, just means possible.
+            // the case [0, 10, 10, 40], when idx=10, mid=1,
+            // what we need is 2, so we need to let left = mid+1
+            if ((mid < (offsets.size(0)-1)) &&
+                (idx < offsets[mid+1])){
+                outJIdx[idx] = mid;
+                return;
+            }
+            else
```
fwilliams:
```
} else {
```

src/fvdb/detail/ops/JIdxForJOffsets.cu
```diff
@@ -28,21 +28,36 @@ jIdxForJOffsets(TorchRAcc32<fvdb::JOffsetsType, 1> offsets,
...
+        // offsets = [0, 10, 10, 40], idx=10, mid=1.
+        ///////////////////////////////////////////////////////////////
+
+        // the bin-search should consider when there are some value in offsets
+        if (offsets[mid] <= idx){
+            // if offsets[mid] <= idx, just means possible.
+            // the case [0, 10, 10, 40], when idx=10, mid=1,
+            // what we need is 2, so we need to let left = mid+1
+            if ((mid < (offsets.size(0)-1)) &&
+                (idx < offsets[mid+1])){
```
fwilliams: space before `{`

src/fvdb/detail/ops/JIdxForJOffsets.cu
```diff
@@ -28,21 +28,36 @@ jIdxForJOffsets(TorchRAcc32<fvdb::JOffsetsType, 1> offsets,
...
-            outJIdx[idx] = mid;
-            return;
-        }
-
-        if (offsets[mid] < idx) {
-            // If key greater, ignore left half
-            left = mid + 1;
-        } else {
-            // If key is smaller, ignore right half
+        // if (idx >= offsets[mid] && idx < offsets[mid + 1]) {
```
fwilliams: No need to keep the commented out old code, you can remove it h ere

src/fvdb/detail/ops/JIdxForJOffsets.cu
harrism: @fwilliams @iYuqinL formatting like this should be automatic with clang-format. PR should fail. 

@iYuqinL please run clang-format on the changed files. (I have my editor set up to format on save.) The .clang-format config file will automatically be applied.

iYuqinL:
@fwilliams @harrism hello, I have formatted the code using clang-format as you suggested.
And I have also added a unit test function to tests/unit/test_jagged_tensor.py, which named `test_jidxforjoffsets`.

tests/unit/test_jagged_tensor.py
```diff
@@ -2574,6 +2574,22 @@ def test_round(self, device, dtype):
         jt1.round_()
         self.assertTrue(torch.allclose(jt2.jdata, jt1.jdata))
 
+    @parameterized.expand(all_device_dtype_combos)
+    def test_jidxforjoffsets(self, device, dtype):
```
harrism, line 2578: Does this test actually call the kernel you changed? I can't follow the plumbing through.

I would like to make sure this exercises the fix.  Does this test fail without your fix to the kernel?

Also, since we have C++ unit tests for JaggedTensor that exercise this kernel via calls to `JaggedTensor::jidx_from_joffsets()` (e.g. [here](https://github.com/openvdb/fvdb-core/blob/8fe92dfecce15c41bbaf98c66cd890cde67831c8/src/tests/JaggedTensorTest.cpp#L791)), please add a test case to that gtest that also exercises repeated values.

Thanks!

harrism (changes requested):
Thanks so much for the great contribution! Can you please add a C++ test? Should only need a couple lines in an existing test.

tests/unit/test_jagged_tensor.py
iYuqinL: Yes, it does call the kernel I modified. 
I've actually tested both before and after the modification; the version before the modification failed the unit tests.

In fact, the kernel function I modified was called when JaggedTensor initialization. 
in src code file: `src/fvdb/JaggedTensor.cpp`, line 258, 319, 511, etc.

```c
ret.mBatchIdx      = jidx_from_joffsets(offsets, data.size(0));
```

src/tests/JaggedTensorTest.cpp
```diff
@@ -101,6 +101,80 @@ TEST_F(JaggedTensorTest, LSizesConstructor) {
     EXPECT_EQ(computed_lsizes[2], 1);
 }
 
+// Test jidx_from_joffsets, JaggedTensor on CPU
+TEST_F(JaggedTensorTest, JIdxFromJOffsetsCPU) {
```
harrism, line 105: **👏 praise:** ‏Thank you for adding these excellent gtests!

src/tests/JaggedTensorTest.cpp
```diff
@@ -101,6 +101,80 @@ TEST_F(JaggedTensorTest, LSizesConstructor) {
...
+        std::vector<int64_t> expected_vec2 = {1, 1, 1, 4, 4, 5};
+        auto expected2                     = torch::tensor(
+            expected_vec2, torch::TensorOptions().dtype(torch::kInt64).device(torch::kCPU));
+
+        EXPECT_TRUE(torch::equal(jidx2_cpu, expected2));
+    }
+}
+
+// Test jidx_from_joffsets, JaggedTensor on CUDA
+TEST_F(JaggedTensorTest, JIdxFromJOffsetsCUDA) {
```
harrism, line 142: **💡 suggestion (non-blocking):** ‏These tests are identical except in the device parameter. I won't block merging this PR unless you want to do it, but it would be nice to use parameterized tests for this.

harrism (approved):
Thank you for this excellent PR! We really appreciate your contribution @iYuqinL.

src/tests/JaggedTensorTest.cpp
iYuqinL: I didn't see any examples of parameterized unit tests in the cpp unit tests. 
I consulted relevant materials and asked chatgpt. It seems that adding a parameterized unit test wouldn't make things much simpler. Therefore, I decided to keep the original unit tests. Perhaps it would be a better choice to refactor them later when there are more unit tests that need parameterization.

PR #328, opened 2025-11-12
Fix documentation for setting TORCH_CUDA_ARCH_LIST

matthewdcong:
1. Quotes are required for the `TORCH_CUDA_ARCH_LIST` so bash doesn't only pick up the first arch. Add that to the `-h` printout for `./build.sh`.
2. Update the README to use `--cuda-arch-list` instead of setting `TORCH_CUDA_ARCH_LIST`.


harrism (approved):
Thanks for the fix! If it's not hard, it might also be good to support the ENV variable. That way one can set it in their shell so it can be reused.

PR #329, opened 2025-11-13
Fix GHA error in (currently unused) nightly workflows

harrism:
To disable these two nightly workflows, the entire "on:" section was commented out. That is an error since a GHA workflow must have an "on:". This PR just uncomments "on:" and "workflow_dispatch".

(I was seeing daily errors from these on my fork because I had actions enabled. Not sure why they don't cause errors on the main repo.)


swahtz (approved):
Great, thanks!

PR #332, opened 2025-11-14
Add viz bindings for wait and add_image

phapalova:
Viewer function for adding and updating the 2D image:
```
import fvdb

fvdb.viz.init(ip_address="127.0.0.1", port=8080)
scene = fvdb.viz.Scene("My Scene")

width, height = 512, 512
image_data = torch.zeros((height, width, 4), dtype=torch.uint8

for j in range(height):
    for i in range(width):
        image_data[j, i, 0] = (255 * i) // (width - 1)  # R
        image_data[j, i, 1] = (255 * j) // (height - 1)  # G
        image_data[j, i, 2] = 0  # B
        image_data[j, i, 3] = 255  # A

image_view = scene.add_image(
    name="test_image",
    rgba_image=image_data.flatten(),
    width=width,
    height=height,
)
fvdb.viz.show()
fvdb.viz.wait_for_interrupt()
...
image_view.update(new_image_data)
```
`wait_for_interrupt()` replaces the `time.sleep(10000)` which was needed to keep the main thread running so the viewer is not destroyed.

image

### DIFF
--- a/fvdb/viz/_image_view.py
+++ b/fvdb/viz/_image_view.py
@@ -119,4 +119,3 @@ class ImageView:
 
         server = _get_viewer_server_cpp()
         server.add_image(self._scene_name, self._name, rgba_tensor, self._width, self._height)
-


areidmeyer (approved):
Looks good to me.
One question/request. Have we exposed add_nanovdb() itself properly yet?
Jonathan did say we now had a customer ask for it.
If not, probably should be another PR anyways, since in practice we probably have to make it all work nicely with GridBatch.

PR #336, opened 2025-11-17
Fixes for `fvdb.nn.SimpleUNet`

### LINKED ISSUE #335: `fvdb.nn.ReLU` usage in `fvdb.nn.simple_unet`
`fvdb.nn.ReLU` is being used in `fvdb.nn.simple_unet` but these activation function modules were removed.  `SimpleUNet` needs to be updated to use `torch.nn.ReLU`

https://github.com/openvdb/fvdb-core/blob/main/fvdb/nn/simple_unet.py#L90

swahtz:
While working on fixing `fvdb.nn.SimpleUNet` for the issues reported in #335 where the previously removed `fvdb.nn.ReLU` module was being used, a few other issues were discovered.  This PR proposes to fix the following:

1. In `SimpleUNet`, replace the use of `fvdb.nn.ReLU(inplace=True)` with the functional `fvdb.relu_`
2. Remove an erroneous check in `SparseConvPackInfo::SparseConvPackInfo` that threw an exception if a kmap was being built for a 1x1 conv
3. Fix the semantics of the parameters to transposed conv to match PyTorch's conventions.  Reading [PyTorch's Transposed Conv module docs](https://docs.pytorch.org/docs/stable/generated/torch.nn.ConvTranspose2d.html), their convention is that to express a convolution A's transpose convolution B (i.e. the convolution B that inverts the operation performed by A), the channel parameters for B should be the reverse of A (e.g. A's in/out 4/16, B's in/out 16/4) and the 'topological' parameters should be carried over from A (i.e. A's stride=2, B's stride=2; not 1/2, etc.).
In our API, we seem to expect inverse expressions of both types of parameters (but our `SimpleUNet` code was expressing neither inversely).
Therefore, in this PR, `SparseConvolutionKernelMap::forward` and `SparseConvolutionKernelMap::backward` have been fixed to treat normal and transposed in/out channel specifications the same way (it is up to the user to reverse them if that's the desired effect per PyTorch's pattern) _and_ `SimpleUNet` was changed to reverse the ordering of the supplied source/target grid, to match what the source/target girds are of the convolution it wants to invert (also per PyTorch's pattern of specifying arguments like stride from the inverse convolution).

Fixes #335

### DIFF
--- a/fvdb/nn/simple_unet.py
+++ b/fvdb/nn/simple_unet.py
@@ -543,7 +543,7 @@ class SimpleUNetUnpad(nn.Module):
 
     def forward(self, data: JaggedTensor, padded_grid: GridBatch, grid: GridBatch) -> JaggedTensor:
         plan = ConvolutionPlan.from_grid_batch_transposed(
-            kernel_size=self.kernel_size, stride=1, source_grid=padded_grid, target_grid=grid
+            kernel_size=self.kernel_size, stride=1, source_grid=grid, target_grid=padded_grid
         )
         return self.deconv(data, plan)
 
--- a/src/fvdb/detail/autograd/SparseConvolutionKernelMap.cpp
+++ b/src/fvdb/detail/autograd/SparseConvolutionKernelMap.cpp
@@ -56,41 +56,23 @@ SparseConvolutionKernelMap::forward(AutogradContext *ctx,
             3,
         },
         opt);
-    if (!transposed) {
-        TORCH_CHECK_VALUE(inFeatures.size(0) == sizes[0],
-                          "The number of input features must match the number of voxels");
-        TORCH_CHECK_VALUE(
-            kernels.dim() == 5,
-            std::string(
-                "Expected kernels to have 5 dimensions (shape (out_ch, in_ch, d, h, w)) but got ") +
-                std::to_string(kernels.dim()) + " dimensions");
-        TORCH_CHECK_VALUE(
-            kernels.size(1) == inFeatures.size(1),
-            "Expected input channels of kernels (" + std::to_string(kernels.size(1)) +
-                ") to equal input channels of features: " + std::to_string(inFeatures.size(1)));
-        const int outC = kernels.size(0), inC = kernels.size(1);
-        kWidth[0] = kernels.size(2);
-        kWidth[1] = kernels.size(3);
-        kWidth[2] = kernels.size(4);
-        kernels   = kernels.permute({2, 3, 4, 1, 0}).reshape({-1, inC, outC}).contiguous();
-    } else {
-        TORCH_CHECK_VALUE(inFeatures.size(0) == sizes[1],
-                          "The number of input features must match the number of voxels");
-        TORCH_CHECK_VALUE(
-            kernels.dim() == 5,
-            std::string(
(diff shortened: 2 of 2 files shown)


fvdb/nn/simple_unet.py
```diff
@@ -556,7 +543,7 @@ def reset_parameters(self) -> None:
 
     def forward(self, data: JaggedTensor, padded_grid: GridBatch, grid: GridBatch) -> JaggedTensor:
         plan = ConvolutionPlan.from_grid_batch_transposed(
-            kernel_size=self.kernel_size, stride=1, source_grid=padded_grid, target_grid=grid
+            kernel_size=self.kernel_size, stride=1, source_grid=grid, target_grid=padded_grid
```
blackencino, line 546: This looks backwards to me, and counter to the intention. I need to look at the torch.convtranspose3d code more carefully, and create a testing framework that confirms the ordering. Generally speaking, we should not be reordering arguments like this, it indicates a semantic mismatch. If the transpose plans are wrong compared to pytorch, then we should fix it in the plan, not in the unet.

src/fvdb/detail/autograd/SparseConvolutionKernelMap.cpp
```diff
@@ -56,41 +56,23 @@ SparseConvolutionKernelMap::forward(AutogradContext *ctx,
...
-            "Expected input channels of kernels (" + std::to_string(kernels.size(0)) +
-                ") to equal input channels of features: " + std::to_string(inFeatures.size(1)));
-        const int inC = kernels.size(0), outC = kernels.size(1);
-        kWidth[0] = kernels.size(2);
-        kWidth[1] = kernels.size(3);
-        kWidth[2] = kernels.size(4);
-        kernels   = kernels.permute({2, 3, 4, 0, 1}).reshape({-1, inC, outC}).contiguous();
-    }
+
+    TORCH_CHECK_VALUE(!transposed ? inFeatures.size(0) == sizes[0] : inFeatures.size(0) == sizes[1],
```
blackencino, line 60: This looks backwards to me. I need to confirm what is expected in torch before agreeing that this is how it should behave. It may be that we need to switch how our transposed convolution works.

fvdb/nn/simple_unet.py
swahtz: Currently, with `source_grid=paded_grid, target_grid=grid`, an exception will be thrown as if the order is incorrect.  Investigating whether that's a mistake in the transposed convolution code or the ordering of the arguments, is what I was explaining in the PR description above, as to which direction to take this in.  If you look at the 'topological' arguments to a PyTorch transposed conv (like stride), a stride of 2 will upsample (insert zeros in the input) in the inverse way a stride of 2 in a conv operator will downsample.  So does that imply that we should order our source and target arguments to the transposed conv as if they were the source and target of the conv operator (since we'd also use stride=2, etc. to have the same meaning and not expect stride=1/2)?

I'm fine with not doing this and source/target take the natural meanings, I'm just trying to determine what is the intent both of PyTorch and the state of the transposed conv kmap code.  You could also read the PyTorch docs for transposed conv as redefining what 'stride' means in transposed conv as basically 'inverse stride'.

harrism:
@swahtz what's the status of this?

swahtz:
> @swahtz what's the status of this?

I think now that we've had the Conv API fixes, I should be able to resolve this.

PR #338, opened 2025-11-17
`build.sh` debug build fixes

### LINKED ISSUE #337: Problem with `debug` build configuration
Trying to create a 'debug' build running the `build.sh` with these options:

`./build.sh install verbose debug  gtests`

I'm not seeing the compilation flags I would expect to see in the verbose output (e.g. I'm seeing `-O3` and not `-O0`).

Reading a bit, I believe our `pyproject.toml` is taking precedence setting our `CMAKE_BUILD_TYPE` [here](https://github.com/openvdb/fvdb-core/blob/main/pyproject.toml#L41) and 'debug' is never being set appropriately as the build type.

@harrism

swahtz:
These are proposed changes which fixes #337.  Primarily this is to make sure the flags we specify for debug builds are passed through to the compiler correctly when using `build.sh` and that debug symbols and line numbers are being preserved by the linker.  The name of the build I changed to 'Debug' because cmake is case sensitive and I wanted to be consistent with CMake ['typical values' ](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html)

### DIFF
--- a/src/tests/CMakeLists.txt
+++ b/src/tests/CMakeLists.txt
@@ -104,7 +104,7 @@ function(ConfigureTest CMAKE_TEST_NAME)
 
   # Add linker flags for Debug builds to preserve symbols
   target_link_options(${CMAKE_TEST_NAME} PRIVATE
-    $<$<CONFIG:Debug>:-rdynamic -g>
+    $<$<CONFIG:Debug>:-rdynamic>
   )
 
   # Link the FINAL executable against necessary libraries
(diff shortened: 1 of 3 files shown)


src/tests/CMakeLists.txt
```diff
@@ -102,6 +102,11 @@ function(ConfigureTest CMAKE_TEST_NAME)
                INSTALL_RPATH "\$ORIGIN/../../../lib"
   )
 
+  # Add linker flags for Debug builds to preserve symbols
+  target_link_options(${CMAKE_TEST_NAME} PRIVATE
+    $<$<CONFIG:Debug>:-rdynamic -g>
```
harrism: Does CMake not automatically pass -g to the linker when build type is `Debug`?

swahtz: Well I noticed the effect that even with build type `Debug` I wasn't seeing any symbols in the binaries I was building.  Let me double check without these

swahtz: You're right, I do see `-g`, I don't see `-rdynamic` which apparently affects symbol visibility when using tools like backtracers.  I can remove -g and just keep `-rdynamic`.

PR #342, opened 2025-11-18
Fix Gaussian rasterization shared memory alignment

swahtz:
Gaussian2D is 32-byte aligned but when we declare the pointer to shared memory, we were not specifying an alignment (leaving us with 4 or 16 byte alignment).  Then casting the pointer with lower alignment to one requiring 32-byte alignment is undefined.  This fixes this potential issue.


harrism (approved):
Great catch.

PR #343, opened 2025-11-19
Support backgrounds in Gaussian Rasterization

### LINKED ISSUE #80: Expose backgrounds parameter in rendering
## 📋 Originally created by @fwilliams

> **Original author:** @fwilliams
> **Date:** 2025-07-06T11:40:59Z

---

Backgrounds are useful if you want to do compositing. 

We support this functionality and should expose it. 

We should also accept an optional depth argument for depth rendering and use the depth to choose the compositing order.

harrism:
Fixes #80

This adds support for backgrounds (per-camera background color / feature) to fVDB's Gaussian Splatting. The kernels already supported backgrounds, but it was not plumbed into the public interface. This PR:

1. Adds gtests to verify kernel correctness
2. Plumbs backgrounds parameters all the way up to the Python public interface.
3. Adds pytests that exercise backgrounds.

### DIFF
--- a/src/fvdb/GaussianSplat3d.cpp
+++ b/src/fvdb/GaussianSplat3d.cpp
@@ -411,11 +411,6 @@ GaussianSplat3d::renderCropFromProjectedGaussiansImpl(
     // Rasterize projected Gaussians to pixels (differentiable)
     // NOTE:  projectGaussians* performs input checking, we need to apply some further
     // checking before GaussianRasterizeToPixels
-    std::optional<torch::Tensor> backgroundsOpt = std::nullopt;
-    if (backgrounds.has_value()) {
-        backgroundsOpt = backgrounds.value();
-    }
-
     auto outputs = detail::autograd::RasterizeGaussiansToPixels::apply(
         projectedGaussians.perGaussian2dMean,
         projectedGaussians.perGaussianConic,
@@ -429,7 +424,7 @@ GaussianSplat3d::renderCropFromProjectedGaussiansImpl(
         projectedGaussians.tileOffsets,
         projectedGaussians.tileGaussianIds,
         false,
-        backgroundsOpt);
+        backgrounds);
     torch::Tensor renderedImage  = outputs[0];
     torch::Tensor renderedAlphas = outputs[1];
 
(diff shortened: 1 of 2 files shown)


src/fvdb/GaussianSplat3d.cpp
```diff
@@ -410,6 +411,11 @@ GaussianSplat3d::renderCropFromProjectedGaussiansImpl(
     // Rasterize projected Gaussians to pixels (differentiable)
     // NOTE:  projectGaussians* performs input checking, we need to apply some further
     // checking before GaussianRasterizeToPixels
+    std::optional<torch::Tensor> backgroundsOpt = std::nullopt;
```
fwilliams: Why do we make an additional copy of the std::optional here rather than passing the one in the argument?

src/fvdb/GaussianSplat3d.cpp
```diff
@@ -1264,6 +1285,11 @@ gaussianRenderJagged(const JaggedTensor &means,     // [N1 + N2 + ..., 3]
     }
 
     // Rasterize projected Gaussians to pixels [differentiable]
+    std::optional<torch::Tensor> backgroundsOpt = std::nullopt;
```
fwilliams: same q. as above -- why not pass in the backgrounds optional directly?

fwilliams (approved):
A few small questions but overall looks good. We may want to have a mode where we pass in a full background image rather than a fixed color which would let us implement e.g. a learnable skybox, but this can be a different PR when we decide to actually implement that feature.

src/fvdb/GaussianSplat3d.cpp
harrism: LLM fart. :) In the autograd files the input is a Variable, not a Tensor, so it has to be done this way, I think. But in this file you are right, it does not need to be unpacked.

harrism: Actually it's not needed in GaussianRender.cpp in spite of the different type. Changing.

src/fvdb/GaussianSplat3d.cpp
```diff
@@ -737,7 +740,10 @@ GaussianSplat3d::renderDepths(const torch::Tensor &worldToCameraMatrices,
     settings.tileSize       = tileSize;
     settings.renderMode     = RenderSettings::RenderMode::DEPTH;
 
-    return renderImpl(worldToCameraMatrices, projectionMatrices, settings);
+    const ProjectedGaussianSplats state =
```
swahtz, line 743: **🤓 nitpick (non-blocking):** ‏ Pedantic but this could probably be a const & and avoid the copy of the RenderSettings object (that has quite a few members that would get copied).

harrism: I don't think it should be -- I believe modern C++ with RVO and move semantics should invoke the move constructor here, and making it a reference might prevent that and also make the code less clear because readers (like me) would have the cognitive load of figuring out what is being referenced.

swahtz: Ah yes this is my C++98 brain not trusting compilers, all good.

PR #348, opened 2025-11-20
plumb sparse rendering functions

fwilliams:
plumbs sparse rendering functions from C++ to Python and adds unit tests for forward and backward


fvdb/_fvdb_cpp.pyi
```diff
@@ -205,6 +236,22 @@ class GaussianSplat3d:
         eps_2d: float = ...,
         antialias: bool = ...,
     ) -> tuple[torch.Tensor, torch.Tensor]: ...
+    def render_features_and_depths_sparse(
```
swahtz: Why `render_x_sparse` and not `sparse_render_x`?  I thought we agreed on verb (sparse_render) then noun?  We have `sparse_conv_3d` and `sparse_transpose_conv_3d`, not `conv_3d_sparse`.

fvdb/__init__.pyi
```diff
@@ -103,6 +103,8 @@ __all__ = [
     "GaussianSplat3d",
     "ProjectedGaussianSplats",
     "ConvolutionPlan",
+    "ProjectionType",
```
swahtz, line 104: Good catch

src/fvdb/detail/autograd/GaussianRasterizeSparse.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include <fvdb/detail/autograd/GaussianRasterizeSparse.h>
+#include <fvdb/detail/ops/gsplat/GaussianRasterizeBackward.h>
+#include <fvdb/detail/ops/gsplat/GaussianRasterizeForward.h>
+#include <fvdb/detail/utils/Nvtx.h>
+#include <fvdb/detail/utils/Utils.h>
+
+namespace fvdb::detail::autograd {
+
+RasterizeGaussiansToPixelsSparse::VariableList
+RasterizeGaussiansToPixelsSparse::forward(
```
swahtz, line 13: **💡 suggestion:** ‏I believe this should be 'SparseRasterize' as the verb like 'SparseConv'

fwilliams: If we're being ultra pedantic we should say convolve_..._sparse but "SparseConv" is its own concept that is well established in the literature (and used as a noun). I think this is fine.

fvdb/_fvdb_cpp.pyi
fwilliams: sparse is an adjective, render is a verb so this conforms. 

The reason to remove is so that when you tab complete, methods that do the same thing but differ in sparsity show up beside each other.

src/fvdb/detail/autograd/GaussianRasterizeSparse.cpp
```diff
@@ -0,0 +1,199 @@
...
+                                                                       activeTiles,
+                                                                       tilePixelMask,
+                                                                       tilePixelCumsum,
+                                                                       pixelMap,
+                                                                       absgrad);
+    });
+    Variable dLossDMean2dAbs;
+    if (absgrad) {
+        dLossDMean2dAbs = std::get<0>(variables);
+        // means2d.absgrad = dLossDMean2dAbs;
```
swahtz: Can remove

tests/unit/test_gaussian_splat_3d.py
```diff
@@ -2247,11 +2247,367 @@ def test_gaussian_contributors_scene_dense_pixels_sparse_render(self):
...
+
+        dense_features, dense_alphas = self.gs3d.render_features(
+            self.cam_to_world_mats[0:1],
+            self.projection_mats[0:1],
+            self.width,
+            self.height,
+            self.near_plane,  # near_plane
+            self.far_plane,  # far_plane
+        )
+        torch.cuda.synchronize()
```
swahtz: **❓ question:** ‏Why the synchronizes here?  Were there any issues without them?

tests/unit/test_gaussian_splat_3d.py
```diff
@@ -2247,11 +2247,367 @@ def test_gaussian_contributors_scene_dense_pixels_sparse_render(self):
...
+            "Sparse quats grad does not match dense quats grad at specified pixels",
+        )
+        self.assertTrue(
+            torch.allclose(sparse_log_scales_grad, dense_log_scales_grad, atol=1e-4, rtol=1e-8),
+            "Sparse log scales grad does not match dense log scales grad at specified pixels",
+        )
+        self.assertTrue(
+            torch.allclose(sparse_logit_opacities_grad, dense_logit_opacities_grad, atol=1e-4, rtol=1e-8),
+            "Sparse logit opacities grad does not match dense logit opacities grad at specified pixels",
+        )
```
swahtz, line 2489: **🚩 issue:** ‏Shouldn't we also be getting and testing the SH grads?

tests/unit/test_gaussian_splat_3d.py
```diff
@@ -2247,11 +2247,367 @@ def test_gaussian_contributors_scene_dense_pixels_sparse_render(self):
...
```
swahtz, line 2617: **❓ question:** ‏As above, shouldn't we be getting and testing the SH grads here?

fvdb/_fvdb_cpp.pyi
swahtz: 'Sparse' relates to the rendering, so this would have to be `render_features_and_depths_sparsely` to be grammatically correct and be what we mean.

We also use 'sparse render' throughout other parts of the API (in the functions called further down by this function).

fwilliams: I renamed the other functions in this file too. It's all consistent. Unless there is a good reason, I'd like to keep this naming.

tests/unit/test_gaussian_splat_3d.py
fwilliams: copilot. I'll rm

tests/unit/test_gaussian_splat_3d.py
fwilliams: The particular superclass for this test doesn't have requires_grad on these for some reason. I'll add this.

src/fvdb/detail/autograd/GaussianRasterizeSparse.cpp
swahtz: 'sparse render' is also an established term

fvdb/_fvdb_cpp.pyi
swahtz: What's the good reason for changing it?

swahtz: The underlying functions are called things like `GaussianSplat3d::sparseRenderImpl`

swahtz: Why don't we do this, why don't we refactor these functions to eliminate the sparse variants and just make the sparse pixels an optional argument?  So `render_features` just takes a `pixels_to_render` optional arg.  The return types would have to `JaggedTensorOrTensorT` but the return types of the sparse functions is already that.  From what I read, all the other args are the same between the two functions and have the same meanings.

harrism: I will need some time if you want me to weigh in on this debate.

swahtz (changes requested):
Thanks for plumbing this functionality up through the python API and putting in the pytests.  We should @harrism some time to review on his workweek to be able to chat about the naming of the functions or potentially merging the public interface to sparse/dense render into the one function.

fvdb/_fvdb_cpp.pyi
fwilliams: For the sake of getting this functionality in, I reverted all naming changes and will make that a seperate PR where we can discuss.

fvdb/gaussian_splatting.py
```diff
@@ -256,12 +257,12 @@ class GaussianSplat3d:
...
       representation (:meth:`render_images`, :meth:`render_images_and_depths`).
     - Rendering depth maps (:meth:`render_depths`, :meth:`render_images_and_depths`).
-    - Rendering features at arbitrary sparse pixel locations (:meth:`sparse_render_features`).
+    - Rendering features at arbitrary sparse pixel locations (:meth:`sparse_render_images`, :meth:`sparse_render_images_and_depths`).
     - Rendering depths at arbitrary sparse pixel locations (:meth:`sparse_render_depths`).
     - Computing which gaussians contribute to each pixel in an image plane
       (:meth:`render_num_contributing_gaussians`, :meth:`render_contributing_gaussian_ids`).
     - Computing the set of Gaussians which contribute to a set of sparse pixel locations
-      (:meth:`sparse_render_num_contributing_gaussians`, :meth:`sparse_render_contributing_gaussian_ids`).
+      (:meth:`render_num_contributing_gaussians_sparse`, :meth:`render_contributing_gaussian_ids_sparse`).
```
swahtz: **⌨️ typo:** ‏I think this revert to the name change was just left should be `sparse_render_contributing...` yes?

fvdb/gaussian_splatting.py
```diff
@@ -1952,7 +2240,7 @@ def render_num_contributing_gaussians(
         )
 
     @overload
-    def sparse_render_num_contributing_gaussians(
+    def render_num_contributing_gaussians_sparse(
```
swahtz: **⌨️ typo:** ‏I think the changes here to this function name (and the ones below) were just left out of the revert, should be `sparse_render_num_contributing_gaussians`, etc.

tests/unit/test_gaussian_splat_3d.py
```diff
@@ -1648,7 +1647,7 @@ def test_gaussians_center_render(self):
         pixels_to_render = JaggedTensor([torch.tensor([[h // 2 - 1, w // 2 - 1]])]).to(self.device)
 
         # test sparse render num contributing gaussians
-        sparse_num_contributing_gaussians, sparse_alphas = gs3d.sparse_render_num_contributing_gaussians(
+        sparse_num_contributing_gaussians, sparse_alphas = gs3d.render_num_contributing_gaussians_sparse(
```
swahtz: **⌨️ typo:** ‏Corresponding to the mentions above, I think these just were left out of the naming change revert

src/fvdb/detail/autograd/GaussianRasterizeSparse.cpp
harrism: In the python code in this PR we have `sparse` first, and in the C++ it is last?

The advantage of putting it last is that like functions and classes get grouped when sorted lexicographically. Of course, that might be a disadvantage to?

src/fvdb/GaussianSplat3d.cpp
```diff
@@ -443,6 +445,47 @@ GaussianSplat3d::renderImpl(const torch::Tensor &worldToCameraMatrices,
         state, settings.tileSize, settings.imageWidth, settings.imageHeight, 0, 0);
 }
 
+std::tuple<JaggedTensor, JaggedTensor>
+GaussianSplat3d::sparseRenderImpl(const JaggedTensor &pixelsToRender,
```
harrism, line 449: Here we have "sparse" first and in GaussianRasterizeSparse.h it is last. In general naming is not consistent in this PR.

PR #351, opened 2025-11-25
Fix potential oversubscription of nvcc * cmake parallel threads

swahtz:
@harrism and I discovered that in high arch compilation situations and high RAM machines, it's possible to get the logic that tries to spread thread parallelism between nvcc and the cmake parallel job execution, that you can end up with a value of nvcc threads * cmake threads that is greater than your available thread count.  Adding logic to cap this to the thread count of the machine.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -42,29 +42,44 @@ setup_parallel_build_jobs() {
   fi
   JOB_RAM_GB=3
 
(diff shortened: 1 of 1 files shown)


harrism (approved):
Thanks, looks good.

PR #352, opened 2025-11-26
Hot fixes for Jcat errors

blackencino:
Jcat was returning the C++ backend JaggedTensor object.

This, oddly, was not failing in many cases because most code paths immediately call ".jdata", which worked on both the python wrapper and the C++ backend, so we weren't seeing this fail.

I added additional tests for this.


harrism:
Can you explain what the correct behavior is and why you expected this to fail?

PR #355, opened 2025-11-27
Fix inverted logic on computing abs(gradient) in backward Gaussian rasterization

### LINKED ISSUE #354: Absgrad logic is inverted in backward rasterization kernel
https://github.com/openvdb/fvdb-core/blob/3b990ac7dd4bb1c9a74949994e9de5fc2e978469/src/fvdb/detail/ops/gsplat/GaussianRasterizeBackward.cu#L507

`mAbsGrad` is true if we are computing absolute values of the gradient of the loss wrt the splat means.  But this code checks if it is false before computing them. I looked back at gsplat and it does the reverse. I think we haven't been testing absgrad. This will be a single-character bug fix.

harrism:
Fixes #354

Fixes inverted logic on AbsGrad computation. 


swahtz (approved):
Hm, that's a strange one, seems pretty wrong!  Thanks

PR #361, opened 2025-12-02
Added backwards/gradient tests to the default convolution unit tests.

blackencino:
Backwards/Gradient tests for default backend convolution unit tests.

New tests:

* test_single_impulse_backward

### DIFF
--- a/tests/unit/test_conv_default.py
+++ b/tests/unit/test_conv_default.py
@@ -444,3 +444,157 @@ class TestConvDefault(unittest.TestCase):
             dense_convolved_at_nonzero_coords_sorted_flat, sparse_convolved_sorted, rtol=1e-5, atol=1e-6
         )
         print("Confirmed that the dense convolved at the nonzero coords matches the sparse convolved.")
+
+    # ==================================================================================
+    # Backward convolution tests
+    # ==================================================================================
+
+    @parameterized.expand(all_device_dtype_combos)
+    def test_single_impulse_backward(self, device: DeviceIdentifier, dtype: torch.dtype):
+        """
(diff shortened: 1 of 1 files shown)


tests/unit/test_conv_default.py
```diff
@@ -444,3 +444,157 @@ def sort_coords(coords):
...
+        # Anti-symmetric kernel with requires_grad
+        kernel = fourier_anti_symmetric_kernel(self.KERNEL_SIZE, dtype=dtype, device=device)
+        self.assertFalse(has_any_symmetry(kernel))
+        kernel_with_channels = kernel.reshape(1, 1, *self.KERNEL_SIZE).clone()
+        kernel_with_channels.requires_grad_(True)
+        print("Created anti-symmetric kernel with requires_grad=True.")
+
+        # Create convolution plan
+        conv_plan = ConvolutionPlan.from_grid_batch(
+            kernel_size=self.KERNEL_SIZE, stride=1, source_grid=grid_batch, target_grid=dst_grid_batch
```
swahtz, line 502: **❓ question:** ‏Have we not implemented the stride functionality yet?  Should we test over values of this too?

blackencino: We have not validated the stride functionality yet. It is next on the validation list. I'm adding these tests starting at the ground up, doing them one at a time.

blackencino: So yes - we should absolutely be testing against striding, but that should be a test that builds on top of these, not woven into these.

swahtz: OK, great, thanks, just was curious.

Issue #242, opened 2025-10-14
Improve Attention


swahtz:
After finding our `ScaledDotProductAttention` operator fails to run on Blackwell, we introduced a work around in #220 by running PyTorch's SDPA on each jagged tensor member individually.  We should revisit this after the 0.3.0 release and update our implementation in a few ways:

1. We should evaluate if our SDPA implementation should be fixed or removed in favour of Pytorch's which can operate on Pytorch's Nested Tensors and we can trivially wrap a `JaggedTensor` without copying to use Pytorch's solution
2. If continuing to use our SDPA, we should think about a switch to PyTorch's CUDNN since our min requirement of PyTorch 2.8 should now align to their inclusion of CUDNN 9
3. We should evaluate using Pytorch's built-in Flash Attention operator which also is compatible with their Nested Tensor which we could trivially wrap our `JaggedTensor` with.  We should use Flash Attention when hardware and data types allow.

PR #364, opened 2025-12-03
Use`scaled_dot_product_attention` operator from Torch, remove our SDPA

### LINKED ISSUE #362: Fix our SDPA or use Pytorch's oeprator
We should evaluate if our SDPA implementation should be fixed or removed in favour of Pytorch's which can operate on Pytorch's Nested Tensors and we can trivially wrap a JaggedTensor without copying to use Pytorch's solution.

If continuing to use our SDPA, we should think about a switch to PyTorch's CUDNN since our min requirement of PyTorch 2.8 should now align to their inclusion of CUDNN 9

swahtz:
This relates to #242.  We had worked around the fact that our `ScaledDotProductAttention` operator wouldn't run on platforms earlier than Hopper by simply using PyTorch's implementation in a loop over each Tensor in a JaggedTensor.  This PR removes that loop in favour of constructing Torch _nested_ Tensors from views of input JaggedTensors and similarly using torch's `scaled_dot_product_attention` (which can operate over nested Tensors).

Similar to our previous use of `at::native::scaled_dot_product_attention` and our `ScaledDotProductAttention`, we have to reorder the axes of the inputs to the SDPA, make them contiguous and reorder on the way out, but this was a necessary condition of our previous implementations.

I removed the old source files for our `ScaledDotProductAttention`/`fvdb::autograd::Attention` which no longer had a public-facing function calling them.   I also removed cudnn and cudnn frontend from our build infrastructure since `ScaledDotProductAttention` was the only code requiring these to build.

fixes #362

### DIFF
--- a/src/fvdb/FVDB.cpp
+++ b/src/fvdb/FVDB.cpp
@@ -53,14 +53,14 @@ scaledDotProductAttention(const JaggedTensor &query,
         const int64_t H  = data.size(1);
         const int64_t D  = data.size(2);
 
-        const int64_t stride_L = H * D;
-
-        auto offsets_tensor       = jt.joffsets().cpu(); // (N+1)
+        const int64_t stride_L    = H * D;
         const int64_t num_tensors = jt.num_tensors();
 
-        // Compute lengths: offsets[1:] - offsets[:-1]
-        auto lengths =
-            offsets_tensor.slice(0, 1, num_tensors + 1) - offsets_tensor.slice(0, 0, num_tensors);
+        // Use cached lsizes - avoids GPU->CPU copy if already computed
+        const auto &lsizes = jt.lsizes1(); // std::vector<int64_t>
+
+        // Create lengths tensor
+        auto lengths = torch::tensor(lsizes, torch::TensorOptions().dtype(torch::kLong));
 
         // Construct nested_size: (N, 3) -> [L_i, H, D]
         auto nested_size = torch::empty({num_tensors, 3}, lengths.options());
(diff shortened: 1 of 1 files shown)


src/fvdb/FVDB.cpp
```diff
@@ -47,26 +46,67 @@ scaledDotProductAttention(const JaggedTensor &query,
...
+
+    // Helper to create a nested tensor view from a JaggedTensor
+    auto make_nested_view = [](const JaggedTensor &jt) -> torch::Tensor {
+        const auto &data = jt.jdata(); // (Total, H, D)
+        const int64_t H  = data.size(1);
+        const int64_t D  = data.size(2);
+
+        const int64_t stride_L = H * D;
+
+        auto offsets_tensor       = jt.joffsets().cpu(); // (N+1)
```
harrism: question: looks like this requires a round trip of the offsets to the CPU. Probably too early to optimize this but probably should be aware of the GPU->CPU copy and sync and then CPU->GPU copy this requires.

harrism (approved):
Looks like a solid improvement. Just one comment about perf.

src/fvdb/FVDB.cpp
swahtz: Good point, we could probably use the lsizes cache for this…

swahtz: My LLM friend helped me write some alternative logic which uses `lsizes()` (which could already be cached on the CPU) instead of always copying `joffsets`.

PR #368, opened 2025-12-04
Create and use getMaxSharedMemory utility

harrism:
We were calling `cudaDeviceGetProperties` every time we launch the GaussianRasterizeBackward kernel. This is expensive (> 1ms per call). This definitely impacted our training time.

This PR adds a new utility that uses (faster) `cudaDeviceGetAttribute` instead, and only ever calls it once per device and caches the result.

### DIFF
--- a/src/fvdb/detail/utils/cuda/Utils.cuh
+++ b/src/fvdb/detail/utils/cuda/Utils.cuh
@@ -91,7 +91,8 @@ getMaxSharedMemory(int device) {
     static std::vector<std::atomic<int>> cache(deviceCount);
     static std::vector<std::once_flag> initFlags(deviceCount);
 
-    if (int cached = cache[device].load(std::memory_order_relaxed); cached > 0) {
+    int cached = cache[device].load(std::memory_order_relaxed);
+    if (cached > 0) {
         return cached;
     }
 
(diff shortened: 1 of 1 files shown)


src/fvdb/detail/utils/cuda/Utils.cuh
```diff
@@ -72,6 +72,43 @@ mergeStreams() {
...
+    if (int cached = cache[device].load(std::memory_order_relaxed); cached > 0) {
+        return cached;
+    }
+
+    std::call_once(initFlags[device], [device]() {
+        int maxSharedMemory = 0;
+        TORCH_CHECK(cudaDeviceGetAttribute(&maxSharedMemory,
+                                           cudaDevAttrMaxSharedMemoryPerBlockOptin,
+                                           device) == cudaSuccess,
+                    "Failed to get max shared memory per block optin");
```
blackencino: optin - should this be option? or opt-in?

src/fvdb/detail/utils/cuda/Utils.cuh
```diff
@@ -72,6 +72,43 @@ mergeStreams() {
...
+getMaxSharedMemory(int device) {
+    TORCH_CHECK(device >= 0, "Invalid device index: ", device);
+
+    static const int deviceCount = c10::cuda::device_count();
+    TORCH_CHECK(device < deviceCount, "Device index out of range: ", device);
+
+    static std::vector<std::atomic<int>> cache(deviceCount);
+    static std::vector<std::once_flag> initFlags(deviceCount);
+
+    if (int cached = cache[device].load(std::memory_order_relaxed); cached > 0) {
```
blackencino: can this be separated into two lines? 
```
int cached = ...;
if (cached > 0) ...
```

src/fvdb/detail/utils/cuda/Utils.cuh
harrism: See above the name of the enum being passed: `cudaDevAttrMaxSharedMemoryPerBlockOptin`

src/fvdb/detail/utils/cuda/Utils.cuh
harrism: yea, good catch. That's LLM code I didn't even notice. Ugly!

src/fvdb/detail/utils/cuda/Utils.cuh
harrism: I removed "optin" from the string, it was confusing.

src/fvdb/detail/utils/cuda/Utils.cuh
harrism: Fixed.

PR #397, opened 2025-12-17
Add PR workflow commands

phapalova:
Adds a new GitHub Actions workflow that allows maintainers to trigger CI workflows via PR comments. This is useful when workflows need to be re-run manually or when the automatic triggers didn't fire due to PR draft state.

| Command | Description |
|---------|-------------|
| `/run-all` | Trigger all PR workflows (tests, codestyle, cu128, cu130) |
| `/run-tests` | Trigger `tests.yml` only |
| `/run-codestyle` | Trigger `codestyle.yml` only |
| `/run-cu128` | Trigger `cu128.yml` only |
| `/run-cu130` | Trigger `cu130.yml` only |
| `/rerun-all` | Rerun all failed workflows |

### DIFF
--- a/.github/workflows/codestyle.yml
+++ b/.github/workflows/codestyle.yml
@@ -3,6 +3,7 @@ on:
     pull_request:
       branches:
         - '**'
+    workflow_dispatch:
 
 # Allow subsequent pushes to the same PR or REF to cancel any previous jobs.
 concurrency:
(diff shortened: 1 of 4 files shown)


swahtz:
If the issue this PR is trying to solve is the problem of our CI jobs not triggering when jobs go from draft->open, there's a `ready_for_review` trigger we could add that would fix this.  I'm not sure what other situations this would help solve.  If we added `ready_for_review`, could we just close this?

PR #398, opened 2025-12-29
Fix the docstrings of  `fvdb.viz.Scene.camera_orbit_direction`

swahtz:
Fixed the docstring for `fvdb.viz.Scene.camera_orbit_direction` which mentions that this is the direction pointing from the center to the camera but the values being returned are instead the direction from the camera to the center.  These values are the `eye_direction` in the nanovdb-editor which can be seen as being the opposite of what we describe in the `viz.Scene` docstrings here:

https://github.com/openvdb/nanovdb-editor/blob/main/nanovdb_editor/putil/Camera.h#L381

Perhaps, for future discussion, we should use the same terminology and call these attrs `camera_eye_direction`, `camera_eye_distance`, and so on; I think that language is more standard and correct.  I'm not sure why we use the word 'orbit' except for perhaps legacy reasons when these APIs related to specifically to parameters of the orbits of drones, visualization rendering, or nomenclature in viser?

### DIFF
--- a/fvdb/viz/_scene.py
+++ b/fvdb/viz/_scene.py
@@ -351,7 +351,7 @@ class Scene:
             The camera itself is positioned at: ``camera_position = orbit_center - orbit_radius * orbit_direction``
 
         Returns:
-            direction (torch.Tensor): A tensor of shape ``(3,)`` representing the direction pointing from the orbit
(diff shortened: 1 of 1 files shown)


matthewdcong (approved):

swahtz:
I'll start an issue as to whether we want to change the names of these attributes.

PR #400, opened 2025-12-31
Fix NaNs in `rasterizeTopContributingGaussianIdsForward` weights

swahtz:
I was finding that `rasterizeTopContributingGaussianIdsForward` during training could occasionally produce NaNs in its returned weights.  Logically I don't understand how a condition like this could be reached, but initializing the output weights to zeros (instead of `torch::empty`), mitigates this issue.


harrism:
Is this only an output array? i.e. the data are never read? If so, this is strange that zeroing it matters.  Perhaps it is being read somewhere subtle, perhaps with a `+=` or an atomic increment / addition?

PR #401, opened 2026-01-01
Plumb Sparse Gaussian TileIntersection

swahtz:
When profiling sparse rasterization in functions like `sparseRenderTopContributingGaussianIds`, I was noticing that the projection phase, specifically the gaussian tile intersection, dominated the runtime.  However, these intersections were being computed for _all_ tiles when we only eventually rasterize a small number of them.  I went back and plumbed `dispatchGaussianSparseTileIntersection` which had been written back in February of 2025 and this has provided a 7x speedup for the tile intersection cost and a 4x speedup to `sparseRenderTopContributingGaussianIds` overall.

The most significant downstream change is that now `tileOffsets` can be either a tensor of dimension 3 `[C, numTilesH, numTilesW]` in the usual dense rendering case or `[active_tiles + 1]` in the sparse case.  This just necessitated changing some functions that relied on `tileOffsets`'s sizes to infer any of those quantities to get them from elsewhere.

### DIFF
--- a/src/fvdb/detail/ops/gsplat/GaussianRasterize.cuh
+++ b/src/fvdb/detail/ops/gsplat/GaussianRasterize.cuh
@@ -134,7 +134,6 @@ template <typename ScalarType, size_t NUM_CHANNELS, bool IS_PACKED> struct Raste
           mConics(initAccessor<ScalarType, NUM_OUTER_DIMS + 1>(conics, "conics")),
           mOpacities(initAccessor<ScalarType, NUM_OUTER_DIMS>(opacities, "opacities")),
           mTileGaussianIds(initAccessor<int32_t, 1>(tileGaussianIds, "tileGaussianIds")),
-          // Detect sparse tile offsets from tensor dimensions (1D = sparse, 3D = dense)
           mTileOffsetsAreSparse(tileOffsets.dim() == 1),
           // Initialize 3D accessor for dense mode, or placeholder for sparse mode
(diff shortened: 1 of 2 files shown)


src/fvdb/detail/ops/gsplat/GaussianRasterize.cuh
```diff
@@ -85,8 +85,13 @@ template <typename ScalarType, size_t NUM_CHANNELS, bool IS_PACKED> struct Raste
     VectorAccessor mMeans2d;                  // [C, N, 2] or [nnz, 2]
     VectorAccessor mConics;                   // [C, N, 3] or [nnz, 3]
     ScalarAccessor mOpacities;                // [C, N] or [nnz]
-    TorchRAcc64<int32_t, 3> mTileOffsets;     // [C, nTilesH, nTilesW]
     TorchRAcc64<int32_t, 1> mTileGaussianIds; // [totalIntersections]
+    // Tile offsets: can be 3D [C, nTilesH, nTilesW] (dense) or 1D [AT + 1] (sparse)
+    // We store both accessor types but only one is valid based on mTileOffsetsAreSparse
```
swahtz, line 90: Perhaps at some point we should spend some time to make this more elegant/less complex by finding a solution that does not require two accessors depending on the sparse projection/tile intersection results.  This would also dry up some of the functionality we repeat in this PR to check or use these accessors.  The issue is that the format and use of these tensors is used by both projection and rasterization, so it needs to be encapsulated in such a way that either can make use of it.

PR #402, opened 2026-01-05
Expose k and t parameters in MCMC relocation

fwilliams:
These parameters control the falloff curve for noise added to each Gaussian during optimization as a function of scale and opacity (more opaque Gaussians are considered more certain and undergo smaller perturbation during optimization). This PR doesn't change the default values of these but exposes them so they can be tuned as hyperparameters.


matthewdcong:
OK, if it's mostly for developer use then we can just keep it as is for now. The comments outline the usage and are probably sufficient for dev purposes.

matthewdcong (approved):

PR #403, opened 2026-01-06
Optimize joffsets construction via pinned memory

matthewdcong:
During Gaussian splat training, we construct several JaggedTensor instances where the small joffsets tensor is initialized on the host and then copied over to the device. Page-locking these small tensors on the host enables them to be read at a higher bandwidth by the device which accelerates the subsequent host to device transfer. Furthermore, with the Unified Memory backend, this reduces the amount of host <-> device synchronization required enabling the the subsequent small kernels to be launched with lower latency.

NB: The `torch::tensor(...)` call with the initialization list performs the initialization on the host before copying the contents to the device.


matthewdcong:
With Unified Memory, the training runs about 1.06x faster. The benefits for the PyTorch CUDA backend are more marginal because host to device transfers call `cudaStreamSynchronize` by default and I don't want to make the transfer non-blocking right now (though we could in the future) in case something else in the CUDA backend is relying on the sync.

swahtz (approved):
Nice, good catch.  I wonder if there are other situations with similar small CPU tensors that need to be created and then moved that we could apply this to.

PR #409, opened 2026-01-12
Fix/improve radix sort synchronization

matthewdcong:
The item offsets and counts arrays are only accessed on the host. Prior to this PR, these were allocated in Unified Memory without the appropriate synchronization. This leads to a rare race condition and also necessitates additional host-device synchronization at the beginning of the radix sort loop. Migrating these arrays to the host fixes the race condition and removes the need to synchronize.


swahtz (approved):
That's great, thanks!

PR #412, opened 2026-01-14
Fix the derivation of the number of cameras in the rasterization kernels that would be incorrect when in packed mode.

swahtz:
I noticed that the value for the number of cameras was being incorrectly taken from the batch dimension of `means2d` but when in packed mode, this would be the number of gaussians, not the number of cameras.  Fixed up this logic, made comments about dimensions more complete and added some tests.


harrism (approved):
Thanks for the fix, and adding those tests is a big improvement in coverage!

PR #416, opened 2026-01-15
CMake: create nanovdb_editor_BINARY_DIR before using it as a working directory

harrism:
When running benchmarks in a container, the `fvdb-core` build was failing due `get_nanovdb_editor.cmake` attempting to pip install from a working directory that did not yet exist. Adding a `file(MAKE_DIRECTORY ${nanovdb_editor_BINARY_DIR})` fixes this.

### DIFF
--- a/src/cmake/get_nanovdb_editor.cmake
+++ b/src/cmake/get_nanovdb_editor.cmake
@@ -144,17 +144,24 @@ if (NOT NANOVDB_EDITOR_FORCE)
         if(NUM_LATEST_WHEELS GREATER 0)
             list(GET LATEST_WHEELS 0 LATEST_WHEEL)
             message(STATUS "Found wheel in dist for the latest version ${NANOVDB_EDITOR_LATEST_VERSION}: ${LATEST_WHEEL}")
+            message(STATUS "Installing nanovdb_editor wheel with: ${Python3_EXECUTABLE} -m pip install --force-reinstall -v ${LATEST_WHEEL}")
+            # CPM's nanovdb_editor_BINARY_DIR may not exist yet on this fast-path (installing a prebuilt wheel),
+            # but execute_process requires WORKING_DIRECTORY to exist.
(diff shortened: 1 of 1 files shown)


src/cmake/get_nanovdb_editor.cmake
```diff
@@ -144,17 +144,24 @@ if (NOT NANOVDB_EDITOR_FORCE)
         if(NUM_LATEST_WHEELS GREATER 0)
             list(GET LATEST_WHEELS 0 LATEST_WHEEL)
             message(STATUS "Found wheel in dist for the latest version ${NANOVDB_EDITOR_LATEST_VERSION}: ${LATEST_WHEEL}")
+            message(STATUS "Installing nanovdb_editor wheel with: ${Python3_EXECUTABLE} -m pip install --force-reinstall -v ${LATEST_WHEEL}")
+            # CPM's nanovdb_editor_BINARY_DIR may not exist yet on this fast-path (installing a prebuilt wheel),
+            # but execute_process requires WORKING_DIRECTORY to exist.
+            file(MAKE_DIRECTORY ${nanovdb_editor_BINARY_DIR})
             execute_process(
-                COMMAND bash -lc "
```
phapalova, line 148: I guess we can skip `bash -lc` because in conda env anyway?

phapalova (approved):
looks good, thanks for the fix

src/cmake/get_nanovdb_editor.cmake
harrism: LLM suggested this change. I asked why and it suggests we don't need bash -lc here because we aren't using shell features like globbing, piping, or using `&&` etc.  Shelling out introduces quoting/injection risks and breaks easily on paths with spaces.

The block lower in the file needs bash because it runs multiple commands. But the cmake idiomatic approach would be to break it into two separate `execute_process` calls and use cmake globbing and lists instead of relying on the shell.

PR #427, opened 2026-01-23
Fix semantics of GaussianTileIntersection's torch::cumsum

swahtz:
In `GaussianTileIntersection` we use `torch::cumsum_out` in two places using the `tilesPerGaussian` tensor as both the input and output arguments.  `cumsum_out` does not provide a guarantee that aliasing the storage of the in/out tensors like this will produce a correct result.  Specifically the schema (from Torch's RegistrationDeclarations.h) is:

`aten::cumsum.out(Tensor self, int dim, *, ScalarType? dtype=None, Tensor(a!) out) -> Tensor(a!)`

Even though this does seem to work today in the current version of PyTorch we test against, to be correct we should use a separate tensor as the output tensor for use with `torch::cumsum_out`.

Instead, I have changed the semantics to just use `torch::cumsum`.  I have measured the difference in runtime to be a few 1/1000ths of a ms (out of 7.621ms) between the two when running a benchmark on just the projection part of the rasterization pipeline.

### DIFF
--- a/src/fvdb/detail/ops/gsplat/GaussianTileIntersection.cu
+++ b/src/fvdb/detail/ops/gsplat/GaussianTileIntersection.cu
@@ -459,12 +459,8 @@ gaussianTileIntersectionCUDAImpl(
                                                  tilesPerGaussianCumsum.data_ptr<int32_t>());
     C10_CUDA_KERNEL_LAUNCH_CHECK();
 
-    // In place cumulative sum to get the total number of intersections
-    CUB_WRAPPER(cub::DeviceScan::InclusiveSum,
-                tilesPerGaussianCumsum.data_ptr<int32_t>(),
-                tilesPerGaussianCumsum.data_ptr<int32_t>(),
-                totalGaussians,
-                stream);
+    // cumulative sum to get the total number of intersections
+    tilesPerGaussianCumsum = torch::cumsum(tilesPerGaussianCumsum, 0, torch::kInt32);
 
     // Allocate tensors to store the intersections
     const int64_t totalIntersections = tilesPerGaussianCumsum[-1].item<int64_t>();
@@ -1038,16 +1034,8 @@ gaussianTileIntersectionPrivateUse1Impl(
 
     mergeStreams();
 
-    // In place cumulative sum to get the total number of intersections
-    {
-        const auto deviceGuard = at::cuda::OptionalCUDAGuard(at::device_of(means2d));
-        auto stream            = at::cuda::getCurrentCUDAStream(means2d.device().index());
-        CUB_WRAPPER(cub::DeviceScan::InclusiveSum,
-                    tilesPerGaussianCumsum.data_ptr<int32_t>(),
-                    tilesPerGaussianCumsum.data_ptr<int32_t>(),
-                    totalGaussians,
-                    stream);
-    }
+    // cumulative sum to get the total number of intersections
+    tilesPerGaussianCumsum = torch::cumsum(tilesPerGaussianCumsum, 0, torch::kInt32);
 
     // Allocate tensors to store the intersections
     const int64_t totalIntersections = tilesPerGaussianCumsum[-1].item<int64_t>();


src/fvdb/detail/ops/gsplat/GaussianTileIntersection.cu
```diff
@@ -1035,7 +1039,14 @@ gaussianTileIntersectionPrivateUse1Impl(
...
     // In place cumulative sum to get the total number of intersections
-    torch::cumsum_out(tilesPerGaussianCumsum, tilesPerGaussianCumsum, 0, torch::kInt32);
+    {
+        auto stream = at::cuda::getCurrentCUDAStream(means2d.device().index());
+        CUB_WRAPPER(cub::DeviceScan::InclusiveSum,
+                    tilesPerGaussianCumsum.data_ptr<int32_t>(),
+                    tilesPerGaussianCumsum.data_ptr<int32_t>(),
+                    totalGaussians,
+                    stream);
+    }
```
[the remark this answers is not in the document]
swahtz: Good catch, thank you, I fixed that now

src/fvdb/detail/ops/gsplat/GaussianTileIntersection.cu
```diff
@@ -1035,7 +1039,15 @@ gaussianTileIntersectionPrivateUse1Impl(
     mergeStreams();
 
     // In place cumulative sum to get the total number of intersections
-    torch::cumsum_out(tilesPerGaussianCumsum, tilesPerGaussianCumsum, 0, torch::kInt32);
+    {
```
matthewdcong: This won't work for multi-GPU. The CUDA guard will fail because the device type is privateuseone. In addition, it willl launch a single kernel on device 0 without the correct multi-device post-kernel synchronization.

With the mGPU backend, `torch::cumsum_out` already dispatches to a mGPU implementation of cumulative sum that uses `InclusiveScan` on each device followed by a reduction across devices. Thus, I think we can keep the existing code here unchanged.

src/fvdb/detail/ops/gsplat/GaussianTileIntersection.cu
```diff
@@ -460,7 +460,11 @@ gaussianTileIntersectionCUDAImpl(
     C10_CUDA_KERNEL_LAUNCH_CHECK();
 
     // In place cumulative sum to get the total number of intersections
-    torch::cumsum_out(tilesPerGaussianCumsum, tilesPerGaussianCumsum, 0, torch::kInt32);
+    CUB_WRAPPER(cub::DeviceScan::InclusiveSum,
```
matthewdcong: How much faster is the CUB implementation? I think `cumsum_out` for kCUDA dispatches to a kernel that's launched on `getCurrentCUDAStream()` under the hood so it should have the correct stream-ordering semantics.

swahtz: I just looked back through my changes (I was carving out a bunch of changes I was doing to optimize/fix the segmentation model into smaller PRs) and I realize this change was not as much a performance as a 'correctness' change.  So the schema for `torch::cumsum_out` does not guarantee the safety of an in-place cumsum on the `tilesPerGaussianCumsum` tensor in our code.  Specifically the schema (from `RegistrationDeclarations.h`) is:

`aten::cumsum.out(Tensor self, int dim, *, ScalarType? dtype=None, Tensor(a!) out) -> Tensor(a!)`

While `cumsum_out` might appear to work correctly in the current version of PyTorch we use, it is not guaranteed to work.  To my understanding, if it were to guarantee we could do this in-place on `tilesPerGaussianCumsum`, the schema would have to be:

`aten::cumsum.out(Tensor(a!) self, int dim, *, ScalarType? dtype=None, Tensor(a!) out) -> Tensor(a!)`

So instead of adding an extra allocation for a new output tensor to make this correct, I switched it to the CUB implementation which offers this guarantee.  Is it a lot of work for us to implement this correctly (e.g. should I look at taking what Torch does to implement it here) or perhaps I should change this PR to add the output allocation?

I'll change the PR summary to be more accurate (I think I just used Copliot to generate the summary).

matthewdcong: Got it, thanks for the info. I did some digging and PyTorch does call `cub::InclusiveScan` under the hood for `torch::cumsum_out` (refer to [this block](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/ScanUtils.cuh#L452)). Thus, I think in the _current_ state of the PyTorch code `cumsum_out` would give the correct results even if it doesn't strictly comply with the schema.

To address the correctness issue, I'd suggest changing the call to `tilesPerGaussianCumsum = torch::cumsum(tilesPerGaussianCumsum, 0, torch::kInt32)` which should have the correct aliasing semantics. That may introduce an additional cost due to allocation but I'm not sure if it would be noticeable. The PyTorch block allocator is relatively efficient when it comes to on-the-fly Tensor allocations which happen in almost every operator invocation so we'd just need to see how the timings compare.

swahtz: I benchmarked the CUB call against the `torch::cumsum` and I get almsot the same times (within just a few 1/1000ths of a ms).  Thanks for having a look, the overhead seems minimal and I'll just adjust this PR to use `torch::cumsum`

matthewdcong: Looks good, thanks for  the changes.

PR #439, opened 2026-02-02
Disable two GCC 13.3 warnings

matthewdcong:
GCC 13.3 doesn't correctly statically analyze the PyTorch compile-time logic for autograd (i.e. counting the number of `Variables` returned from the forward pass) and throw what I believe to be erroneous `Warray-bounds` and `Wstringop-overflow` warnings. These are in particular associated with the `VolumeRender` function and its forward pass which unlike most returns a `vector` of tensors rather than a single one.

PyTorch encountered a similar error in https://github.com/pytorch/pytorch/pull/137092 and disabled the warning.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -208,6 +208,8 @@ target_compile_options(fvdb PRIVATE
     $<$<COMPILE_LANGUAGE:CXX>:
     "-Wno-unknown-pragmas"
     "-Wno-class-memaccess"
+    "-Wno-array-bounds"
+    "-Wno-stringop-overflow"
     "-fdiagnostics-color=always"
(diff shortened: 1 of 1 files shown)


harrism (commented):
Do we really want to disable these warnings everywhere? Can you localize it to the problematic files or lines?

The pytorch issue was for something different, they added `-Wno-error=nonnull`, specifically for a test target.

matthewdcong:
I tried to localize it but wasn't able to find a solution. It doesn't work to surround the header because the actual ordering could be different due to transitive includes. Furthermore, surrounding the function doesn't work either. I even tried converting it to a tuple instead of a vector at the top level but that doesn't work either :disappointed:

harrism (approved):

PR #441, opened 2026-02-03
Improve PyTorch build configuration time

matthewdcong:
Because PyTorch is an extensive module, `import torch` requires around 1.5s of time. When discovering and configuring PyTorch as a part of the CMake build system, we call `import torch` twice in succession to get two different configuration variables. We can speed this up by importing PyTorch once and querying both configuration variables simultaneously.

On my system this brings a no-op incremental build from 9.6s to 8.4 seconds.


blackencino (approved):
LGTM. 

Anything to speed up the build is a godsend, thank you.

PR #444, opened 2026-02-05
Rasterization using 3d gaussians

fwilliams:
## Summary
This PR adds a **dense, differentiable rasterize-from-world** path for Gaussian splatting that renders images **directly from world-space 3D Gaussians** (means/quats/log-scales) instead of relying only on 2D projected conics. It supports **UT-based projection for OpenCV camera models**, integrates into the `GaussianSplat3d` public API, and includes extensive regression/gradient tests.

## Key changes
- **New public API**
  - `GaussianSplat3d.render_images_from_world(...)` (Python + C++ bindings)
  - Supports `CameraModel` + `RollingShutterType` and packed OpenCV distortion coefficients.

- **New kernels + autograd (dense-only)**
  - CUDA forward/backward for rasterization from world-space 3D Gaussians.
  - Autograd wrapper enforces correct backward arity (one grad per forward input).

- **Camera model consolidation + reuse**
  - Shared `CameraModel` / `RollingShutterType` enums (C++) and Python `IntEnum` wrappers in `fvdb.enums`.
  - Shared OpenCV distortion math and pose interpolation helpers used across UT + rasterize-from-world.

- **Masks + backgrounds parity**
  - Optional **tile masks** and **per-camera backgrounds** with behavior aligned to the classic rasterizer.
  - **Deadlock fixes**: block-wide early exits for masked/no-intersection **edge tiles** in both classic and from-world codepaths.

- **Docs + naming cleanup**
  - Expanded Python docstrings for `render_images_from_world`.
  - Removed deprecated aliases and aligned naming (no `*_3dgs` suffix in the public API).

## Test plan
- **C++**
  - `./build.sh ctest`
- **Python**
  - Run from `/tmp` so Python imports the installed extension:
    - `conda run -n fvdb python -m pytest -q --rootdir /path/to/fvdb-core/tests /path/to/fvdb-core/tests`

## Notes / follow-ups
- Rasterize-from-world is currently **CUDA-only** and **dense-only** (no jagged/sparse output).
- `CameraModel` / `RollingShutterType` are kept explicit in some dispatch APIs; folding them into `RenderSettings` could be a follow-up if we want a single transport object everywhere.

### DIFF
--- a/src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
+++ b/src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
@@ -27,9 +27,9 @@ template <uint32_t NUM_CHANNELS> struct SharedGaussian {
 
 template <uint32_t NUM_CHANNELS> struct RasterizeFromWorldForwardArgs {
     RasterizeFromWorldCommonArgs commonArgs;
-    float *__restrict__ outFeatures;  // [C,H,W,D]
-    float *__restrict__ outAlphas;    // [C,H,W,1] packed as [C*H*W]
-    int32_t *__restrict__ outLastIds; // [C,H,W]
+    fvdb::TorchRAcc64<float, 4> outFeatures;  // [C,H,W,D]
+    fvdb::TorchRAcc64<float, 4> outAlphas;    // [C,H,W,1]
+    fvdb::TorchRAcc64<int32_t, 3> outLastIds; // [C,H,W]
 
     inline __device__ void
     volumeRenderTileForward() const {
@@ -39,12 +39,13 @@ template <uint32_t NUM_CHANNELS> struct RasterizeFromWorldForwardArgs {
 
         uint32_t camId, tileRow, tileCol, row, col;
         common.denseCoordinates(camId, tileRow, tileCol, row, col);
-        const bool inside = (row < common.imageHeight && col < common.imageWidth);
-
-        // Pixel index within this camera.
-        const uint32_t pixId       = common.pixelId(row, col);
-        const uint32_t imgBaseFeat = common.outputFeatureBase(camId, pixId);
-        const uint32_t imgBasePix  = common.outputPixelBase(camId, pixId);
+        const bool inside     = (row < common.imageHeight && col < common.imageWidth);
+        float *outFeaturesPtr = outFeatures.data() + camId * outFeatures.stride(0) +
+                                row * outFeatures.stride(1) + col * outFeatures.stride(2);
+        float *outAlphaPtr = outAlphas.data() + camId * outAlphas.stride(0) +
+                             row * outAlphas.stride(1) + col * outAlphas.stride(2);
+        int32_t *outLastIdPtr = outLastIds.data() + camId * outLastIds.stride(0) +
+                                row * outLastIds.stride(1) + col * outLastIds.stride(2);
 
         // Parity with classic rasterizer: masked tiles write background and exit.
         //
@@ -55,11 +56,11 @@ template <uint32_t NUM_CHANNELS> struct RasterizeFromWorldForwardArgs {
         const bool tileMasked = common.tileMasked(camId, tileRow, tileCol);
         if (tileMasked) {
             if (inside) {
-                outAlphas[imgBasePix]  = 0.0f;
-                outLastIds[imgBasePix] = -1;
+                outAlphaPtr[0]  = 0.0f;
+                outLastIdPtr[0] = -1;
 #pragma unroll
                 for (uint32_t k = 0; k < NUM_CHANNELS; ++k) {
-                    outFeatures[imgBaseFeat + k] = common.backgroundValue(camId, k);
+                    outFeaturesPtr[k * outFeatures.stride(3)] = common.backgroundValue(camId, k);
                 }
             }
             return;
@@ -103,11 +104,11 @@ template <uint32_t NUM_CHANNELS> struct RasterizeFromWorldForwardArgs {
         if (rangeEnd <= rangeStart) {
             if (inside) {
                 // alpha=0, output background if provided else 0.
-                outAlphas[imgBasePix]  = 0.0f;
-                outLastIds[imgBasePix] = -1;
+                outAlphaPtr[0]  = 0.0f;
+                outLastIdPtr[0] = -1;
 #pragma unroll
                 for (uint32_t k = 0; k < NUM_CHANNELS; ++k) {
-                    outFeatures[imgBaseFeat + k] = common.backgroundValue(camId, k);
+                    outFeaturesPtr[k * outFeatures.stride(3)] = common.backgroundValue(camId, k);
                 }
             }
             return;
@@ -202,11 +203,11 @@ template <uint32_t NUM_CHANNELS> struct RasterizeFromWorldForwardArgs {
             return;
         }
 
-        outAlphas[imgBasePix]  = 1.0f - transmittance;
-        outLastIds[imgBasePix] = curIdx;
+        outAlphaPtr[0]  = 1.0f - transmittance;
+        outLastIdPtr[0] = curIdx;
 #pragma unroll
         for (uint32_t k = 0; k < NUM_CHANNELS; ++k) {
-            outFeatures[imgBaseFeat + k] =
+            outFeaturesPtr[k * outFeatures.stride(3)] =
                 pixOut[k] + transmittance * common.backgroundValue(camId, k);
         }
     }
@@ -291,10 +292,11 @@ launchForward(const torch::Tensor &means,
     commonArgs.backgrounds = opt.backgrounds;
     commonArgs.masks       = opt.masks;
 
-    RasterizeFromWorldForwardArgs<NUM_CHANNELS> args{commonArgs,
-                                                     outFeatures.data_ptr<float>(),
-                                                     outAlphas.data_ptr<float>(),
-                                                     outLastIds.data_ptr<int32_t>()};
+    RasterizeFromWorldForwardArgs<NUM_CHANNELS> args{
+        commonArgs,
+        outFeatures.packed_accessor64<float, 4, torch::RestrictPtrTraits>(),
+        outAlphas.packed_accessor64<float, 4, torch::RestrictPtrTraits>(),
+        outLastIds.packed_accessor64<int32_t, 3, torch::RestrictPtrTraits>()};
 
     const size_t blockSize = (size_t)tileSize * (size_t)tileSize;
     const size_t sharedMem = blockSize * (sizeof(int32_t) + sizeof(SharedGaussian<NUM_CHANNELS>));


src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.h
```diff
@@ -0,0 +1,64 @@
...
+    // Per-camera quantities
+    const torch::Tensor &features,                // [C, N, D]
+    const torch::Tensor &opacities,               // [C, N]
+    const torch::Tensor &worldToCamMatricesStart, // [C, 4, 4]
+    const torch::Tensor &worldToCamMatricesEnd,   // [C, 4, 4]
+    const torch::Tensor &projectionMatrices,      // [C, 3, 3]
+    const torch::Tensor &distortionCoeffs,        // [C, K] (K=0 or 12)
+    const RollingShutterType rollingShutterType,
+    const CameraModel cameraModel,
+    // Render settings
```
swahtz, line 48: We should use the `RenderSettings` object here like the other dispatchGaussianRasterize functions.

We should also just roll `CameraModel` and `RollingShutterType` into `RenderSettings` as well to help transport them all together.

fwilliams: See comment below, I have another PR lined up that does a lot of cleanup and deduplication in the backend (and it can include this). If I add it here, it makes this PR quite large so I think we should split it up.

src/fvdb/detail/ops/gsplat/GaussianRasterizeForward.cu
```diff
@@ -584,16 +596,15 @@ dispatchGaussianRasterizeForward<torch::kCUDA>(
     const uint32_t imageOriginW,
     const uint32_t imageOriginH,
     const uint32_t tileSize,
-    const torch::Tensor &tileOffsets,              // [C, tile_height, tile_width]
-    const torch::Tensor &tileGaussianIds,          // [n_isects]
-    const at::optional<torch::Tensor> &backgrounds // [C, D]
+    const torch::Tensor &tileOffsets,               // [C, tile_height, tile_width]
+    const torch::Tensor &tileGaussianIds,           // [n_isects]
+    const at::optional<torch::Tensor> &backgrounds, // [C, D]
+    const at::optional<torch::Tensor> &masks        // [C, tile_height, tile_width] bool
```
swahtz, line 600: This PR exposes the functionality of the `masks` argument, should we include some tests for this functionality?

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.h
swahtz: OK thanks

fvdb/gaussian_splatting.py
```diff
@@ -1877,6 +1878,116 @@ def render_images(
...
+        masks: torch.Tensor | None = None,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """
+        Render dense images by rasterizing directly from world-space 3D Gaussians.
+
+        This is similar to :meth:`render_images`, but the rasterization step is performed in 3D
+        using per-pixel rays against the Gaussian ellipsoids (instead of rasterizing 2D conics
+        produced by a projection step). This enables gradients w.r.t. Gaussian geometry
+        (``means``, ``quats``, ``log_scales``) through rasterization, which is useful for
+        UT-based OpenCV camera models.
```
harrism: **💡 suggestion:** ‏Spell out Unscented Transform here. Also, what is an OpenCV camera model, and is it a standard thing for them to be UT-based? Or maybe you mean something more specific here.

src/fvdb/GaussianSplat3d.h
```diff
@@ -905,6 +902,34 @@ class GaussianSplat3d {
...
+    /// This is similar to @ref renderImages but performs rasterization directly from world-space
+    /// Gaussians (means/quats/log-scales) rather than from their 2D projections. This enables
+    /// geometry gradients through the rasterization step.
+    ///
+    /// Tile intersections are still computed using a (non-differentiable) projection step:
+    /// - For `cameraModel == CameraModel::PINHOLE` or `CameraModel::ORTHOGRAPHIC`, we reuse the
+    ///   classic projection path.
+    /// - For OpenCV camera models, we use the UT projection kernel to compute per-Gaussian radii
+    ///   and depths for sorting / tiling, then rasterize with 3DGS.
+    std::tuple<torch::Tensor, torch::Tensor> renderImagesFromWorld(
```
swahtz, line 922: I think as a user I'd find the naming of this function confusing.  I understand where the naming comes from in other projects and after reading the PR description and code that the rendering results are rasterized 'from world-space Gaussians' as opposed to 'from camera-space Gaussians' for the existing function.  Feel free to create an issue from this rather than addressing but I'd be keen to see us have clearer naming

fvdb/gaussian_splatting.py
```diff
@@ -1877,6 +1878,116 @@ def render_images(
...
+
+        Notes:
+            - This is **dense-only**: outputs are dense tensors of shape ``(C, H, W, ...)``.
+            - Tile intersection data is still computed from a (non-differentiable) projection
+              step, so gradients may be discontinuous when Gaussians move between tiles.
```
harrism: **❓ question:** ‏What do you mean "move between tiles"? Do you mean when they project to multiple tiles? I'm trying to understand what a discontinuous gradient means.

src/tests/GaussianRasterizeForwardTest.cpp
```diff
@@ -290,6 +302,127 @@ struct GaussianRasterizeForwardTestFixture : public ::testing::Test {
     uint32_t tileSize;
 };
 
+TEST(GaussianRasterizeForwardMaskedEdgeTile, Child) {
+#if !defined(__linux__)
+    GTEST_SKIP() << "This regression test is Linux-only.";
```
swahtz, line 307: The whole project is linux only yeah?

src/tests/GaussianRasterizeForwardTest.cpp
```diff
@@ -290,6 +302,127 @@ struct GaussianRasterizeForwardTestFixture : public ::testing::Test {
...
+    const auto outAlphasCpu   = outAlphas.cpu();
+
+    EXPECT_EQ(outAlphasCpu[0][16][16][0].item<float>(), 0.0f);
+    EXPECT_EQ(outFeaturesCpu[0][16][16][0].item<float>(), 0.1f);
+    EXPECT_EQ(outFeaturesCpu[0][16][16][1].item<float>(), -0.2f);
+    EXPECT_EQ(outFeaturesCpu[0][16][16][2].item<float>(), 0.3f);
+#endif
+}
+
+TEST(GaussianRasterizeForwardMaskedEdgeTile, NoDeadlock) {
```
harrism, line 380: **❓ question:** ‏Why are we testing for deadlock between multiple processes running the gtests? Is this actually a risk? Is there some motivation for this test?

src/fvdb/GaussianSplat3d.h
```diff
@@ -905,6 +902,34 @@ class GaussianSplat3d {
...
+    /// @brief Render images by rasterizing directly from world-space 3D Gaussians.
+    ///
+    /// This is similar to @ref renderImages but performs rasterization directly from world-space
+    /// Gaussians (means/quats/log-scales) rather than from their 2D projections. This enables
+    /// geometry gradients through the rasterization step.
+    ///
+    /// Tile intersections are still computed using a (non-differentiable) projection step:
+    /// - For `cameraModel == CameraModel::PINHOLE` or `CameraModel::ORTHOGRAPHIC`, we reuse the
+    ///   classic projection path.
+    /// - For OpenCV camera models, we use the UT projection kernel to compute per-Gaussian radii
```
harrism: **💡 suggestion:** ‏Spell out Unscented Transform.

src/fvdb/detail/ops/gsplat/GaussianRasterizeForward.cu
fwilliams: Good catch - let me add some more tests

src/tests/GaussianRasterizeForwardTest.cpp
fwilliams: Yeah it is - I (copilot) found a deadlock and we need this for threading functionality. We could probably rewrite this using the standard library but since it's linux only and this is test code, I think it's okay.

tests/unit/test_rasterize_from_world.py
```diff
@@ -0,0 +1,803 @@
...
+        sh_degree_to_use=1,
+        tile_size=16,
+        min_radius_2d=0.0,
+        eps_2d=0.3,
+        antialias=False,
+        backgrounds=None,
+    )
+
+    loss = (rendered * rendered).sum() + alphas.sum()
+    loss.backward()
```
harrism, line 440: **💡 suggestion:** ‏This code (define means, quats, scales, create GS3D, create matrices, render_images_From_world, compute loss) is repeated in a lot of tests in this file. It would save a lot of repetition to provide a helper function that takes the numbers that vary as inputs and runs the common code.

tests/unit/test_rasterize_from_world.py
```diff
@@ -0,0 +1,803 @@
...
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
+def test_gaussiansplat3d_render_images_from_world_masks_edge_tile_does_not_deadlock():
+    """
+    Regression test: masked *edge tiles* must not deadlock.
+
+    Edge tiles (where image dimensions are not multiples of tile size) have threads that are
+    `!inside`, but the kernel still uses block-level barriers. Any early-return taken by only
+    the `inside` threads (e.g. for masked tiles) can deadlock the whole kernel launch.
```
harrism, line 677: **🚩 issue:** ‏Really? Can't we write the kernel to be deadlock free?

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
```diff
@@ -0,0 +1,283 @@
...
+#include <fvdb/detail/ops/gsplat/GaussianCameraModels.h>
+#include <fvdb/detail/ops/gsplat/GaussianUtils.cuh>
+
+#include <nanovdb/math/Math.h>
+
+#include <cstdint>
+
+namespace fvdb::detail::ops {
+
+constexpr __device__ float kAlphaThreshold = 0.999f;
```
harrism, line 25: **❓ question:** ‏Would this be better as `__constant__`?

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
```diff
@@ -0,0 +1,283 @@
...
+
+template <typename T> struct WorldRay {
```
harrism: **☑️ todo:** ‏Nothing in this file is documented. Please document the classes and functions in this file.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
```diff
@@ -0,0 +1,283 @@
...
+    if (!(n2 > T(0))) {
+        return nanovdb::math::Vec3<T>(T(0), T(0), T(0));
+    }
+    const T n    = sqrt(n2);
+    const T invn = T(1) / n;
+    // v_x = (I/n - x x^T / n^3) v_y
+    const T invn3 = invn * invn * invn;
+    const T xdotv = x.dot(v_y);
+    return v_y * invn - x * (xdotv * invn3);
+}
```
harrism, line 152: **🤓 nitpick (non-blocking):** ‏For the future, GPUs have an RSQRT instruction which can be a lot faster if this is compute bound. Could look into whether it gets used or not here and whether explicit use of `rsqrt()` helps.

Also, is it sufficiently safe to make sure `n2 > 0`? Or do we need to make sure it's greater than some epsilon?

src/fvdb/GaussianSplat3d.h
fwilliams: Yeah fair enough - I agree we ought to reconsider the naming of methods in this class anyway and that should be a bigger PR with a discussion

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -0,0 +1,554 @@
...
+template <uint32_t NUM_CHANNELS> struct SharedGaussian {
+    int32_t id;                       // flattened id in [0, C*N)
+    nanovdb::math::Vec3<float> mean;  // world mean
+    nanovdb::math::Mat3<float> isclR; // S^{-1} R^T
+    float opacity;
+};
+
+template <uint32_t NUM_CHANNELS>
+__global__ void
+rasterizeFromWorld3DGSForwardKernel(
```
harrism: **🚩 issue:** ‏This kernel and the dispatch process doesn't match the existing rasterizer kernels style. Specifically, they have a rasterizer args class that derives from RasterizerCommonArgs, and the kernel only takes that structure as a single parameter.

means, conics, opacities, features, masks. backgrounds etc. are all part of common args.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -0,0 +1,554 @@
...
+    // Gaussians
+    const torch::PackedTensorAccessor32<float, 2, torch::RestrictPtrTraits> means,     // [N,3]
```
swahtz: This should use PackedTensorAccessor64 like the other rasterization kernels

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
swahtz: This kernel should probably be refactored to match the other rasterization kernels that use a `RasterizeComonArgs` (or derive a child class from it) and implement the render function there.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -0,0 +1,554 @@
...
+            for (uint32_t k = 0; k < NUM_CHANNELS; ++k) {
+                outFeatures[imgBaseFeat + k] =
+                    (backgrounds != nullptr) ? backgrounds[camId * NUM_CHANNELS + k] : 0.0f;
+            }
+        }
+        return;
+    }
+
+    // Build camera pose for this camera (start/end).
+    nanovdb::math::Mat3<float> R_wc_start, R_wc_end;
```
swahtz: We should name these variables in a consistent style to our typical descriptive variable naming.    e.g. worldToCamRotationStart (or some such)

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -0,0 +1,554 @@
...
+        return;
+    }
+
+    // Shared memory for batched gaussians.
+    extern __shared__ char smem[];
+    int32_t *idBatch = reinterpret_cast<int32_t *>(smem);                      // [blockSize]
+    auto *gaussBatch =
+        reinterpret_cast<SharedGaussian<NUM_CHANNELS> *>(&idBatch[blockSize]); // [blockSize]
+
+    float T        = 1.0f;
```
swahtz: We should name this transmission (or accumulatedTransmission?)

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -0,0 +1,554 @@
...
+
+        block.sync();
+
+        const uint32_t batchSize =
+            min((uint32_t)blockSize, (uint32_t)max(0, rangeEnd - batchStart));
+        for (uint32_t t = 0; (t < batchSize) && !done; ++t) {
+            const SharedGaussian<NUM_CHANNELS> g   = gaussBatch[t];
+            const nanovdb::math::Vec3<float> gro   = g.isclR * (ray.origin - g.mean);
+            const nanovdb::math::Vec3<float> grd   = normalizeSafe<float>(g.isclR * ray.dir);
+            const nanovdb::math::Vec3<float> gcrod = grd.cross(gro);
```
swahtz: I think it might be helpful to reference that these calculations are computing fig 11 from the 3dgut paper (or at least that's my interpretation thus far reading through) and might help disambiguate the variable names

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Good catch

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -0,0 +1,554 @@
...
+    const uint32_t imgBasePix = (camId * imageHeight * imageWidth + pixId); // alpha/last base
+
+    // Parity with classic rasterizer: masked tiles write background and exit.
+    //
+    // IMPORTANT: this kernel uses block-level barriers later (`__syncthreads_count`, `block.sync`).
+    // Any early return must be taken by *all* threads in the block, otherwise edge tiles can
+    // deadlock when some threads are `!inside`. So we make the return block-wide.
+    const bool tileMasked =
+        (masks != nullptr) &&
+        (!masks[camId * tileExtentH * tileExtentW + tileRow * tileExtentW + tileCol]);
```
harrism: **💡 suggestion:** ‏Store this index in a `tileId` variable to make this code more readable and make it more obvious this is block-uniform.

Was the deadlock potential something that actually happened in practice (I see gsplat's implementation has a potential deadlock)?

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -0,0 +1,554 @@
...
+        const float m02e = worldToCamEnd[camId][0][2];
+        const float m10e = worldToCamEnd[camId][1][0];
+        const float m11e = worldToCamEnd[camId][1][1];
+        const float m12e = worldToCamEnd[camId][1][2];
+        const float m20e = worldToCamEnd[camId][2][0];
+        const float m21e = worldToCamEnd[camId][2][1];
+        const float m22e = worldToCamEnd[camId][2][2];
+        R_wc_end = nanovdb::math::Mat3<float>(m00e, m01e, m02e, m10e, m11e, m12e, m20e, m21e, m22e);
+        t_wc_end = nanovdb::math::Vec3<float>(
+            worldToCamEnd[camId][0][3], worldToCamEnd[camId][1][3], worldToCamEnd[camId][2][3]);
```
harrism: **💡 suggestion:** ‏This is pretty verbose. No need to extract into variables first. Better yet, provide a utility since you do it twice.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -0,0 +1,554 @@
...
+    torch::Tensor opacitiesBatched = opacities;
+    if (opacitiesBatched.dim() == 1) {
+        TORCH_CHECK_VALUE(opacitiesBatched.size(0) == N,
+                          "opacities must have shape [N] or [C,N] matching N");
+        opacitiesBatched = opacitiesBatched.unsqueeze(0).repeat({C, 1});
+    }
+    TORCH_CHECK_VALUE(opacitiesBatched.dim() == 2, "opacities must have shape [C,N]");
+    TORCH_CHECK_VALUE(opacitiesBatched.size(0) == C && opacitiesBatched.size(1) == N,
+                      "opacities must have shape [C,N] matching features and N");
+    opacitiesBatched = opacitiesBatched.contiguous();
```
swahtz, line 369: We should leave a TODO/Issue to be able to just create a view of the opacities (if they're shared across cameras) to be used by the kernel in a style similar to how PR #451 was recently implemented for the other rasterization functions.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
```diff
@@ -0,0 +1,283 @@
...
+    return S_inv * R.transpose();
+}
+
+template <typename T>
+inline __device__ void
+isclRotVectorJacobianProduct(const nanovdb::math::Vec4<T> &quat_wxyz,
+                             const nanovdb::math::Vec3<T> &scale,
+                             const nanovdb::math::Mat3<T> &dLossDIsclRot,
+                             nanovdb::math::Vec4<T> &dLossDQuat,
+                             nanovdb::math::Vec3<T> &dLossDLogScale) {
```
swahtz, line 245: **☑️ todo:** ‏ This function should just return these values to be consistent with functions above and other functions in the library like `transformCovarianceWorldToCamVectorJacobianProduct`

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
swahtz: There's actually quite a lot of variable naming that is the short-style we moved away from (possibly inherited from glsplat?).  It's worth doing a pass over at some point to have the same style of names as the other rasterization kernels.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -0,0 +1,554 @@
...
+        rangeEnd = tileOffsets[camId][tileRow][tileCol + 1];
+    } else {
+        // wrap to next row/camera
+        if (tileRow + 1 < tileExtentH) {
+            rangeEnd = tileOffsets[camId][tileRow + 1][0];
+        } else {
+            // next camera
+            rangeEnd = tileOffsets[camId + 1][0][0];
+        }
+    }
```
harrism: **☑️ todo:** ‏The gaussian range logic should be the same as existing rasterization, right? That is implemented in `RasterizeCommonArgs::tileGaussianRange()`.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -0,0 +1,554 @@
...
+            const int32_t cid                      = flatId / (int32_t)means.size(0);
+            const float op                         = opacities[cid][gid];
+
+            gaussBatch[threadRank].id      = flatId;
+            gaussBatch[threadRank].mean    = mean_w;
+            gaussBatch[threadRank].isclR   = isclR;
+            gaussBatch[threadRank].opacity = op;
+        }
+
+        block.sync();
```
harrism: **💡 suggestion:** ‏Be consistent -- above you have `__syncthreads_count()` but here `block.sync()`. Is there a cg:: equivalent of `__syncthreads_count()`? Or you could just use `__syncthreads()` here.

swahtz (commented):
There's quite a few inconsistencies in this rasterization kernel compared to the precedents set by the other rasterization kernels (building a RasterizeCommonArgs, variable naming and documentation of functions).  I'd prefer to see those specific aspects addressed to be more maintainable and other comments could be left for later issues.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.h
fwilliams: Addressed: rasterize-from-world dispatch now takes a RenderSettings (matching other rasterizers). Deferred: rolling CameraModel / rolling-shutter into RenderSettings is a broader API change; keeping explicit parameters here for now.

src/fvdb/detail/ops/gsplat/GaussianRasterizeForward.cu
fwilliams: Addressed: added coverage for `masks` behavior (incl. masked edge-tile regression) in `src/tests/GaussianRasterizeForwardTest.cpp` (and 3D-from-world mask behavior in python unit tests).

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.h
fwilliams: Ack — thanks.

src/fvdb/GaussianSplat3d.h
fwilliams: Deferred: agreed the naming could be clearer, but renaming is a public API change. Suggest follow-up issue to evaluate alternatives without blocking this PR.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Addressed: switched the rasterize-from-world 3DGS forward/backward kernels to `PackedTensorAccessor64` for consistency (see `ef7beb4`).

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Deferred: refactoring to `RasterizeCommonArgs` would be a bigger structural change. Added a TODO near the kernel and kept this PR focused on correctness + parity.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Addressed: renamed camera pose locals to descriptive names (e.g. `worldToCamRotationStart/End`, `worldToCamTranslationStart/End`).

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Addressed: renamed the accumulator to `transmittance` for clarity.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Addressed: added a short comment mapping the whitened-space visibility calc to 3D-GUT Fig. 11 (to disambiguate gro/grd/gcrod).

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Addressed: added a TODO to avoid materializing a repeated [C,N] tensor when opacities are shared across cameras (similar to PR #451 approach).

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
fwilliams: Deferred (small follow-up): changing this VJP helper to return values would be a signature change; added a TODO suggesting that for consistency with other VJP helpers.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Addressed in part: did a pass to move the most confusing locals toward our descriptive style; remaining naming cleanups can be a follow-up to keep this PR scoped.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.h
fwilliams: Ack — thanks.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Addressed in part: I renamed some of the most ambiguous locals to our more descriptive style (and added TODOs for the rest). A full naming sweep is a follow-up so we can keep this PR scoped.

fvdb/gaussian_splatting.py
fwilliams: Addressed in part: I spelled out Unscented Transform (UT) in the python docstring. I kept the OpenCV camera-model wording concise here, but can add a short pointer to `fvdb.enums.CameraModel` docs in a follow-up if you want more detail inline.

fvdb/gaussian_splatting.py
fwilliams: Addressed: clarified this note to explicitly call out that discontinuities happen when small parameter changes cause a Gaussian to enter/leave a tile (or change overlap set).

src/tests/GaussianRasterizeForwardTest.cpp
fwilliams: Good question. This test intentionally uses a child process so a true kernel deadlock can be detected via timeout without hanging the entire gtest runner. A single-process deadlock test would wedge the whole suite.

src/fvdb/GaussianSplat3d.h
fwilliams: Addressed: spelled out Unscented Transform (UT) in the C++ API doc comment as well.

tests/unit/test_rasterize_from_world.py
fwilliams: Deferred for scope: agreed this file has repetition and could use helper extraction, but I kept this PR focused on correctness/parity and review-blocking issues. Happy to do that cleanup in a follow-up PR.

tests/unit/test_rasterize_from_world.py
fwilliams: Agreed. The kernel has been updated to be deadlock-safe (block-uniform early exits around barriers). This test remains as a regression guard so we don’t accidentally reintroduce the bug.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
fwilliams: No code change for now: `constexpr __device__` is already compile-time constant and inlined where used. If we see perf pressure from this path, we can benchmark a `__constant__` variant separately.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
fwilliams: Addressed: added API comments for the main constants/helpers in `GaussianRasterizeFromWorld.cuh` to document intent and usage.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
fwilliams: Deferred (non-blocking): I left `normalizeSafe` behavior unchanged for now. I agree this is worth profiling and possibly tightening with rsqrt/epsilon, but I’d prefer to do that in a focused perf/numerics follow-up.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Deferred: agreed this would be cleaner with a `RasterizeCommonArgs`-style path, but that’s a larger structural refactor. I’ve left this as a follow-up item to keep this PR scoped.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Addressed: introduced a block-uniform `tileId` local for readability in the masked-tile check.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Deferred for now: agreed this extraction is verbose and a utility would be better. I kept behavior stable in this PR and can fold this into the follow-up kernel-structure cleanup.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Deferred with intent: yes, this should ideally share the `tileGaussianRange()` path once the kernel is moved toward `RasterizeCommonArgs`. I’ve kept existing logic here to avoid broad refactor risk in this PR.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Addressed: made sync usage consistent at this point by switching to `__syncthreads()` here while keeping `__syncthreads_count(done)` for the early all-done exit logic.

swahtz (dismissed):
Thank you for making some of the requested changes.

I think in the next steps, we should first refactor this to match the patterns of other rasterization functions (`RasterizeCommonArgs`, etc.).  After that I think we should discuss with @harrism refactoring of common functionality.  There's a few changes I've also wanted to be make to the general pipeline but have held off such as separating Camera definition settings; Rasterizer method-agnostic settings and Rasterizer-specific settings in a semantically useful way as well as build in the ability use other shading evaluation methods and not just force all rasterization to use the SH evaluation path.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Addressed in `6721e4f`: the from-world forward/backward kernels now use a shared `RasterizeFromWorldCommonArgs` helper that centralizes dense tile coordinates, tile range lookup, and optional input plumbing. This brings the implementation much closer to the `RasterizeCommonArgs` pattern used by the other rasterizers while preserving current behavior.

tests/unit/test_rasterize_from_world.py
fwilliams: Addressed in `46d4cde`: deduplicated `tests/unit/test_rasterize_from_world.py` by factoring shared helpers for GS construction, camera setup, render invocation, and common gradient assertions. Coverage is unchanged, but the file is significantly easier to read/maintain.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Addressed in `6721e4f`: tile gaussian range computation is now routed through the shared `RasterizeFromWorldCommonArgs::tileGaussianRange(...)` helper used by both forward and backward kernels, so this logic is no longer duplicated per kernel.

tests/unit/test_rasterize_from_world.py
fwilliams: Kept this regression because it guards a real deadlock mode (masked edge tiles + block barriers). I did reduce surrounding test duplication in `46d4cde` so the test file is cleaner now, but I believe this subprocess guard should remain to prevent accidental reintroduction.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
fwilliams: Agree this is worth investigating, but I am keeping rsqrt/normalization tuning out of this PR to avoid blending perf/numerics changes with the readability + structural refactors. I suggest a follow-up perf pass with benchmarks + tolerance checks before switching implementations.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
fwilliams: Reconsidered after this refactor: I still prefer to defer converting this to `__constant__` in this PR. With current code shape, this is low expected gain and would be better evaluated with a focused perf microbenchmark (to avoid mixing behavioral/readability changes with perf tuning).

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
fwilliams: I did another documentation pass while implementing the shared from-world args refactor (`6721e4f`), and the key helpers/structs in `GaussianRasterizeFromWorld.cuh` now have API comments. If there are still specific gaps you want documented in that file, I can add those directly.

src/fvdb/detail/ops/gsplat/GaussianCameraSharedMemory.cuh
```diff
@@ -0,0 +1,72 @@
...
+
+#include <nanovdb/math/Math.h>
+
+#include <cstdint>
+
+namespace fvdb::detail::ops {
+
+template <typename ScalarType, typename Acc33>
+inline __device__ void
+copyMat3AccessorToShared(const int64_t C,
```
swahtz: **❓ question:** ‏These functions aren't specific to shared memory, they're just helpers in later functions that happen to use them for shared memory copies, right?  So can their names be more generic like `copyBatchCameraAccessorToMat3`

fwilliams: Addressed in the latest update. I renamed these helpers to generic accessor/math-oriented names since they are utility copy helpers and not inherently shared-memory-specific:
- `copyBatchAccessorToMat3`
- `copyBatchPoseAccessorRotationToMat3`
- `copyBatchPoseAccessorTranslationToVec3`
- `copyBatchAccessorToFlatRowMajor`

I also updated all call sites (classic projection + UT projection) to keep naming consistent.

src/benchmarks/dispatch/omp_gelu.cpp
```diff
@@ -9,7 +9,9 @@
 
 #include <ATen/cpu/vec/vec.h>
 
+#ifdef _OPENMP
```
harrism, line 12: **❓ question:** ‏Why is this change in this PR?

src/fvdb/detail/ops/gsplat/GaussianCameraSharedMemory.cuh
harrism: **💡 suggestion:** ‏Remove "To Shared" from the name of this function and the others in this file. Rename the file.

src/fvdb/detail/ops/gsplat/GaussianCameraSharedMemory.cuh
```diff
@@ -0,0 +1,72 @@
...
+                               const Acc44 &acc /* [C,4,4] */) {
+    constexpr int64_t kElementsPerMat3 = 9;
+    for (int64_t i = threadIdx.x; i < C * kElementsPerMat3; i += blockDim.x) {
+        const int64_t camId      = i / kElementsPerMat3;
+        const int64_t entryId    = i % kElementsPerMat3;
+        const int64_t rowId      = entryId / 3;
+        const int64_t colId      = entryId % 3;
+        out[camId][rowId][colId] = acc[camId][rowId][colId];
+    }
+}
```
harrism: **💡 suggestion:** ‏Remove this function and just use `copyMat3Accessor`. It's templated on accessor type and has the same implementation as `copyMat3Accessor` above so the compiler can generate this implementation for either Mat3 or Mat4 accessor inputs.

src/fvdb/detail/ops/gsplat/GaussianOpenCVDistortion.cuh
```diff
@@ -0,0 +1,125 @@
...
+/// - Unused coefficients for a given model are ignored (but still read from the packed array).
+///
+/// @param cameraModel Camera model selector.
+/// @param p_normalized Normalized camera-plane coordinate.
+///   - perspective: (x/z, y/z)
+///   - orthographic: (x, y)
+/// @param distortionCoeffs Pointer to packed coefficients (may be null for non-OpenCV models).
+/// @param numCoeffs Number of coefficients per camera (0 for PINHOLE/ORTHOGRAPHIC, 12 for OPENCV).
+template <typename T>
+inline __host__ __device__ nanovdb::math::Vec2<T>
```
harrism: **💡 suggestion:** ‏If this is only ever called from device code, make it device only.

```suggestion
inline __device__ nanovdb::math::Vec2<T>
```

src/fvdb/detail/ops/gsplat/GaussianOpenCVDistortion.cuh
```diff
@@ -0,0 +1,125 @@
...
+    return nanovdb::math::Vec2<T>(x_dist, y_dist);
+}
+
+/// @brief Iteratively undistort a packed OpenCV distortion model using fixed-point iteration.
+///
+/// This solves for `x` such that `applyOpenCVDistortionPacked(x) == p_distorted`.
+/// It uses a small fixed iteration count (matching the UT kernel style) and is intended for use in
+/// ray generation.
+template <typename T>
+inline __host__ __device__ nanovdb::math::Vec2<T>
```
harrism: **💡 suggestion:** ‏Device-only.


```suggestion
inline __device__ nanovdb::math::Vec2<T>
```

src/fvdb/detail/ops/gsplat/GaussianProjectionUtils.cuh
```diff
@@ -0,0 +1,109 @@
...
+#include <fvdb/detail/ops/gsplat/GaussianCameraMatrixUtils.cuh>
+#include <fvdb/detail/ops/gsplat/GaussianIntrinsics.cuh>
+#include <fvdb/detail/ops/gsplat/GaussianUtils.cuh>
+
+#include <nanovdb/math/Math.h>
+
+#include <cuda/std/cmath>
+
+namespace fvdb::detail::ops {
+
```
harrism, line 16: **☑️ todo:** ‏All functions in this file need docs.

src/fvdb/detail/ops/gsplat/GaussianProjectionUtils.cuh
```diff
@@ -0,0 +1,109 @@
...
+    // Matches the classic FVDB/gsplat heuristic: use the largest eigenvalue of the 2x2 covariance
+    // (via trace/determinant) and take `sigma` standard deviations.
+    const T b  = T(0.5) * (covar2d[0][0] + covar2d[1][1]);
+    const T v1 = b + ::cuda::std::sqrt(nanovdb::math::Max(detClamp, b * b - det));
+    return ::cuda::std::ceil(sigma * ::cuda::std::sqrt(v1));
+}
+
+template <typename T>
+inline __device__ void
+storeConicRowMajor3(const nanovdb::math::Mat2<T> &covar2dInverse, T *outConic3) {
```
harrism: **❓ question:** ‏Could the output just be a vec3, and could you just return it? Edit: I see `packConicRowMajor3` does what I suggest. Do we need both?

src/fvdb/detail/ops/gsplat/GaussianProjectionUtils.cuh
```diff
@@ -0,0 +1,109 @@
...
+    outConic3[0] = covar2dInverse[0][0];
+    outConic3[1] = covar2dInverse[0][1];
+    outConic3[2] = covar2dInverse[1][1];
+}
```
harrism: **❓ question:** ‏I don't know what the math is here, but it seems weird to get the top left, middle left, and bottom middle of a 3x3 into a vector. Is this correct? Docs would help.

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
```diff
@@ -185,17 +164,16 @@ projectionBackwardKernel(const int32_t offset,
     const nanovdb::math::Mat3<T> &covarCamSpace = transformCovarianceWorldToCam(R, covar);
 
     // vjp: camera projection
-    const T fx = projectionMatrices[0], cx = projectionMatrices[2], fy = projectionMatrices[4],
-            cy                                     = projectionMatrices[5];
+    const CameraIntrinsics<T> intr                 = loadIntrinsicsRowMajor3x3(projectionMatrices);
```
harrism: **💡 suggestion:** Avoid unhelpful abbreviation.

```suggestion
    const CameraIntrinsics<T> intrinsics                = loadIntrinsicsRowMajor3x3(projectionMatrices);
```

src/fvdb/detail/ops/gsplat/GaussianProjectionForward.cu
```diff
@@ -136,15 +138,26 @@ template <typename T, bool Ortho> struct ProjectionForward {
         const Mat3 covarCamSpace = transformCovarianceWorldToCam(worldToCamRotMatrix, covar);
 
         // camera projection
-        const T fx = projectionMatrix[0][0], cx = projectionMatrix[0][2],
-                fy = projectionMatrix[1][1], cy = projectionMatrix[1][2];
-        auto [covar2d, mean2d] = [&]() {
+        const CameraIntrinsics<T> intr = loadIntrinsics(projectionMatrix);
```
harrism:
```suggestion
        const CameraIntrinsics<T> intrinsics = loadIntrinsics(projectionMatrix);
```

src/fvdb/detail/ops/gsplat/GaussianIntrinsics.cuh
```diff
@@ -0,0 +1,29 @@
...
+template <typename T>
+inline __host__ __device__ CameraIntrinsics<T>
+loadIntrinsics(const nanovdb::math::Mat3<T> &K) {
+    return CameraIntrinsics<T>{K[0][0], K[1][1], K[0][2], K[1][2]};
+}
+
+template <typename T>
+inline __host__ __device__ CameraIntrinsics<T>
+loadIntrinsicsRowMajor3x3(const T *K9) {
+    return CameraIntrinsics<T>{K9[0], K9[4], K9[2], K9[5]};
```
harrism: **❓ question:** ‏Do we have a conversion between `Mat3<T>` and `const T*`? If so, we wouldn't need both of these functions...

src/fvdb/detail/ops/gsplat/GaussianProjectionForward.cu
harrism: **💡 suggestion:** ‏`CameraIntrinsics<T>` is just a a POD struct around 4 floats. Give it constructors to load the 4 values from a matrix and delete the `loadIntrinsics` free functions. Either that or get rid of the struct and have `loadIntrinsics()` return a 4-tuple, so that you can just write

```
const T [fx, cx, fy, cy] = loadIntrinsics(...);
```

src/fvdb/detail/ops/gsplat/GaussianProjectionForward.cu
```diff
@@ -157,69 +170,49 @@ template <typename T, bool Ortho> struct ProjectionForward {
 
         const Mat2 covar2dInverse = covar2d.inverse();
 
-        // take 3 sigma as the radius
-        const T b      = 0.5f * (covar2d[0][0] + covar2d[1][1]);
-        const T v1     = b + sqrt(max(0.01f, b * b - det));
-        const T radius = ceil(3.f * sqrt(v1));
+        // take 3 sigma as the radius (non-differentiable)
```
harrism, line 174: **💡 suggestion:** ‏This comment should be the docs for the new function.

src/fvdb/detail/ops/gsplat/GaussianProjectionForward.cu
```diff
@@ -157,69 +170,49 @@ template <typename T, bool Ortho> struct ProjectionForward {
...
-                worldToCamRotMatsShared[camId][rowId][colId] =
-                    mWorldToCamMatricesAcc[camId][rowId][colId];
-            } else {
-                const auto baseIdx                    = i - 2 * C * 9;
-                const auto camId                      = baseIdx / 3;
-                const auto entryId                    = baseIdx % 3;
-                worldToCamTranslation[camId][entryId] = mWorldToCamMatricesAcc[camId][entryId][3];
-            }
-        }
+        uint8_t *ptr         = reinterpret_cast<uint8_t *>(sharedMemory);
```
harrism: **💡 suggestion:** ‏IMO the original shared memory offset code above was better. Only 3 `reinterpret_cast`s instead of 4, and 3 very clear lines of code instead of 7.

src/fvdb/detail/ops/gsplat/GaussianCameraMatrixUtils.cuh
```diff
@@ -0,0 +1,61 @@
...
+#ifndef FVDB_DETAIL_OPS_GSPLAT_GAUSSIANCAMERAMATRIXUTILS_CUH
+#define FVDB_DETAIL_OPS_GSPLAT_GAUSSIANCAMERAMATRIXUTILS_CUH
+
+#include <nanovdb/math/Math.h>
+
+namespace fvdb::detail::ops {
+
+template <typename T>
+inline __host__ __device__ void
+loadWorldToCamRtRowMajor4x4(const T *m44, nanovdb::math::Mat3<T> &R, nanovdb::math::Vec3<T> &t) {
```
swahtz: **💡 suggestion:** ‏ To be consistent with other Utils functions that need to return multiple values, we should just return a tuple of the results.  This pattern is what we already switched to in our other Utils (e.g. `quaternionAndScaleToCovarianceVectorJacobianProduct`, `transformPointWorldToCamVectorJacobianProduct`, `transformCovarianceWorldToCamVectorJacobianProduct`, etc.)

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
```diff
@@ -0,0 +1,254 @@
...
+
+#include <cstdint>
+
+namespace fvdb::detail::ops {
+
+/// Opacity threshold used by 3DGS alpha compositing.
+constexpr __device__ float kAlphaThreshold = 0.999f;
+
+/// Common dense-tile rasterization arguments shared by from-world forward/backward kernels.
+struct RasterizeFromWorldCommonArgs {
```
harrism, line 28: **🚩 issue:** ‏The idea with `RasterizeCommonArgs` is it's a struct that the various flavors of GaussianRasterize kernels (forward, backward, NumContributingGaussians forward and backward, etc.) share because the arguments are common. This struct creates yet another "CommonArgs" that duplicates those arguments. Instead, There should be a `RasterizeFromWorldArgs` that *has a* `RasterizeCommonArgs` struct as a member, as we do in all the other uses.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
```diff
@@ -0,0 +1,254 @@
...
+        return cameraId * numTilesH * numTilesW + tileRow * numTilesW + tileCol;
+    }
+
+    inline __device__ bool
+    tileMasked(const uint32_t cameraId, const uint32_t tileRow, const uint32_t tileCol) const {
+        return (masks != nullptr) && (!masks[tileId(cameraId, tileRow, tileCol)]);
+    }
+
+    inline __device__ cuda::std::tuple<int32_t, int32_t>
+    tileGaussianRange(const uint32_t cameraId,
```
harrism, line 93: **❓ question:** ‏I'm surprised that this function and `denseCoordinates` aren't the same as our other implementations. Why are they different?

src/fvdb/detail/ops/gsplat/GaussianRollingShutter.cuh
```diff
@@ -0,0 +1,53 @@
...
+///
+/// FVDB convention:
+/// - vertical:    time varies with image row (y)
+/// - horizontal:  time varies with image column (x)
+/// - none:        time is 0
+///
+/// Pixel coordinates are in image space (not normalized), and we use `floor` (matching existing
+/// kernels) before normalizing by (dim-1).
+template <typename T>
+inline __host__ __device__ T
```
harrism: **💡 suggestion:** ‏Does this need to be `__host__`?


```suggestion
inline __device__ T
```

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -0,0 +1,471 @@
...
+template <uint32_t NUM_CHANNELS> struct SharedGaussian {
+    int32_t id;                       // flattened id in [0, C*N)
+    nanovdb::math::Vec3<float> mean;  // world mean
+    nanovdb::math::Mat3<float> isclR; // S^{-1} R^T
+    float opacity;
+};
+
+template <uint32_t NUM_CHANNELS>
+__global__ void
+rasterizeFromWorld3DGSForwardKernel(
```
harrism: **💡 suggestion:** ‏Suggest matching naming style of other raster kernels. Now need to name a kernel "Kernel", it's obvious it's a kernel.


```suggestion
rasterizeGaussiansFromWorld(
```

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -0,0 +1,471 @@
...
+    // Per-camera
+    const torch::PackedTensorAccessor64<float, 3, torch::RestrictPtrTraits> features,  // [C,N,D]
+    const torch::PackedTensorAccessor64<float, 2, torch::RestrictPtrTraits> opacities, // [C,N]
+    const torch::PackedTensorAccessor64<float, 3, torch::RestrictPtrTraits>
+        worldToCamStart, // [C,4,4] (view as 3D accessor for contiguous)
+    const torch::PackedTensorAccessor64<float, 3, torch::RestrictPtrTraits>
+        worldToCamEnd,   // [C,4,4]
+    const torch::PackedTensorAccessor64<float, 3, torch::RestrictPtrTraits> K, // [C,3,3]
+    const torch::PackedTensorAccessor64<float, 2, torch::RestrictPtrTraits>
+        distortionCoeffs,                                                      // [C,K]
```
harrism: **💡 suggestion:** ‏Our other rasterization kernels pack all the args into the Args struct (hence the name)...

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -0,0 +1,471 @@
...
+    int32_t id;                       // flattened id in [0, C*N)
+    nanovdb::math::Vec3<float> mean;  // world mean
+    nanovdb::math::Mat3<float> isclR; // S^{-1} R^T
+    float opacity;
+};
+
+template <uint32_t NUM_CHANNELS>
+__global__ void
+rasterizeFromWorld3DGSForwardKernel(
+    const RasterizeFromWorldCommonArgs commonArgs,
```
harrism: **💡 suggestion:** In the other rasterize kernels, ‏commonArgs is the shared / common struct, not the specific struct for the kernel. Just call it "args".

```suggestion
    const RasterizeFromWorldCommonArgs args,
```

harrism (changes requested):
This is a big improvement, but I still have quite a few suggestions. 

Also, I still don't think we need to duplicate the deadlock test in Python and C++.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
swahtz: **🚩 issue:** ‏The logic in this function should be put into a function like `RasterizeFromWorldCommonArgs::volumeRenderTileForward` and all the inputs/outputs should be members of `RasterizeFromWorldCommonArgs`.  Then this function's only arguments are the `args` and all the logic in this function happens in `volumeRenderTileForward` (with `RasterizeFromWorldCommonArgs` writing to outputs using its own functions).  `rasterizeGaussiansForward` in `GaussianRasterizeForward.cu` is an example of this pattern.

src/fvdb/detail/ops/gsplat/GaussianCameraSharedMemory.cuh
fwilliams: fixed

fwilliams: also renamed the header to GaussianCameraAccessorCopy.cuh

src/fvdb/detail/ops/gsplat/GaussianOpenCVDistortion.cuh
fwilliams: done

src/fvdb/detail/ops/gsplat/GaussianOpenCVDistortion.cuh
fwilliams: done

src/fvdb/detail/ops/gsplat/GaussianCameraSharedMemory.cuh
fwilliams: done

src/fvdb/detail/ops/gsplat/GaussianProjectionUtils.cuh
fwilliams: good catch - fixed in favor of keeping packConicRowMajor3 and deleting the other.

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
fwilliams: fixed

src/fvdb/detail/ops/gsplat/GaussianProjectionForward.cu
fwilliams: fixed

src/fvdb/detail/ops/gsplat/GaussianIntrinsics.cuh
fwilliams: In principle, we don't need both and can convert between Mat3 and pointers but  converting `T*` -> `Mat3<T>` just to call the `Mat3` overload adds extra indirection/noise in hot device code,
and converting `Mat3<T>` -> `T*` isn’t naturally safe unless you rely on layout assumptions (which is probably fine in our codebase). 

I propose keeping the two functions but renaming to `loadIntrinsicsFromMat3` and `loadIntrinsicsFromRowMajor3x3Ptr`.

fwilliams: Actually nevermind, just read your comments below and I agree. Let's just get rid of these.

src/fvdb/detail/ops/gsplat/GaussianProjectionForward.cu
fwilliams: Done using the constructor. I like the struct personally. I find it easier to parse when reading.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: fixed

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: done

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
fwilliams: Good catch. I agree with the concern about duplicating “common args” logic.
I did try adding a `RasterizeFromWorldArgs` wrapper, but it was just a thin pass-through in this path (no extra fields/behavior), so it added indirection without helping readability. I reverted that and kept `RasterizeFromWorldCommonArgs` directly.
I also kept a small readability cleanup in-kernel (`const auto& common = args`) so call sites are less noisy.

I didn’t embed `RasterizeCommonArgs` directly because that struct currently includes classic 2D/sparse-specific fields/checks that from-world doesn’t use. Pulling that in here would couple unrelated code.
If we want to unify more, the next step is extracting a smaller shared tile/image base that both classic and from-world can compose.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
fwilliams: Good catch. They were different because `RasterizeFromWorld` is dense-only and originally used a simplified mapping/range path, while classic `RasterizeCommonArgs` also handles sparse/offset cases.

I’ve now rewritten `denseCoordinates` and `tileGaussianRange` in `GaussianRasterizeFromWorld.cuh` to mirror the classic dense implementation style (same structure/naming), with no functional change. So we keep dense-only behavior here, but the code path is now much easier to compare side-by-side with the existing rasterizer.

src/fvdb/detail/ops/gsplat/GaussianProjectionUtils.cuh
fwilliams: Good call. I added docs/comments for all helpers in this file, including packed layout conventions and gradient unpacking notes.

src/fvdb/detail/ops/gsplat/GaussianProjectionUtils.cuh
fwilliams: Also clarified this directly in code: packed conic is `[a, b, c]` for the symmetric matrix `[[a, b], [b, c]]`. So those indices are intentional.

src/fvdb/detail/ops/gsplat/GaussianProjectionForward.cu
fwilliams: Agreed. I promoted that into function-level docs on the projection helper path so it’s documented where it’s used.

src/fvdb/detail/ops/gsplat/GaussianProjectionForward.cu
fwilliams: Agree — I switched this back to the simpler shared-memory offset form (fewer casts/lines, same layout/behavior).

src/benchmarks/dispatch/omp_gelu.cpp
fwilliams: This came in because CI was failing in benchmarks due to OpenMP build/link behavior on this branch. It’s not functionally tied to 3D rasterization, just unblocked CI. If you prefer stricter scope, I can split it out.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: I tried the wrapper-style RasterizeFromWorldArgs, but in this path it was just a pass-through and added indirection without adding state. I reverted it and kept RasterizeFromWorldCommonArgs directly, with const auto& common = args for readability. Also aligned denseCoordinates/tileGaussianRange to match classic dense style. If we want fuller convergence, I think the right follow-up is a shared tile/image base struct.

src/fvdb/detail/ops/gsplat/GaussianCameraMatrixUtils.cuh
fwilliams: good catch - fixed

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: done

src/fvdb/detail/ops/gsplat/GaussianCameraIntrinsics.cuh
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+#ifndef FVDB_DETAIL_OPS_GSPLAT_GAUSSIANCAMERAINTRINSICS_CUH
+#define FVDB_DETAIL_OPS_GSPLAT_GAUSSIANCAMERAINTRINSICS_CUH
+
+#include <nanovdb/math/Math.h>
+
+namespace fvdb::detail::ops {
+
+template <typename T> struct CameraIntrinsics {
```
harrism, line 16: **☑️ todo:** ‏Document this class.

src/fvdb/detail/ops/gsplat/GaussianCameraAccessorCopy.cuh
```diff
@@ -0,0 +1,57 @@
...
+#ifndef FVDB_DETAIL_OPS_GSPLAT_GAUSSIANCAMERAACCESSORCOPY_CUH
+#define FVDB_DETAIL_OPS_GSPLAT_GAUSSIANCAMERAACCESSORCOPY_CUH
+
+#include <nanovdb/math/Math.h>
+
+#include <cstdint>
+
+namespace fvdb::detail::ops {
+
+template <typename ScalarType, typename Acc33>
```
harrism, line 17: **☑️ todo:** ‏Document this file / functions.

src/fvdb/detail/ops/gsplat/GaussianCameraMatrixUtils.cuh
```diff
@@ -0,0 +1,60 @@
...
+#define FVDB_DETAIL_OPS_GSPLAT_GAUSSIANCAMERAMATRIXUTILS_CUH
+
+#include <nanovdb/math/Math.h>
+
+#include <cuda/std/tuple>
+
+namespace fvdb::detail::ops {
+
+template <typename T>
+inline __host__ __device__ cuda::std::tuple<nanovdb::math::Mat3<T>, nanovdb::math::Vec3<T>>
```
harrism, line 18: **☑️ todo:** ‏Document this file / functions.

harrism (approved):
Thanks for addressing my comments. I'm approving, but I'd like you to generate missing documentation before merging (see 3 comments).

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldBackward.cu
```diff
@@ -68,38 +68,36 @@ rasterizeFromWorld3DGSBackwardKernel(
 ) {
     auto block               = cg::this_thread_block();
     const uint32_t blockSize = blockDim.x * blockDim.y;
+    const auto &common       = args;
```
swahtz: **❓ question:** ‏Why this indirection?  Can't we just use `args` directly?

fwilliams: vestige from an old refactor where args was the member of a struct so we had to do xxx.args. The compiler will optimize this away so not impactful but I'll remove it since it's redundant

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -25,207 +25,206 @@ template <uint32_t NUM_CHANNELS> struct SharedGaussian {
...
+    torch::PackedTensorAccessor64<float, 3, torch::RestrictPtrTraits> features;         // [C,N,D]
+    torch::PackedTensorAccessor64<float, 2, torch::RestrictPtrTraits> opacities;        // [C,N]
+    torch::PackedTensorAccessor64<float, 3, torch::RestrictPtrTraits> worldToCamStart;  // [C,4,4]
+    torch::PackedTensorAccessor64<float, 3, torch::RestrictPtrTraits> worldToCamEnd;    // [C,4,4]
+    torch::PackedTensorAccessor64<float, 3, torch::RestrictPtrTraits> K;                // [C,3,3]
+    torch::PackedTensorAccessor64<float, 2, torch::RestrictPtrTraits> distortionCoeffs; // [C,K]
+    int64_t numDistCoeffs;
+    RollingShutterType rollingShutterType;
+    CameraModel cameraModel;
+    float *__restrict__ outFeatures;  // [C,H,W,D]
```
swahtz: **🚩 issue:** ‏These should be accessors following the pattern of `RasterizeForwardArgs`

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
```diff
@@ -25,207 +25,206 @@ template <uint32_t NUM_CHANNELS> struct SharedGaussian {
...
-    // Any early return must be taken by *all* threads in the block, otherwise edge tiles can
-    // deadlock when some threads are `!inside`. So we make the return block-wide.
-    const bool tileMasked = commonArgs.tileMasked(camId, tileRow, tileCol);
-    if (tileMasked) {
-        if (inside) {
-            outAlphas[imgBasePix]  = 0.0f;
-            outLastIds[imgBasePix] = -1;
+template <uint32_t NUM_CHANNELS> struct RasterizeFromWorldForwardArgs {
+    RasterizeFromWorldCommonArgs commonArgs;
+    torch::PackedTensorAccessor64<float, 2, torch::RestrictPtrTraits> means;            // [N,3]
```
swahtz: **🚩 issue:** ‏All of these inputs should be members of `RasterizeFromWorldCommonArgs` since they are common amongst the forward and backwards.  See the same pattern on `RasterizeCommonArgs` and `RasterizeForwardArgs`

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldBackward.cu
```diff
@@ -32,7 +32,7 @@ template <uint32_t NUM_CHANNELS> struct SharedGaussian {
 template <uint32_t NUM_CHANNELS>
 __global__ void
 rasterizeFromWorld3DGSBackwardKernel(
-    const RasterizeFromWorldCommonArgs commonArgs,
+    const RasterizeFromWorldCommonArgs args,
     // Gaussians
     const torch::PackedTensorAccessor64<float, 2, torch::RestrictPtrTraits> means,     // [N,3]
```
swahtz: **🚩 issue:** Same notes as about the forward kernel below.  The inputs belong in the common args and the pattern of the `RasterizeBackwardArgs` wasn't followed.‏

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldBackward.cu
fwilliams: Fixed. I switched this to the same args pattern as the other kernels, so the kernel takes a single `RasterizeFromWorldBackwardArgs` and reads shared inputs via `commonArgs` (no extra alias indirection on top of that).

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Addressed. The shared from-world inputs now use accessor members on `RasterizeFromWorldCommonArgs`, and forward/backward both consume that common args struct through their per-kernel args wrappers.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Done. I moved the shared forward/backward inputs into `RasterizeFromWorldCommonArgs` so both kernels read from the same common payload, matching the `RasterizeCommonArgs`/`RasterizeForwardArgs` pattern.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldBackward.cu
fwilliams: Done on backward too. It now mirrors forward: shared inputs live in `RasterizeFromWorldCommonArgs`, backward-specific tensors/pointers are in `RasterizeFromWorldBackwardArgs`.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldBackward.cu
```diff
@@ -29,46 +29,28 @@ template <uint32_t NUM_CHANNELS> struct SharedGaussian {
...
+    fvdb::TorchRAcc64<float, 4> dLossDRenderedFeatures; // [C,H,W,D]
+    fvdb::TorchRAcc64<float, 4> dLossDRenderedAlphas;   // [C,H,W,1]
     // Outputs (grads)
-    float *__restrict__ dMeans,     // [N,3]
-    float *__restrict__ dQuats,     // [N,4]
-    float *__restrict__ dLogScales, // [N,3]
-    float *__restrict__ dFeatures,  // [C,N,D]
-    float *__restrict__ dOpacities  // [C,N]
-) {
+    float *__restrict__ dMeans;     // [N,3]
```
swahtz: **🚩 issue:** ‏Per previous comments, please follow the pattern of the other Args classes where outputs should be accessors too

fwilliams: Good catch — updated this to follow the same args pattern. Backward outputs are now accessors in `RasterizeFromWorldBackwardArgs` (instead of raw pointers), and writes use accessor-backed pointers for atomics.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldForward.cu
fwilliams: Follow-up: I also updated the forward outputs to match the same pattern. `RasterizeFromWorldForwardArgs` now carries output accessors (`outFeatures`, `outAlphas`, `outLastIds`) instead of raw pointers, and writes are done via accessor-backed pointers/strides in-kernel.

swahtz (approved):
This is great, thanks for addressing all my notes.  In principle, we could refactor all the feature/alpha writing to `writeFeature`/`writeAlpha` functions like on the other rasterization pipelines' `Args` classes but that can be left to further cleanup.

PR #446, opened 2026-02-06
Optimize Gaussian tile intersection for mGPU

matthewdcong:
1. Using `tilesPerGaussianCumsum`, we can exactly prefetch the range of intersection keys and values needed for the subsequent `computeGaussianTileIntersections` kernel. This significantly improves the performance and variance in execution time of the kernel, going from 15 to 30ms to consistently 3ms. This results in an end to end performance increase of about 7-8%.
2. The overhead of the prefetch when merging keys in the multi-GPU radix sort was larger than the penalty occurred for (rare) page faults. Removing the prefetch marginally increases performance.
3. Some small `const` fixes


swahtz (approved):
Nice!  Thanks for that.

PR #452, opened 2026-02-09
dispatch framework `for_each`, views, and tag canonicalization

blackencino:
## `for_each`, views, and tag canonicalization

This is a large diff. I won't pretend otherwise. It touches 38 files,
adds about 6400 lines, and removes about 2300. My honest assessment is
that trying to build this up in smaller increments would have wasted
more of everyone's time in aggregate, just in less visible installments,
because the pieces here are mutually dependent.

I'd suggest reading this in three passes. The first is short and covers
only what op authors see. The second explains the CPU parallelism story.
The third is about why the internals had to change to support all of
this.

## What op authors see

The user-facing changes are small.

The include structure is reorganized. Torch-specific dispatch utilities
live under `dispatch/torch/` now:

```cpp
#include "dispatch/torch/dispatch.h"   // device guard, contiguity helpers
#include "dispatch/torch/for_each.h"   // parallel iteration
#include "dispatch/torch/views.h"      // flat_in, flat_out, tensor_in, tensor_out
#include "dispatch/torch/types.h"      // scalar type mappings, axis typedefs
```

Axes that share the same underlying C++ type now need labels to
distinguish them. If you have two axes both typed `torch::ScalarType`,
the system needs a way to tell them apart at compile time. For most ops
this never comes up because each axis already has a distinct type. Where
it does come up (the sparse convolution case discussed below), you wrap
one in `named<"your.label", T>`.

Tags must have unique types. `tag<torch::kCPU, torch::kFloat>` is fine
because `DeviceType` and `ScalarType` are different types.
`tag<torch::kFloat, torch::kDouble>` is rejected at compile time
because both are `ScalarType`.

`with_value` replaces the old `tag_match_one`. The concept names are
`with_value<Tag, V>` (tag contains this specific value) and
`with_type<Tag, T>` (tag contains some value of this type).
`with_value` is defined in terms of `with_type`, so C++20 concept
subsumption works correctly: a function constrained with `with_value`
is more specific than one constrained with only `with_type`.

That's it for conceptual changes.

## `for_each` and views: writing less op code

The point of this PR, from an op author's perspective, is that
elementwise ops get much shorter.

Compare `relu.cu` (the "before" pattern) with `softplus.cu` (the
"after" pattern). Both do the same category of work: apply a scalar
function elementwise across a tensor.

`relu.cu` requires:

- Two concept-constrained scalar overloads (one for half types, one
  for builtin floats)
- A separate `__global__` CUDA kernel function
- Two `op()` overloads inside the op struct: one for CPU with a manual
  serial loop, one for CUDA with manual device guard, stream
  acquisition, grid/block computation, kernel launch, and
  `C10_CUDA_KERNEL_LAUNCH_CHECK()`
- Only handles 1D contiguous tensors

`softplus.cu` requires:

- One `__hostdev__` scalar function that works on all devices
- One impl function for all devices, all scalar types, both
  contiguities, any rank
- No manual CUDA kernel, no device guard code, no grid/block math

The impl function for softplus looks like this:

```cpp
template <typename Tag>
    requires with_type<Tag, torch::DeviceType>
          && with_type<Tag, torch::ScalarType>
          && with_type<Tag, contiguity>
void softplus_impl(Tag tg, torch::Tensor const &input,
                   torch::Tensor &output,
                   double beta_d, double threshold_d) {
    constexpr auto dev    = tag_get<torch::DeviceType>(tg);
    constexpr auto stype  = tag_get<torch::ScalarType>(tg);
    constexpr auto contig = tag_get<contiguity>(tg);

    using scalar_t  = torch_scalar_cpp_type_t<stype>;
    using compute_t = torch_compute_type_t<stype>;

    auto guard = make_device_guard(tg, input);

    auto in  = flat_in<dev, stype, contig>(input);
    auto out = flat_out<dev, stype, contig>(output);

    auto const beta      = static_cast<scalar_t>(beta_d);
    auto const threshold = static_cast<scalar_t>(threshold_d);

    for_each(tg, input.numel(),
        [=] __hostdev__ (Tag, int64_t idx) {
        out[idx] = softplus_scalar<scalar_t, compute_t>(
            in[idx], beta, threshold);
    });
}
```

The tag carries three compile-time coordinates: `DeviceType`,
`ScalarType`, and `contiguity`. `for_each` reads the device from the
tag and does the right thing: on CUDA it launches a grid-stride kernel
with configurable ILP grain; on CPU it uses the thread pool. The views
(`flat_in`/`flat_out`) handle arbitrary rank and strided tensors.
Contiguity is resolved at dispatch time via the dispatch table, not in
the op code. The contiguous specialization is just `data[idx]`; the
strided specialization does unravel + stride math. The op author writes
zero device-specific, contiguity-specific, or rank-specific code.

The dispatch table wiring is also shorter:

```cpp
struct softplus_op {
    template <typename Tag>
    static void op(Tag tg,
                   torch::Tensor const &input,
                   torch::Tensor &output,
                   double beta, double threshold) {
        softplus_impl(tg, input, output, beta, threshold);
    }

    using space = axes<torch_cpu_cuda_device_axis,
                       torch_full_float_stype_axis,
                       full_contiguity_axis>;
    using subspaces = coverage<space>;
    using dispatcher = dispatch_table<space,
        void(torch::Tensor const &,
             torch::Tensor &, double, double)>;
};
```

Two view families are provided. `flat_in`/`flat_out` give rank-free
`operator[]` access for elementwise patterns. `tensor_in`/`tensor_out`
give multi-index `operator()` access with a compile-time rank parameter,
for structured patterns like gather-scatter or morton encoding. Both
families specialize on contiguity.

## CPU parallelism

`for_each` on CPU uses a thread pool rather than `at::parallel_for` or
raw OpenMP. The pool is a singleton that persists for the process
lifetime, avoiding thread creation/teardown costs on every op call.

Two scheduling strategies are provided. `broadcast_pool` uses static
partitioning where each thread gets a contiguous chunk.
`work_stealing_pool` uses Chase-Lev deques and randomized victim
selection, which handles imbalanced workloads where static partitioning
would leave threads idle. The default is work-stealing.

Both pools support nested `parallel_for` safely: a nested call runs
serially on the calling thread, avoiding deadlock and oversubscription.
The caller thread participates as a worker (N+1 threads for N pool
threads). Workers use a brief spin-wait before sleeping to amortize OS
wake/sleep costs.

The benchmarks (described below) drove the choice of work-stealing as
the default. For balanced elementwise workloads, both strategies perform
similarly. For unbalanced voxel workloads, work-stealing adapts.

## Tag and axis canonicalization

This is the part that required the most reworking of internals, and the
part that matters least for day-to-day op writing. But it matters for
correctness and it matters for where sparse convolution dispatch is
headed.

Tags and axes are now self-normalizing. `tag<A, B>` and `tag<B, A>`
resolve to the same type. `axes<Axis0, Axis1>` and
`axes<Axis1, Axis0>` resolve to the same type. Canonical order is
determined by `type_label<decltype(V)>`, using compile-time
lexicographic comparison of fixed-length strings. Each axis value type
must have a `type_label` specialization (torch types and the dispatch
enums have theirs provided; user types register by specializing the
trait).

The reason this matters is that the concept constraints on an op
function are a conjunction of `with_value` and `with_type` clauses,
and conjunctions are associative. The order you write them in shouldn't
affect which instantiation the compiler picks. By making the underlying
tag type order-independent, we guarantee that the C++20 subsumption
rules work correctly regardless of how the author orders the template
parameters or the requires clauses.

This is directly relevant to sparse convolution dispatch, where the
dispatch matrix is large. A broad catch-all for all CUTLASS CUDA
channel-last combinations would look like:

```cpp
template <typename Tag>
    requires with_value<Tag, torch::kCUDA>
          && with_value<Tag, layout::channel_last>
          && with_value<Tag, algorithm::cutlass>
void sparse_conv_forward(Tag, torch::Tensor input,
                         torch::Tensor weights,
                         torch::Tensor output) {
    // input_stype, weight_stype, output_stype, mma_stype
    // are all still compile-time known via Tag — just
    // unconstrained here. This catches ALL cutlass + CUDA +
    // channel_last combinations.
}
```

And a fully specialized variant pinning every scalar type:

```cpp
template <typename Tag>
    requires with_value<Tag, torch::kCUDA>
          && with_value<Tag, layout::channel_last>
          && with_value<Tag, algorithm::cutlass>
          && with_value<Tag, input_stype(torch::kFloat16)>
          && with_value<Tag, weight_stype(torch::kFloat16)>
          && with_value<Tag, output_stype(torch::kFloat16)>
          && with_value<Tag, mma_stype(torch::kFloat16)>
void sparse_conv_forward(Tag, torch::Tensor input,
                         torch::Tensor weights,
                         torch::Tensor output) {
    // Fully pinned: fp16 everything, CUDA, channel_last,
    // cutlass. Subsumes the 3-constraint version above —
    // compiler picks this one when all 7 match.
}
```

The 7-constraint overload subsumes the 3-constraint one. The compiler
picks the more specific match when it applies, and falls through to the
general version otherwise. This is exactly how MLIR separates the
configuration of an operation (attribute matching) from the invocation,
and it scales to dispatch matrices with dozens of axes without
combinatorial if-constexpr nesting.

For this to work, `tag<kCUDA, channel_last, cutlass, ...>` must be the
same type regardless of the order those values appear in the template
parameter list. That's what the canonicalization achieves.

The `named<Label, T>` wrapper exists for the sparse conv case
specifically. `input_stype`, `weight_stype`, `output_stype`, and
`mma_stype` are all `torch::ScalarType` under the hood. Tags require
unique types, so each one gets a distinct label. The label participates
in the canonical ordering, so `named<"input_stype", ScalarType>` and
`named<"weight_stype", ScalarType>` are distinct types that sort
deterministically.

## Benchmarks

Three dispatch benchmarks are added.

`cpu_pool_comparison` compares the broadcast pool, work-stealing pool,
a simple queue-based pool, a spin-wait pool, OpenMP,
`at::parallel_for`, and `torch::gelu` on a SIMD GELU workload. This is
the benchmark that informed the choice of work-stealing as the default
pool.

`synthetic_voxel_benchmark` tests uniform and unbalanced compute-bound
workloads representative of per-voxel processing.

`for_each_benchmark` measures `for_each` with flat views across
contiguous/strided, CPU/CUDA, and float32/float64, with speed-of-light
baselines (memcpy, `torch::softplus`). All tensors are pre-allocated so
allocation is never measured.

## Test coverage

New test files: `for_each_test.cu` (828 lines),
`with_value_test.cpp` (292 lines),
`label_tag_axis_test.cpp` (289 lines),
`detail_index_math_test.cpp` (453 lines),
`thread_pool_test.cpp` (128 lines). Existing test files updated for
the new API surface.

## File inventory

New files:
- `dispatch/detail/core_types.h` (tags, axes, labels,
  canonicalization)
- `dispatch/with_value.h` (concept-based tag matching)
- `dispatch/torch/for_each.h` (parallel iteration)
- `dispatch/torch/views.h` (flat and ranked tensor views)
- `dispatch/torch/dispatch.h` (device guard, contiguity helpers)
- `dispatch/thread_pool.h` (broadcast and work-stealing pools)
- `dispatch/macros.h` (DISPATCH_UNROLL, DISPATCH_SPIN_PAUSE,
  \_\_hostdev\_\_)
- `examples/softplus.cu`, `examples/softplus.h`

Moved/renamed:
- `dispatch/detail.h` → `dispatch/detail/index_math.h`
- `dispatch/axes_map.h` → `dispatch/detail/axes_map.h`
- `dispatch/visit_spaces.h` → `dispatch/detail/visit_spaces.h`
- `dispatch/torch_types.h` → `dispatch/torch/types.h`

Removed:
- `dispatch/types.h` (replaced by `detail/core_types.h`)
- `dispatch/torch.h` (split into `dispatch/torch/dispatch.h` and
  `dispatch/torch/types.h`)
- `tests/types_test.cpp` (replaced by `label_tag_axis_test.cpp` and
  `with_value_test.cpp`)

Attached are the results of the visualized benchmarks.

benchmark_synthetic_uniform

benchmark_cpu_pool
benchmark_for_each_cpu_float32
benchmark_for_each_cpu_float64
benchmark_for_each_cuda_float32
benchmark_for_each_cuda_float64
benchmark_for_each_device_comparison
benchmark_synthetic_comparison
benchmark_synthetic_unbalanced

### DIFF
--- a/src/dispatch/dispatch/torch/for_each.h
+++ b/src/dispatch/dispatch/torch/for_each.h
@@ -25,6 +25,12 @@
 //   The CUDA and PrivateUse1 overloads extract block_dim from the tag if
 //   present, otherwise use 256.
 //
+// GPU grid size:
+//   Determined by cudaOccupancyMaxActiveBlocksPerMultiprocessor — launches
+//   exactly enough blocks to saturate all SMs on the current device, capped
+//   at the number of blocks the work actually requires. The grid-stride loop
+//   handles any count regardless of grid size.
+//
 // Vectorization:
 //   for_each is a scalar index generator — one element per functor call,
 //   one thread per element (interleaved), optimal coalescing for scalar
@@ -90,6 +96,26 @@ for_each_block_dim() {
     }
 }
 
+// Compute the grid size for a grid-stride kernel: enough blocks to saturate
+// all SMs on the current device, but no more than the work requires.
+template <int BlockDim, typename Kernel>
+int
+for_each_grid_dim(Kernel kernel, int64_t count) {
+    int blocks_per_sm = 0;
+    C10_CUDA_CHECK(cudaOccupancyMaxActiveBlocksPerMultiprocessor(
+        &blocks_per_sm, kernel, BlockDim, /*dynamicSMemSize=*/0));
+
+    int device = 0;
+    C10_CUDA_CHECK(cudaGetDevice(&device));
+
+    int sm_count = 0;
+    C10_CUDA_CHECK(cudaDeviceGetAttribute(&sm_count, cudaDevAttrMultiProcessorCount, device));
+
+    int64_t const max_active_blocks = static_cast<int64_t>(blocks_per_sm) * sm_count;
+    int64_t const blocks_needed     = (count + BlockDim - 1) / BlockDim;
+    return static_cast<int>(std::min(blocks_needed, max_active_blocks));
+}
+
 } // namespace detail
 
 template <int BlockDim, typename Tag, typename Func>
@@ -111,16 +137,14 @@ for_each(Tag tag, int64_t count, Func &&func) {
     if (count == 0)
         return;
 
-    constexpr int kBlockDim   = detail::for_each_block_dim<Tag>();
-    constexpr int kMaxGridDim = 65535;
+    constexpr int kBlockDim = detail::for_each_block_dim<Tag>();
 
-    int64_t const blocks_needed = (count + kBlockDim - 1) / kBlockDim;
+    using F = std::decay_t<Func>;
     int const grid_dim =
-        static_cast<int>(std::min(blocks_needed, static_cast<int64_t>(kMaxGridDim)));
+        detail::for_each_grid_dim<kBlockDim>(for_each_cuda_kernel<kBlockDim, Tag, F>, count);
 
     cudaStream_t stream = c10::cuda::getCurrentCUDAStream().stream();
 
-    using F = std::decay_t<Func>;
     for_each_cuda_kernel<kBlockDim>
         <<<grid_dim, kBlockDim, 0, stream>>>(tag, count, F(std::forward<Func>(func)));
     C10_CUDA_KERNEL_LAUNCH_CHECK();
@@ -149,8 +173,7 @@ for_each(Tag tag, int64_t count, Func &&func) {
     if (count == 0)
         return;
 
-    constexpr int kBlockDim   = detail::for_each_block_dim<Tag>();
-    constexpr int kMaxGridDim = 65535;
+    constexpr int kBlockDim = detail::for_each_block_dim<Tag>();
 
     using F = std::decay_t<Func>;
     F f(std::forward<Func>(func));
@@ -173,9 +196,8 @@ for_each(Tag tag, int64_t count, Func &&func) {
         }
 
         if (chunk_count > 0) {
-            int64_t const blocks_needed = (chunk_count + kBlockDim - 1) / kBlockDim;
-            int const grid_dim =
-                static_cast<int>(std::min(blocks_needed, static_cast<int64_t>(kMaxGridDim)));
+            int const grid_dim = detail::for_each_grid_dim<kBlockDim>(
+                for_each_pvt1_kernel<kBlockDim, Tag, F>, chunk_count);
 
             for_each_pvt1_kernel<kBlockDim>
                 <<<grid_dim, kBlockDim, 0, stream>>>(tag, chunk_count, chunk_offset, f);


src/CMakeLists.txt
```diff
@@ -10,10 +10,12 @@ option(FVDB_BUILD_TESTS "Configure CMake to build tests" ON)
 option(FVDB_BUILD_BENCHMARKS "Configure CMake to build (google & nvbench) benchmarks" OFF)
 option(FVDB_STRIP_SYMBOLS "Strip symbols from the build" OFF)
 option(FVDB_LINEINFO "Enable lineinfo in the build" OFF)
+option(FVDB_USE_OPENMP "Enable OpenMP for CPU parallelization" ON)
```
swahtz, line 13: Per the slack discussion, is OpenMP supposed to be OFF here?

PR #462, opened 2026-02-11
Fix CI checks showing 'waiting for status' when paths-ignore skips workflows

### LINKED ISSUE #456: CI checks show 'waiting for status' instead of passing when paths-ignore skips workflows
## Problem

When a PR only changes files that are excluded by `paths-ignore` in our workflow triggers (e.g., markdown files, scripts, notebooks), the workflows are not run at all. GitHub reports these checks as "waiting for status to be reported" rather than marking them as passed.

This affects any PR that only touches:
- `*.md` files
- `CODEOWNERS`
- `debug/`, `examples/`, `notebooks/`, `scripts/` directories
- `devtools/` (for workflows that ignore it)

For example, PR #445 (which only adds shell scripts and markdown) shows all test checks as perpetually pending.

If branch protection requires these checks to pass, such PRs cannot be merged without manual intervention or workarounds.

### Affected workflows

- `tests.yml` (fVDB Unit Tests)
- `codestyle.yml`
- Potentially `nightly.yml`, `cu128.yml`, `cu130.yml`, etc.

## Proposed solution

Replace `paths-ignore` at the workflow trigger level with a **change-detection job** that always runs. Make the real jobs conditional on whether relevant files changed. When a conditional job is skipped due to an `if` condition evaluating to false, GitHub marks it as "passed" for branch protection purposes.

Use [`dorny/paths-filter`](https://github.com/dorny/paths-filter) with **negation patterns** and `predicate-quantifier: 'every'` to mirror the existing `paths-ignore` list as a blacklist. This avoids maintaining a whitelist of source paths that could get out of sync when new directories are added.

### Sketch

```yaml
on:
  pull_request_target:
    branches:
      - '**'
    # Remove paths-ignore entirely
  workflow_dispatch:
    inputs:
      branch:
        description: "Branch to build"
        required: true
        default: "main"

jobs:
  check-changes:
    name: Check for relevant changes
    runs-on: ubuntu-latest
    outputs:
      should_test: ${{ steps.filter.outputs.src }}
    steps:
      - uses: dorny/paths-filter@v3
        id: filter
        with:
          predicate-quantifier: 'every'
          filters: |
            src:
              - '**'
              - '!**/*.md'
              - '!CODEOWNERS'
              - '!debug/**'
              - '!examples/**'
              - '!notebooks/**'
              - '!scripts/**'

  start-build-runner:
    name: Start CPU-only EC2 runner for build
    needs: check-changes
    if: needs.check-changes.outputs.should_test == 'true'
    # ... rest of job unchanged
```

The filter reads as: "match any changed file, EXCEPT markdown files, CODEOWNERS, debug/, examples/, notebooks/, scripts/." This is a direct mirror of the current `paths-ignore` block in each workflow. New source directories are automatically included without updating the filter — you'd only need to update it when adding a new directory to *ignore*, which is the same maintenance burden as today.

All downstream jobs that `need` the conditional job will automatically be skipped (and marked as passed by GitHub) when `should_test` is false.

### Considerations

- **`pull_request_target` compatibility**: `dorny/paths-filter` works with `pull_request_target` and uses the GitHub REST API to fetch changed files (no checkout needed for PRs), but may need the `base` and `ref` parameters configured explicitly since `pull_request_target` runs in the context of the base branch.
- **`workflow_dispatch`**: The change-detection job should default to `should_test: 'true'` for manual triggers so `workflow_dispatch` always runs tests. This can be done with a fallback expression in the output.
- **Cost**: The `check-changes` job runs on `ubuntu-latest` (free tier) and takes ~10 seconds, so there's negligible cost.
- **Multiple workflows**: The same pattern should be applied to all workflows that currently use `paths-ignore`.

## Alternatives considered

- **Whitelist approach**: Specifying paths to match (e.g., `src/**`, `tests/**`) instead of paths to ignore. This is fragile because new directories would be silently excluded from CI until someone remembers to update the filter.
- **Separate "skip" workflow**: Create a second workflow that triggers on the ignored paths and reports success for the same check names. This is fragile and requires keeping check names in sync.
- **Repository rulesets with path-based bypass**: GitHub doesn't natively support this.

swahtz:
## Summary

Fixes #456

- Replaces `paths-ignore` trigger filters with a `check-changes` gate job in `tests.yml`, `cu128.yml`, and `cu130.yml`
- When a PR only touches ignored paths (markdown, scripts, notebooks, etc.), the `check-changes` job detects no relevant changes and all downstream jobs are skipped -- GitHub marks these as "passed" for branch protection instead of "waiting for status"
- Uses `dorny/paths-filter@v3` with negation patterns and `predicate-quantifier: 'every'` to mirror the existing `paths-ignore` lists exactly
- `workflow_dispatch` triggers default to `should_test: 'true'` so manual runs are unaffected

## Changes

Three workflow files modified with identical pattern:

1. **Removed `paths-ignore`** from `pull_request_target` triggers so workflows always fire
2. **Added `check-changes` job** (runs on `ubuntu-latest`, ~10s) that uses `dorny/paths-filter@v3` to detect relevant file changes
3. **Gated `start-build-runner`** on `needs.check-changes.outputs.should_test == 'true'`, preserving the existing draft-PR guard

No EC2 runners are started when `should_test` is false, so there is zero cost for docs-only PRs.

## Test plan

- [ ] After merging, open a docs-only PR (touching only `.md` files) and confirm all check names appear as "passed" (skipped) rather than "waiting for status"
- [ ] Open a code PR and confirm all tests still run normally
- [ ] Trigger `tests.yml` via `workflow_dispatch` and confirm it runs all tests (the fallback defaults to `true` for non-PR events)

> **Note:** Since these workflows use `pull_request_target`, the changes only take effect after merging to main. The PR itself will still run the old workflow from the base branch.

### DIFF
--- a/.github/workflows/cu128.yml
+++ b/.github/workflows/cu128.yml
@@ -23,28 +23,7 @@ jobs:
   # CHECK FOR RELEVANT CHANGES
   ##############################################################################
   check-changes:
-    name: Check for relevant changes
-    runs-on: ubuntu-latest
-    outputs:
-      should_test: ${{ github.event_name != 'pull_request_target' || steps.filter.outputs.src == 'true' }}
-    steps:
-      # predicate-quantifier is supported in code but missing from action.yml in v3;
-      # produces a non-blocking "Invalid action input" warning. Upstream fix: PR #226.
-      - uses: dorny/paths-filter@v3
-        id: filter
-        if: github.event_name == 'pull_request_target'
-        with:
-          predicate-quantifier: 'every'
-          filters: |
-              src:
-                - '**'
-                - '!**/*.md'
-                - '!CODEOWNERS'
-                - '!debug/**'
-                - '!examples/**'
-                - '!notebooks/**'
-                - '!devtools/**'
-                - '!scripts/**'
+    uses: ./.github/workflows/check-changes.yml
 
   ##############################################################################
   # START FVDB BUILD RUNNER
--- a/.github/workflows/check-changes.yml
+++ b/.github/workflows/check-changes.yml
@@ -1,0 +1,43 @@
+# Copyright Contributors to the OpenVDB Project
+# SPDX-License-Identifier: Apache-2.0
+#
+# Reusable workflow: detect whether a PR touches source files or only
+# documentation / non-code paths. Callers gate their real jobs on the
+# `should_test` output so that docs-only PRs are marked as passed
+# (not perpetually "waiting for status").
+
+name: Check for relevant changes
+
+on:
+  workflow_call:
+    outputs:
+      should_test:
+        description: "'true' when the change includes source files, 'false' for docs-only"
+        value: ${{ jobs.check-changes.outputs.should_test }}
+
+# Permissions are inherited from the calling workflow; do not set them here.
+
+jobs:
+  check-changes:
+    name: Check for relevant changes
+    runs-on: ubuntu-latest
+    outputs:
+      should_test: ${{ github.event_name != 'pull_request_target' || steps.filter.outputs.src == 'true' }}
+    steps:
+      # predicate-quantifier is supported in code but missing from action.yml in v3;
+      # produces a non-blocking "Invalid action input" warning. Upstream fix: PR #226.
+      - uses: dorny/paths-filter@v3
+        id: filter
+        if: github.event_name == 'pull_request_target'
+        with:
+          predicate-quantifier: 'every'
+          filters: |
+            src:
+              - '**'
+              - '!**/*.md'
+              - '!CODEOWNERS'
+              - '!debug/**'
+              - '!examples/**'
+              - '!notebooks/**'
+              - '!devtools/**'
+              - '!scripts/**'
--- a/.github/workflows/cu130.yml
+++ b/.github/workflows/cu130.yml
@@ -23,28 +23,7 @@ jobs:
   # CHECK FOR RELEVANT CHANGES
   ##############################################################################
   check-changes:
-    name: Check for relevant changes
-    runs-on: ubuntu-latest
-    outputs:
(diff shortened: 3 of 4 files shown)


.github/workflows/cu128.yml
```diff
@@ -26,13 +19,42 @@ permissions:
...
+      # predicate-quantifier is supported in code but missing from action.yml in v3;
+      # produces a non-blocking "Invalid action input" warning. Upstream fix: PR #226.
+      - uses: dorny/paths-filter@v3
+        id: filter
+        if: github.event_name == 'pull_request_target'
+        with:
+          predicate-quantifier: 'every'
+          filters: |
+              src:
+                - '**'
```
harrism: **💡 suggestion:** ‏Could you also add `!/devtools/**`, or `!/devtools/worktree`?

.github/workflows/cu128.yml
```diff
@@ -26,13 +19,42 @@ permissions:
   id-token: write
 
 jobs:
+  ##############################################################################
```
harrism, line 22: **❓ question:** ‏Is there a way to factor this job out into a common file so we don't have to edit it in every dependent workflow (3 currently, could grow).

harrism (commented):
A couple of questions first.

.github/workflows/cu128.yml
swahtz: Sure thing.  Done

.github/workflows/cu128.yml
swahtz: Apparently you can!  Thanks for the suggestion, I've made a change to share this job definition between the 3 workflows

harrism: This is awesome!

PR #468, opened 2026-02-12
Fix JaggedTensor single-element constructor unconditionally initializing CUDA via pinned_memory

### LINKED ISSUE #467: JaggedTensor single-element constructor unconditionally initializes CUDA via pinned_memory
## Summary

Constructing a `JaggedTensor` with a single tensor (`tensors.size() == 1`) unconditionally allocates pinned (page-locked) memory for the internal `mOffsets` tensor. Pinned memory allocation calls `cudaHostAlloc` under the hood, which forces CUDA runtime initialization -- even when the tensor is CPU-only. This causes:

1. **Crashes in forked `DataLoader` worker processes**, where re-initializing the CUDA runtime after `fork()` is forbidden (error: `RuntimeError: CUDA error: initialization error`).
2. **Unnecessary CUDA initialization overhead** on CPU-only workloads.

The multi-element path (`tensors.size() > 1`) does *not* use pinned memory, so it works correctly in all contexts. This makes the bug specific to batch size 1.

## Root Cause

There are two locations where `.pinned_memory(true)` is used unconditionally in code paths reachable from CPU tensors:

### Location 1: `src/fvdb/JaggedTensor.cpp` line 82-86

In the `JaggedTensor(const std::vector<torch::Tensor>&)` constructor, single-element branch:

```cpp
mOffsets = torch::tensor({JOffsetsType(0), mData.size(0)},
                         torch::TensorOptions()
                             .dtype(JOffsetsScalarType)
                             .device(mData.device())
                             .pinned_memory(true));  // <-- forces CUDA init
```

### Location 2: `src/fvdb/detail/ops/JOffsetsFromJIdx.cu` line 26-27

In `joffsetsFromJIdx()`, called by the `JaggedTensor(torch::Tensor data)` constructor for the single-tensor case:

```cpp
torch::Tensor ret =
    torch::empty({2}, torch::TensorOptions().dtype(JOffsetsScalarType).pinned_memory(true));
```

This function is shared by CPU, CUDA, and PrivateUse1 dispatches, so the unconditional pinning is harmful when called from the CPU path.

### Locations that are fine (no changes needed)

- `JaggedTensorIndex.cu` line 563: Only executes on CUDA (launches CUDA kernels), so `pinned_memory(true)` is appropriate.
- `Bindings.cpp` lines 37, 64: Pass through a user-specified `pin_memory` parameter (default `false`).

## Reproduction

```python
import torch
from torch.utils.data import Dataset, DataLoader
import fvdb

class SimpleDataset(Dataset):
    def __len__(self):
        return 100
    def __getitem__(self, idx):
        # Single-element JaggedTensor on CPU triggers CUDA init
        return fvdb.JaggedTensor([torch.randn(5)])

ds = SimpleDataset()
# num_workers > 0 uses fork(); CUDA re-init in child process crashes
loader = DataLoader(ds, batch_size=1, num_workers=2)
for batch in loader:  # RuntimeError: CUDA error: initialization error
    pass
```

## Proposed Fix

Make `pinned_memory` conditional on the device being CUDA:

- **Location 1** (`JaggedTensor.cpp`): Change `.pinned_memory(true)` to `.pinned_memory(mData.device().is_cuda())`
- **Location 2** (`JOffsetsFromJIdx.cu`): Change `.pinned_memory(true)` to `.pinned_memory(jdata.device().is_cuda())`

This preserves the performance benefit of pinned memory for CUDA tensors (fast async CPU-to-GPU copies) while avoiding CUDA initialization for CPU-only tensors.

A test should also be added to verify that constructing a single-element CPU `JaggedTensor` does not trigger CUDA initialization.

## Environment

- PyTorch DataLoader with `num_workers > 0` (forked workers)
- CPU-only JaggedTensors with batch size 1

swahtz:
## Summary

- Make `pinned_memory` conditional on the tensor device being CUDA in two locations where single-element `JaggedTensor` construction unconditionally allocated pinned (page-locked) memory via `cudaHostAlloc`, which forced CUDA runtime initialization even for CPU-only tensors.
- This caused crashes in forked `DataLoader` worker processes (where re-initializing CUDA after `fork()` is forbidden) and added unnecessary overhead for CPU-only workloads.
- Add a test verifying that CPU single-element `JaggedTensor` offsets are not pinned.

Fixes #467

### `src/fvdb/JaggedTensor.cpp`
`.pinned_memory(true)` → `.pinned_memory(mData.device().is_cuda())` in the `JaggedTensor(const std::vector<torch::Tensor>&)` single-element branch.

### `src/fvdb/detail/ops/JOffsetsFromJIdx.cu`
`.pinned_memory(true)` → `.pinned_memory(jdata.device().is_cuda())` in `joffsetsFromJIdx()`, which is the shared implementation called by CPU, CUDA, and PrivateUse1 dispatch paths.

### `tests/unit/test_jagged_tensor.py`
New `test_cpu_single_element_no_cuda_init` verifying both constructor paths produce non-pinned offsets for CPU tensors.

## Test plan

- [x] `python -m pytest tests/unit/test_jagged_tensor.py::TestJaggedTensor::test_cpu_single_element_no_cuda_init -v` passes
- [x] `python -m pytest tests/unit/test_jagged_tensor.py::TestJaggedTensor::test_batch_size_one_cpu_float32 -v` passes (existing single-element test)
- [x] `python -m pytest tests/unit/test_jagged_tensor.py -v` full test suite passes
- [x] Reproduction script from issue #467 (CPU JaggedTensor in forked DataLoader) no longer crashes

### DIFF
--- a/src/fvdb/detail/ops/JOffsetsFromJIdx.cu
+++ b/src/fvdb/detail/ops/JOffsetsFromJIdx.cu
@@ -23,13 +23,14 @@ joffsetsFromJIdx(torch::Tensor jidx, torch::Tensor jdata, int64_t numTensors) {
     TORCH_CHECK_VALUE(jidx.dim() == 1, "jidx must be a 1D tensor");
 
     if (jidx.size(0) == 0 && numTensors == 1) {
-        torch::Tensor ret = torch::empty({2},
-                                         torch::TensorOptions()
-                                             .dtype(JOffsetsScalarType)
-                                             .pinned_memory(jdata.device().is_cuda()));
-        auto acc          = ret.accessor<JOffsetsType, 1>();
-        acc[0]            = 0;
-        acc[1]            = jdata.size(0);
+        torch::Tensor ret = torch::empty(
+            {2},
+            torch::TensorOptions()
+                .dtype(JOffsetsScalarType)
+                .pinned_memory(jdata.device().is_cuda() || jdata.device().is_privateuseone()));
+        auto acc = ret.accessor<JOffsetsType, 1>();
+        acc[0]   = 0;
+        acc[1]   = jdata.size(0);
         return ret.to(jdata.device());
     }
 
--- a/src/fvdb/JaggedTensor.cpp
+++ b/src/fvdb/JaggedTensor.cpp
@@ -79,11 +79,12 @@ JaggedTensor::JaggedTensor(const std::vector<torch::Tensor> &tensors) {
                     "assigned data must have shape [N, ...], but got data.dim() = 0");
         mBatchIdx =
             torch::empty({0}, torch::TensorOptions().dtype(JIdxScalarType).device(mData.device()));
-        mOffsets = torch::tensor({JOffsetsType(0), mData.size(0)},
-                                 torch::TensorOptions()
-                                     .dtype(JOffsetsScalarType)
-                                     .device(mData.device())
-                                     .pinned_memory(mData.device().is_cuda()));
+        mOffsets = torch::tensor(
+            {JOffsetsType(0), mData.size(0)},
+            torch::TensorOptions()
+                .dtype(JOffsetsScalarType)
+                .device(mData.device())
+                .pinned_memory(mData.device().is_cuda() || mData.device().is_privateuseone()));
         mListIdx = torch::empty(
             {0, 1}, torch::TensorOptions().dtype(JLIdxScalarType).device(mData.device()));
         mNumOuterLists = 1;


src/fvdb/detail/ops/JOffsetsFromJIdx.cu
```diff
@@ -23,11 +23,13 @@ joffsetsFromJIdx(torch::Tensor jidx, torch::Tensor jdata, int64_t numTensors) {
     TORCH_CHECK_VALUE(jidx.dim() == 1, "jidx must be a 1D tensor");
 
     if (jidx.size(0) == 0 && numTensors == 1) {
-        torch::Tensor ret =
-            torch::empty({2}, torch::TensorOptions().dtype(JOffsetsScalarType).pinned_memory(true));
-        auto acc = ret.accessor<JOffsetsType, 1>();
-        acc[0]   = 0;
-        acc[1]   = jdata.size(0);
+        torch::Tensor ret = torch::empty({2},
```
matthewdcong: This should be `is_cuda() || is_privateuseone()` so that we use pinned memory for mGPU as well. Same below.

swahtz: Thanks for the catch, that's been fixed up

PR #472, opened 2026-02-17
Skip stopping runners that were skipped for CUDA 12.8 and 13.0 tests.

### LINKED ISSUE #470: Stop-runner jobs run and fail on docs-only PRs when unit tests are skipped
# Stop-runner jobs run and fail on docs-only PRs when unit tests are skipped

**Labels:** `bug`, `ci`

## Problem

When a PR only touches markdown and other ignored paths, `.github/workflows/check-changes.yml` sets `should_test` to `'false'`. In `.github/workflows/tests.yml`, `start-build-runner` is correctly skipped, so the rest of the pipeline (build, start-tests-gpu-runner, tests) is skipped. The cleanup jobs **fvdb-build-stop-runner** and **fvdb-tests-stop-gpu-runner** use `if: ${{ always() }}`, so they still run. Their `needs` include the start-runner jobs, which were skipped and produce no outputs. The stop step then runs with empty `label` and `ec2-instance-id`, and fails (e.g. "The 'ec2-instance-id' input is required for the 'stop' mode").

**Example failure:** https://github.com/openvdb/fvdb-core/actions/runs/22082088985/job/63809217082?pr=466

## Root cause

No check that the start-runner job actually ran. `always()` correctly ensures we stop the runner when build/tests fail, but it also runs the stop job when the start job was skipped (docs-only PR), when there is no runner to stop.

## Proposed fix

Only run each stop-runner job when the corresponding start-runner job was not skipped (`needs.<job_id>.result` is `'skipped'` when a job is skipped).

- **fvdb-build-stop-runner:**  
  `if: ${{ always() }}` → `if: ${{ always() && needs.start-build-runner.result != 'skipped' }}`

- **fvdb-tests-stop-gpu-runner:**  
  `if: ${{ always() }}` → `if: ${{ always() && needs.start-tests-gpu-runner.result != 'skipped' }}`

Then: runner started (success or failure) → stop job runs; runner never started (docs-only) → stop job skipped, no failure.

harrism:
Fixes #470.

(For real this time)

### DIFF
--- a/.github/workflows/cu128.yml
+++ b/.github/workflows/cu128.yml
@@ -181,7 +181,9 @@ jobs:
       - start-build-runner # required to get output from the start-build-runner job
       - fvdb-build # required to wait when the main job is done
     runs-on: ubuntu-latest
-    if: ${{ always() }} # required to stop the runner even if the error happened in the previous jobs
+    # required to stop the runner even if the error happened in the previous jobs
+    # but only if the start-tests-gpu-runner job was not skipped
+    if: ${{ always() && needs.start-build-runner.result != 'skipped' }}
     steps:
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v4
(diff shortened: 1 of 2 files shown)


swahtz (approved):
lgtm!

PR #473, opened 2026-02-18
Sparse Conv Default full feature support

blackencino:
# Consolidate sparse convolution to GatherScatterDefault backend

## Summary

Replace all legacy sparse convolution backends with a single, well-tested default: **GatherScatterDefault** (a compacted-CSR grouped GEMM approach). This PR also adds transposed convolution support, lifts the stride=1 restriction from `ConvolutionPlan`, and significantly expands test coverage.

### New default backend: GatherScatterDefault

The previous codebase carried multiple sparse convolution backends accumulated during experimentation:

- **SparseConvolutionKernelMap** -- the original legacy backend with C++ autograd
- **SparseConvolutionHalo** -- halo-exchange backend limited to 3x3x3 kernels
- **GatherScatter** -- dense kernel-map GEMM backend
- **GatherScatterFused** -- fused single-kernel path for small channel counts
- **CutlassGroupedGemm** -- fp16 TensorCore path (sm80+, channels must be multiples of 32)

After benchmarking across all usage patterns, these have been consolidated into a single **GatherScatterDefault** backend. It:

- Builds a compacted CSR topology directly via two-pass atomic counting (no dense intermediate)
- Streams gather, GEMM, and scatter-add per kernel offset, with working buffers sized to `max_pairs_per_k` rather than `total_pairs`, reducing peak GPU memory from O(total_pairs * C) to O(max_pairs_per_k * C)
- Supports forward, backward, transposed forward, and transposed backward
- Works on both CPU and CUDA with fp16, bf16, fp32, fp64
- Guards against int32 voxel index overflow with a `TORCH_CHECK` on total voxels per grid (<= INT32_MAX)
- Uses `__CUDA_ARCH__` guards for half-precision atomics: native `atomicAdd(__half)` on sm_70+, native `atomicAdd(__nv_bfloat16)` on sm_80+, with CAS fallbacks for older architectures

### Transposed convolution

- Added `conv_transpose_grid()` to `GridBatch` and `Grid` (C++ and Python) for computing the output grid topology of transposed convolution
- Added `BuildGridForConvTranspose.cu/h` implementing the transposed output grid construction
- `ConvolutionPlan` now supports transposed convolution via `from_grid_batch_transposed()` and `from_grid_transposed()`, plus `from_plan_transposed()` for creating encoder-decoder pairs
- Stride > 1 and non-uniform strides are fully supported for both forward and transposed directions

### Removed code

**Deleted backends** (13 files, ~2,800 lines):
- `SparseConvolutionHalo.h/cpp/cu` + `SparseConvolutionHaloGrad.cu/h` (autograd + backend)
- `SparseConvolutionKernelMap.h/cpp/cu` (autograd + backend)
- `ConvPragmaMessage.h`
- `pack_info/ConvolutionKernelMap.h/cu`

**Removed public API surface**:
- `GridBatch.sparse_conv_halo()` / `Grid.sparse_conv_halo()` (Python and C++)
- `_fvdb_cpp.build_kernel_map()` and `_fvdb_cpp.sparse_conv_kernel_map()` (pybind)
- `_fvdb_cpp.gs_conv_fused*` family of bindings (pybind)
- `GridBatch::computeConvolutionKernelMap()` and all `gatherScatterConvFused*` static methods (C++)

**Removed Python backend options**:
- `_GatherScatterFusedBackend`, `_GatherScatterOldBackend` from `convolution_plan.py`
- `gather_scatter_fused` and `gather_scatter_old` expert config values

### Dispatch infrastructure

New shared dispatch helpers added under `src/fvdb/detail/dispatch/`:
- `AtomicAdd.cuh` -- device-dispatched atomic add with architecture-guarded CAS fallbacks
- `ForEachActiveVoxel.cuh` -- iterate over active voxels in a grid
- `GridAccessor.h` -- grid accessor factory
- `JaggedView.h` -- jagged tensor view utilities
- `TensorChecks.h` -- tensor precondition macros

### Python bindings

- `GatherScatterDefaultTopology` class with camelCase C++ members (`gatherIndices`, `scatterIndices`, `offsets`, `totalPairs`, etc.) exposed as snake_case Python properties (`gather_indices`, `scatter_indices`, etc.)
- `gs_build_topology` / `gs_conv` / `gs_conv_backward` / `gs_build_transpose_topology` / `gs_conv_transpose` / `gs_conv_transpose_backward` route to GatherScatterDefault

### Documentation

- Doxygen comments on all new public API items: `ConvDirection` enum, `GatherScatterDefaultTopology` struct and all members, topology builders, convolution entry points, `dispatchBuildGridForConvTranspose`, `jagged_in` template, and `GridBatch` static topology methods
- `@file` blocks on new headers (`GatherScatterDefault.h`, `BuildGridForConvTranspose.h`, `JaggedView.h`)
- Int32 voxel limit documented in `GatherScatterDefaultTopology` Doxygen

### Tests

**New test files**:
- `tests/unit/test_conv_transpose_default.py` -- topology, forward, and backward tests for transposed convolution against dense PyTorch `conv_transpose3d`, including strided tests with stride (2,2,2), (1,2,3), and (3,3,3)
- `tests/unit/test_conv_transpose_ground_truth.py` -- validates PyTorch ground truth assumptions
- `src/tests/GatherScatterDefaultConvTest.cu` -- C++ gtest with forward/backward correctness, adjoint identity, kernel-size variations, and error-path coverage

**Updated tests**:
- `tests/unit/test_conv_default.py` -- rewritten to use `ConvolutionPlan` API with `torch.autograd.gradcheck` tests for forward and transposed convolution (stride-1 and strided, CPU + CUDA)
- `tests/unit/test_conv_ground_truth.py` -- expanded dense convolution validation
- `fvdb/utils/tests/convolution_utils.py` -- shared test utilities including `conv_transpose_ground_truth_strided` for strided transposed ground truth computation

**New benchmarks**:
- `src/benchmarks/convolution/gather_scatter_conv_benchmark.cu` -- C++ microbenchmarks for topology build and convolution
- `src/benchmarks/convolution/benchmark_sparse_conv_comparison.py` -- Python comparison benchmark against other sparse convolution frameworks (self-contained dependencies)

### Naming conventions

fVDB-proper code uses camelCase; the dispatch framework uses snake_case (STL-like convention). At the boundary:
- `GatherScatterDefaultTopology` struct members use camelCase (`gatherIndices`, `kernelSize`, etc.)
- Internal helpers in `GatherScatterDefault.cu` use camelCase (`gsDefaultGather`, `mmOutSafe`, etc.)
- Dispatch op structs (`twopass_topology_op`, `gs_default_conv_op`, etc.) use snake_case per the dispatch framework convention
- `conv_transpose_grid` follows the pre-existing `conv_grid` snake_case convention on `GridBatch`

## Breaking changes

- `GridBatch.sparse_conv_halo()` and `Grid.sparse_conv_halo()` are removed. Use `ConvolutionPlan` instead.
- `_fvdb_cpp.build_kernel_map()` and `_fvdb_cpp.sparse_conv_kernel_map()` are removed.
- `_fvdb_cpp.GatherScatterTopology` is replaced by `_fvdb_cpp.GatherScatterDefaultTopology`.
- The `gather_scatter_fused` and `gather_scatter_old` backend expert config values are removed.

Users of the public `ConvolutionPlan` API are unaffected -- the default backend is now faster and supports transposed convolution and arbitrary strides.

## Test plan

- [x] C++ gtests pass: `./build.sh ctest`
- [x] Python convolution tests pass: `pytest tests/unit/test_conv_default.py tests/unit/test_conv_transpose_default.py -v`
- [x] Ground truth validation tests pass: `pytest tests/unit/test_conv_ground_truth.py tests/unit/test_conv_transpose_ground_truth.py -v`
- [x] Autograd gradcheck tests pass (forward + transposed, stride-1 + strided, CPU + CUDA)
- [x] Full test suite: `pytest tests/ -v`

### DIFF
--- a/tests/unit/test_conv_transpose_default.py
+++ b/tests/unit/test_conv_transpose_default.py
@@ -620,9 +620,7 @@ class TestConvTransposeValues(DisableTF32Mixin, unittest.TestCase):
         assert dense_input_grad is not None, "Dense input grad is None"
 
         try:
-            torch.testing.assert_close(
-                sparse_input_grad, dense_input_grad, rtol=input_grad_rtol, atol=input_grad_atol
-            )
+            torch.testing.assert_close(sparse_input_grad, dense_input_grad, rtol=input_grad_rtol, atol=input_grad_atol)
         except AssertionError:
             diag = diagnose_tensor_mismatch(
                 f"Transpose input gradient (stride={stride}, dtype={dtype})",


blackencino:
conv_benchmark_sparsity_large

src/fvdb/detail/ops/convolution/GatherScatterDefault.h
```diff
@@ -0,0 +1,132 @@
...
+namespace detail {
+namespace ops {
+
+// =============================================================================
+// ConvDirection -- forward vs transposed convolution
+// =============================================================================
+
+enum class ConvDirection { Forward, Transposed };
+
+// =============================================================================
```
harrism: **☑️ todo:** ‏Should this be a doxygen comment block?

src/fvdb/detail/ops/convolution/GatherScatterDefault.h
```diff
@@ -0,0 +1,132 @@
...
+    nanovdb::Coord stride;
+
+    ConvDirection direction;
+};
+
+// =============================================================================
+// Topology builders
+// =============================================================================
+
+/// Build a compacted forward topology via two-pass atomic counting.
```
harrism: **☑️ todo:** ‏Documentation in this file is inconsistent. Some functions have just a brief, some have @param, some have @return.

src/fvdb/detail/ops/convolution/GatherScatterDefault.cu
```diff
@@ -0,0 +1,737 @@
...
+    using dispatcher = dispatch_table<space,
+                                      GatherScatterDefaultTopology(GridBatchImpl const &,
+                                                                   GridBatchImpl const &,
+                                                                   nanovdb::Coord,
+                                                                   nanovdb::Coord,
+                                                                   ConvDirection)>;
+};
+
+static GatherScatterDefaultTopology
+buildTopologyTwoPass(GridBatchImpl const &feature_grid,
```
harrism, line 258: **🚩 issue:** ‏Naming seems all over the place. Some functions use snake_case, others camelCase.  We should stick to one.

src/fvdb/detail/ops/convolution/GatherScatterDefault.cu
```diff
@@ -0,0 +1,737 @@
...
+        return;
+
+    constexpr auto stype = tag_get<torch::ScalarType>(tg);
+    using scalar_t       = torch_scalar_cpp_type_t<stype>;
+
+    auto const *src_ptr = src.data_ptr<scalar_t>();
+    auto *dst_ptr       = dst.data_ptr<scalar_t>();
+    auto const *idx_ptr = indices.data_ptr<int32_t>();
+
+    for_each(tg, total_pairs * C, [=] __hostdev__(Tag, int64_t idx) {
```
harrism, line 318: **❓ question:** ‏How does this compare to, e.g. `thrust::gather` in performance?

fvdb/convolution_plan.py
```diff
@@ -35,6 +35,35 @@ def _channel_pair_supported(in_channels: int, out_channels: int, channel_pairs:
     return (in_channels, out_channels) in channel_pairs
 
 
+# ============================================================
```
harrism, line 38: The autograd function stores `ctx.topo = topo` which holds GPU tensors (`gather_indices`, `scatter_indices`). These stay alive until backward completes. For deep networks with many conv layers, this increases peak GPU memory. This was true of the old approach too (it saved `neighbor_map`), but the new topology is potentially larger (separate gather + scatter arrays vs. a single neighbor_map). Have you compared memory usage before and after? Do we document memory usage anywhere? (Disclosure: LLM review turned this up)

src/fvdb/detail/ops/convolution/GatherScatterDefault.cu
```diff
@@ -0,0 +1,737 @@
...
+                            continue;
+                        probe = nanovdb::Coord(r0 / stride[0], r1 / stride[1], r2 / stride[2]);
+                    } else {
+                        probe = nanovdb::Coord(ijk[0] * stride[0] + k0,
+                                               ijk[1] * stride[1] + k1,
+                                               ijk[2] * stride[2] + k2);
+                    }
+
+                    if (feat_tree_acc.isActive(probe)) {
+                        dispatch::atomic_fetch_add_i32(tg_inner, &counts_ptr[k], 1);
```
harrism, line 152: **💡 suggestion:** ‏ ‏If a single kernel offset has >2B pairs, the int32 counter overflows. Unlikely in practice but worth noting for very large grids? (e.g. in docs)

src/fvdb/detail/ops/convolution/GatherScatterDefault.cu
```diff
@@ -0,0 +1,737 @@
...
+        auto off_acc = topo.offsets.accessor<int64_t, 1>();
+
+        auto feat_buf = torch::empty({TP, C_in}, features.options());
+        auto grad_buf = torch::empty({TP, C_out}, features.options());
+        gs_default_gather(tg, features, feat_buf, topo.gather_indices, TP, C_in);
+        gs_default_gather(tg, grad_output, grad_buf, topo.scatter_indices, TP, C_out);
+
+        auto grad_feat_buf = torch::empty({TP, C_in}, features.options());
+
+        for (int64_t k = 0; k < K; ++k) {
```
harrism, line 635: **❓ question:** ‏The loop structure in the transpose backward function differs from the non-transpose backward. Here we have a single combined loop, there there are two separate loops. Seems fine but just wondering why.

src/fvdb/detail/dispatch/AtomicAdd.cuh
```diff
@@ -0,0 +1,130 @@
...
+template <typename Tag, typename T>
+    requires ::dispatch::with_type<Tag, torch::DeviceType>
+__hostdev__ void
+atomic_add(Tag /*tg*/, T *dst, T src) {
+#if defined(__CUDA_ARCH__)
+    // Device compilation: CUDA atomicAdd.
+    // c10::Half and c10::BFloat16 are layout-compatible with __half and
+    // __nv_bfloat16 but are distinct types, so we reinterpret_cast.
+    if constexpr (std::is_same_v<T, c10::Half>) {
+        atomicAdd(reinterpret_cast<__half *>(dst), static_cast<__half>(src));
```
harrism, line 125: **🚩 issue:** `atomicAdd(__half)` sm_70+, and `atomicAdd(__nv_bfloat16)` requires sm_80+. If we enforce a minimum SM version of 80 in our builds then this is OK. Otherwise we need some guards here, I think.

src/tests/GatherScatterDefaultConvTest.cu
```diff
@@ -0,0 +1,1607 @@
+// Copyright Contributors to the OpenVDB Project
+// SPDX-License-Identifier: Apache-2.0
+//
+// GatherScatterDefaultConvTest.cu -- Tests for the default gather-scatter sparse convolution op.
```
harrism, line 4: **👏 praise:** ‏Test coverage here looks awesome!

tests/unit/test_conv_default.py
```diff
@@ -2,10 +2,16 @@
 # SPDX-License-Identifier: Apache-2.0
```
harrism, line 2: **💡 suggestion:** ‏While you have adjoint identity tests in the C++ gtests, there is no round-trip test in Python here. Since there are more layers in python (e.g. autograd wrapper) maybe we need coverage in Python as well to catch potential bugs that wouldn't affect the C++ tests.

tests/unit/test_conv_transpose_default.py
```diff
@@ -0,0 +1,709 @@
+# Copyright Contributors to the OpenVDB Project
+# SPDX-License-Identifier: Apache-2.0
+#
+"""
```
harrism, line 4: **❓ question:** ‏Do these test cover strided transpose or just stride=1? LLM review suggested the latter. Might want to add strided coverage here.

harrism (commented):
This is an impressive PR. Great work!  I used Claude to write me a guide to help prioritize reviewing. It also pointed out several concerns, although nothing really major, and I didn't have any major concerns either.

Two really small concerns not in inline comments:
1. Do we need to deprecate the sparse_conv_halo before removing?
2. For 3x3x3 kernels, the forward pass makes 27 sequential cuBLAS calls. The benchmark shows this is competitive at realistic sparsity, but users should know about the `dense` backend option for high-occupancy cases. So perhaps document this bottleneck and alternative.

src/fvdb/detail/ops/convolution/GatherScatterDefault.h
blackencino: All newly created or modified C++ code has been given doxygen comment blocks now.

src/fvdb/detail/ops/convolution/GatherScatterDefault.h
blackencino: All newly created C++ code has been given doxygen comment blocks.

blackencino: As well as all modified existing C++ code.

src/fvdb/detail/ops/convolution/GatherScatterDefault.cu
blackencino: I didn't test against thrust. Thrust's host parallelism requires either omp or tbb, and an explicit selection of the parallelism mode with the thrust tag. Since my previous merge of the dispatch framework added a for_each wrapper that got CPU parallelism through the work stealing and broadcast pools we added, we no longer have a dependency on OMP or TBB, and have a workaround.

fvdb/convolution_plan.py
blackencino: The autograd-retained topology is the same size as the old `neighbor_map`: both store `total_pairs * 8 bytes` on GPU. The old approach interleaved two int32 columns in a `[total_pairs, 2]` tensor; the new approach stores two separate `[total_pairs]` int32 arrays (`gather_indices` and `scatter_indices`). The `offsets` array is `[K+1]` int64 stored on host, so it adds negligible GPU memory.

The new approach actually improves peak GPU memory in two ways:

1. **Topology construction**: The old `build_kernel_map` first allocated a dense `[V_out, K]` int32 temporary (e.g. 27x the output voxel count for a 3x3x3 kernel), then compacted it via `torch::nonzero`. The new two-pass builder constructs the CSR directly with only a `[K]` int32 counter array as temporary.

2. **Working buffers during compute**: The per-offset streaming pattern sizes GEMM buffers to `[max_pairs_per_k, C]` rather than `[total_pairs, C]`, reducing peak compute memory from O(total_pairs * C) to O(max_pairs_per_k * C).

src/fvdb/detail/ops/convolution/GatherScatterDefault.cu
blackencino: Added a `TORCH_CHECK` in `checkTopologyPreconditions` that validates both grids have <= INT32_MAX total voxels, which prevents both the per-offset counter overflow and the voxel index overflow in the gather/scatter arrays. Also documented the int32 constraint in the `GatherScatterDefaultTopology` struct's Doxygen. Using int32 atomics is intentional -- int64 atomics are emulated on most GPU architectures and would be a significant performance regression for the topology build phase.

Fixed this, but we definitely have the constraint that we aren't allowed more than int32max active voxels per grid.

src/fvdb/detail/ops/convolution/GatherScatterDefault.cu
blackencino: Resolved -- the GatherScatterDefault refactoring unified both backward functions to use the same single-loop-per-offset structure. Both now do gather + gather + mm + scatter + mm within a single loop over kernel offsets, with buffers sized to `max_pairs_per_k`. The old two-loop structure in the forward backward was a side effect of the original bulk-gather-then-loop approach.

src/tests/GatherScatterDefaultConvTest.cu
blackencino: Thank you!!! There's so many places where this stuff can go wrong, building up the tests has actually been the main way it has gotten untangled. It has created excellent guardrails for the coding assistants.

src/fvdb/detail/dispatch/AtomicAdd.cuh
blackencino: Good catch. Added `__CUDA_ARCH__` guards: native `atomicAdd(__half *)` on sm_70+, native `atomicAdd(__nv_bfloat16 *)` on sm_80+, with a 32-bit CAS fallback for older architectures. This matches the pattern used by the existing `Atomics.cuh` in the codebase and ensures compilation succeeds for the nightly CI targets (sm_70, sm_75).

tests/unit/test_conv_default.py
blackencino: Added `torch.autograd.gradcheck` tests that verify the Python autograd wrapper (`_GatherScatterConvFn`) by comparing analytical gradients against numerical finite differences. Four test cases cover forward and transposed convolution with stride-1 and strided (2,2,2), each on both CPU and CUDA. This catches potential bugs in the Python autograd layers (ctx.topo caching, save_for_backward, JaggedTensor wrapping) that the C++ adjoint identity tests cannot reach.

tests/unit/test_conv_transpose_default.py
blackencino: Added strided transposed convolution test coverage. The file previously only tested stride=1. New tests cover topology verification, forward value comparison against dense `conv_transpose3d`, and backward gradient comparison (both input and kernel) for stride (2,2,2), (1,2,3), and (3,3,3). A new `conv_transpose_ground_truth_strided` utility in `convolution_utils.py` provides the dense ground truth.

blackencino: And they revealed quite a few errors! Thanks for pointing this out, working on fixes now.

blackencino: Fixed.

src/fvdb/detail/ops/convolution/GatherScatterDefault.cu
blackencino: I've reformatted all the appropriate places to camelCase.

blackencino:
@harrism - I addressed all of your comments. They uncovered one bug in the transpose strided case, so thank you for your care and attention.

sifakis (approved):
Looks nice and clean on my end. (Focused my review on the Gather/Scatter default backend). Awesome work @blackencino

PR #484, opened 2026-02-25
Fix multibatch mGPU race condition in SH backwards op

matthewdcong:
Follow up to #434 . In the mGPU case, different batches can be processed on different GPUs so we need `atomicAdd_system` instead of `gpuAtomicAdd` aka `atomicAdd`.


swahtz (approved):
Looks good, thanks for catching that one!

PR #485, opened 2026-02-26
Refactor Gaussian rendering to use composable camera model for projection and ray generation

fwilliams:
## Summary

This PR continues the gsplat rendering refactor by encapsulating camera-space transform/projection details inside camera operation classes and validating that encapsulated behavior is numerically identical to the previous explicit transform path.

Concretely, projection kernels no longer manually pull `R/t` and perform world-to-camera transform and covariance transform at each callsite. Instead, camera op classes expose world-space projection and world-space VJP entry points, reducing kernel coupling to camera internals while preserving behavior.

## Why this change

This is part of an ongoing series to make the renderer more composable through templated components:

- **Camera type** (`PerspectiveCameraOp`, `OrthographicCameraOp`, distortion/rolling-shutter variants)
- **Tile intersection type** (dense/sparse accessors in follow-up PRs)
- **Shading/rasterization specialization** (existing templated kernels and future shading variants)

The design goal is to keep kernels generic over capabilities while moving model-specific math behind type-specific APIs.

### 1) Camera op API: world-space projection and VJP encapsulation

In `GaussianCameras.cuh`, for perspective and orthographic projection camera ops:

- Added forward helper:
  - `projectWorldGaussianTo2D(...) -> (cov2d, mean2d, depthCam)`
- Added backward helper:
  - `projectWorldGaussianTo2DVJP(...) -> (dCovarWorld, dMeanWorld, dRotation, dTranslation)`

These methods internalize:
- world-to-camera point transform
- world-to-camera covariance transform
- camera projection + projection VJP
- transform Jacobian chaining back to world/camera parameters

### 2) Projection kernels switched to encapsulated camera APIs

Updated kernels:
- `GaussianProjectionForward.cu`
- `GaussianProjectionJaggedForward.cu`
- `GaussianProjectionBackward.cu`
- `GaussianProjectionJaggedBackward.cu`

Refactor effect:
- removed explicit `worldToCamRt` + transform calls from callsites
- replaced with camera-op world-space APIs
- retained output/atomic accumulation behavior and tensor contracts

### 3) Camera-op parity tests (including gradient numerics)

Added `GaussianCamerasOpTest.cu` to directly validate encapsulated methods against the prior explicit reference path on CUDA:

- Forward parity:
  - compares `(cov2d, mean2d, depthCam)` from encapsulated API vs explicit transform+project path
- Backward parity:
  - compares VJP outputs (`dCovarWorld`, `dMeanWorld`, `dRotation`, `dTranslation`) from encapsulated API vs explicit chained reference path

This gives focused unit-level confidence that encapsulation is behavior-preserving for both value and gradient paths.

### 4) Docs and formatting polish

- Clarified and standardized camera-op docs in `GaussianCameras.cuh`.
- Explicitly documented that footprint visibility filtering uses an **AABB overlap check** around the projected ellipse (not exact ellipse clipping).
- Applied `clang-format` to touched C++/CUDA files.

### 5) Shared-memory alignment hardening

Addressed Copilot feedback about implicit alignment assumptions in from-world rasterization:

- Added explicit alignment helpers (`alignUpBytes`, `alignUpAddress`) used for dynamic shared-memory layout.
- Updated distortion camera-op shared-memory packing in `GaussianCameras.cuh` to align each typed sub-region (`Mat3`, `Vec3`, distortion coeff arrays).
- Updated from-world forward/backward kernels to:
  - align the base camera shared-memory pointer before `loadSharedMemory(...)`
  - align `idBatch` and `SharedGaussian` batch regions explicitly
  - budget dynamic shared-memory bytes with alignment padding

These changes are layout-safety hardening only (no intended math or output behavior change).

## Relation to the refactor series

This PR is a foundational camera-layer cleanup in a broader composability effort:

1. **This PR (camera abstraction):** hide camera math internals behind camera op methods.
2. **Next PR(s) (tile intersections):** unify dense/sparse tile traversal/accessor APIs.
3. **Later PR(s) (rasterization/shading composition):** further reduce shared-args surface and compose kernels over camera/intersection/shading types with fewer mode-specific branches.

The end-state is a cleaner type-driven renderer where implementation details are localized by component type rather than spread across kernels.

## Test plan

Run in the `fvdb` conda env:

- `./build.sh install gtests`
- `./build.sh ctest`

Expected result:
- all tests pass, including `GaussianProjection*` and `GaussianCamerasOpTest`.


fvdb/__init__.py
```diff
@@ -190,7 +190,7 @@ def gaussian_render_jagged(
     "JaggedTensor",
     "GaussianSplat3d",
     "ProjectedGaussianSplats",
-    "CameraModel",
+    "DistortionModel",
     "RollingShutterType",
     "ProjectionType",
```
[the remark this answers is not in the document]
fwilliams: This change was introduced pre-release and is not yet used externally. Thus, this rename is fine.

src/fvdb/detail/ops/gsplat/GaussianProjectionJaggedBackward.cu
```diff
@@ -95,7 +89,7 @@ jaggedProjectionBackwardKernel(
     }
 
     // transform Gaussian to camera space
-    const auto [R, t] = loadWorldToCamRtRowMajor4x4(worldToCamMatrices);
+    const auto [R, t] = cameraOp.worldToCamRt(cId);
```
fwilliams: why are we still pulling out R and t here. This should be private state hidden inside the op which only exposes projection.

src/fvdb/detail/ops/gsplat/GaussianProjectionJaggedBackward.h
```diff
@@ -56,6 +58,7 @@ dispatchGaussianProjectionJaggedBackward(const torch::Tensor &gSizes, // [B] gau
                                          const torch::Tensor &cSizes, // [B] camera sizes
                                          const torch::Tensor &worldToCamMatrices, // [C, 4, 4]
                                          const torch::Tensor &projectionMatrices, // [C, 3, 3]
+                                         const bool ortho,
```
fwilliams: same comment about argument re-ordering

src/fvdb/detail/ops/gsplat/GaussianProjectionJaggedForward.h
```diff
@@ -54,13 +56,13 @@ dispatchGaussianProjectionJaggedForward(const torch::Tensor &gSizes, // [B] gaus
                                         const torch::Tensor &cSizes, // [B] camera sizes
                                         const torch::Tensor &worldToCamMatrices, // [C, 4, 4]
                                         const torch::Tensor &projectionMatrices, // [C, 3, 3]
+                                        const bool ortho,
```
fwilliams: same comment about argument re-ordering

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
```diff
@@ -281,6 +253,7 @@ dispatchGaussianProjectionBackward<torch::kCUDA>(
     const torch::Tensor &logScales,                   // [N, 3]
     const torch::Tensor &worldToCamMatrices,          // [C, 4, 4]
     const torch::Tensor &projectionMatrices,          // [C, 3, 3]
+    const bool ortho,
```
fwilliams: same comment about argument re-ordering

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
```diff
@@ -461,6 +443,7 @@ dispatchGaussianProjectionBackward<torch::kPrivateUse1>(
...
```
fwilliams: same comment about argument re-ordering

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
```diff
@@ -650,6 +643,7 @@ dispatchGaussianProjectionBackward<torch::kCPU>(
...
```
fwilliams: same comment about argument re-ordering

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.h
```diff
@@ -60,6 +62,7 @@ dispatchGaussianProjectionBackward(
     const torch::Tensor &scales,                      // [N, 3]
     const torch::Tensor &worldToCamMatrices,          // [C, 4, 4]
     const torch::Tensor &projectionMatrices,          // [C, 3, 3]
+    const bool ortho,
```
fwilliams: same comment about argument re-ordering

src/fvdb/detail/ops/gsplat/GaussianProjectionForward.h
```diff
@@ -54,14 +56,14 @@ dispatchGaussianProjectionForward(const torch::Tensor &means,              // [N
                                   const torch::Tensor &scales,             // [N, 3]
                                   const torch::Tensor &worldToCamMatrices, // [C, 4, 4]
                                   const torch::Tensor &projectionMatrices, // [C, 3, 3]
+                                  const bool ortho,
```
fwilliams: same comment about argument re-ordering

src/fvdb/detail/ops/gsplat/GaussianProjectionJaggedBackward.cu
```diff
@@ -233,6 +205,7 @@ dispatchGaussianProjectionJaggedBackward<torch::kCUDA>(
     const torch::Tensor &cSizes,             // [B] camera sizes
     const torch::Tensor &worldToCamMatrices, // [C, 4, 4]
     const torch::Tensor &projectionMatrices, // [C, 3, 3]
+    const bool ortho,
```
fwilliams: I think no need to re-order arguments here or in other functions, it makes the diff noisier. Also applies to many other places.

src/fvdb/detail/autograd/GaussianRasterizeFromWorld.cpp
```diff
@@ -25,7 +25,7 @@ RasterizeGaussiansToPixelsFromWorld3DGS::forward(
     const RasterizeGaussiansToPixelsFromWorld3DGS::Variable &projectionMatrices,
     const RasterizeGaussiansToPixelsFromWorld3DGS::Variable &distortionCoeffs,
     const fvdb::detail::ops::RollingShutterType rollingShutterType,
-    const fvdb::detail::ops::CameraModel cameraModel,
+    const fvdb::detail::ops::DistortionModel cameraModel,
```
fwilliams: rename variable to distortionModel to be consistent

src/fvdb/detail/autograd/GaussianRasterizeFromWorld.h
```diff
@@ -31,7 +31,7 @@ struct RasterizeGaussiansToPixelsFromWorld3DGS
             const Variable &projectionMatrices,      // [C,3,3]
             const Variable &distortionCoeffs,        // [C,K]
             const fvdb::detail::ops::RollingShutterType rollingShutterType,
-            const fvdb::detail::ops::CameraModel cameraModel,
+            const fvdb::detail::ops::DistortionModel cameraModel,
```
fwilliams: rename variable to distortionModel to be clear

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
```diff
@@ -0,0 +1,1162 @@
...
+isOutsideImageWithRadius(const nanovdb::math::Vec2<T> &mean2d,
+                         const T radiusX,
+                         const T radiusY,
+                         const int32_t imageWidth,
+                         const int32_t imageHeight) {
+    return (mean2d[0] + radiusX <= T(0) || mean2d[0] - radiusX >= T(imageWidth) ||
+            mean2d[1] + radiusY <= T(0) || mean2d[1] - radiusY >= T(imageHeight));
+}
+
+enum class ProjectWorldToPixelStatus : uint8_t { BehindCamera, OutOfBounds, InImage };
```
fwilliams: Does this enum need to be public? Is it used outside of the op classes?

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
```diff
@@ -0,0 +1,1162 @@
...
+inline __device__ void
+loadQuatScaleFromScalesRowMajor(const T *quats4,
+                                const T *scales3,
+                                nanovdb::math::Vec4<T> &outQuatWxyz,
+                                nanovdb::math::Vec3<T> &outScale) {
+    outQuatWxyz = nanovdb::math::Vec4<T>(quats4[0], quats4[1], quats4[2], quats4[3]);
+    outScale    = nanovdb::math::Vec3<T>(scales3[0], scales3[1], scales3[2]);
+}
+
+template <typename T> struct PerspectiveCameraOp {
```
fwilliams: class and methods need docs

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
```diff
@@ -0,0 +1,1162 @@
...
+        } else {
+            dLossDMean3d[2] += -fy * rz3 * dJ[1][2] * ty;
+        }
+        dLossDMean3d[2] += -fx * rz2 * dJ[0][0] - fy * rz2 * dJ[1][1] +
+                           T(2) * fx * tx * rz3 * dJ[0][2] + T(2) * fy * ty * rz3 * dJ[1][2];
+        return {dLossDCovar3d, dLossDMean3d};
+    }
+};
+
+template <typename T> struct OrthographicCameraOp {
```
fwilliams: Class and method need docs

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
```diff
@@ -0,0 +1,1162 @@
...
+                       fy,
+                       T(0)  // 2nd row
+        );
+        const Mat3 dLossDCovar3d(J.transpose() * dLossDCovar2d * J);
+        const Vec3 dLossDMean3d(fx * dLossDMeans2d[0], fy * dLossDMeans2d[1], T(0));
+        return {dLossDCovar3d, dLossDMean3d};
+    }
+};
+
+template <typename T> struct PerspectiveWithDistortionCameraOp {
```
fwilliams: class and methods need docs

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
```diff
@@ -0,0 +1,1162 @@
...
+        copyWorldToCamTranslation<T>(numCameras, worldToCamEndTransShared, worldToCamEndAcc);
+        copyMat3Accessor<T>(numCameras, projectionMatsShared, projectionMatricesAcc);
+        if (numDistCoeffs > 0) {
+            copyDistortionCoeffs<T>(
+                numCameras, numDistCoeffs, distortionShared, distortionCoeffsAcc);
+        }
+    }
+
+    inline __device__ T
+    cameraDepthAtTime(const int64_t cid,
```
fwilliams, line 754: this method and some of the others are part of the public API but it's not clear they need to be or whether we could reduce the API surface area of these ops?

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
```diff
@@ -0,0 +1,1162 @@
...
+        const nanovdb::math::Vec3<T> d_cam =
+            normalizeSafe(nanovdb::math::Vec3<T>(p[0], p[1], T(1)));
+        const nanovdb::math::Vec3<T> origin_w =
+            R_cw * (nanovdb::math::Vec3<T>(T(0), T(0), T(0)) - t_wc);
+        const nanovdb::math::Vec3<T> dir_w = normalizeSafe(R_cw * d_cam);
+        return nanovdb::math::Ray<T>(origin_w, dir_w);
+    }
+};
+
+template <typename T> struct OrthographicWithDistortionCameraOp {
```
fwilliams: class and methods need docs

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
```diff
@@ -141,7 +135,7 @@ projectionBackwardKernel(const int32_t offset,
     }
 
     // transform Gaussian to camera space
-    const auto [R, t] = loadWorldToCamRtRowMajor4x4(worldToCamMatrices);
+    const auto [R, t] = cameraOp.worldToCamRt(cId);
```
fwilliams: do we need R, t here? they get used to transform the covariance into camera space and apply the VJP, but this feels like it could all be in the cameraOp class for cleaner surface

src/fvdb/detail/ops/gsplat/GaussianRasterize.cuh
```diff
@@ -71,10 +71,13 @@ template <typename ScalarType, size_t NUM_CHANNELS, bool IS_PACKED> struct Raste
...
     uint32_t mNumGaussiansPerCamera;
     uint32_t mTotalIntersections;
-    uint32_t mImageWidth;
-    uint32_t mImageHeight;
-    uint32_t mImageOriginW;
-    uint32_t mImageOriginH;
+    // Raster window (crop) dimensions/origin. These are render-target coordinates, not full camera
+    // sensor dimensions. Packed render-window indexing is:
+    // [0] = renderWidth, [1] = renderHeight, [2] = renderOriginX, [3] = renderOriginY.
+    uint32_t mRenderWidth;
```
fwilliams: should we make this state private since we have accessors? Also, do we need numCameras stored here? In general all the mXXX should be private

src/fvdb/GaussianSplat3d.h
```diff
@@ -915,26 +915,28 @@ class GaussianSplat3d {
...
+    renderImagesFromWorld(const torch::Tensor &worldToCameraMatrices,
+                          const torch::Tensor &projectionMatrices,
+                          const size_t imageWidth,
+                          const size_t imageHeight,
+                          const float near,
+                          const float far,
+                          const fvdb::detail::ops::DistortionModel cameraModel =
+                              fvdb::detail::ops::DistortionModel::PINHOLE,
+                          const std::optional<torch::Tensor> &distortionCoeffs = std::nullopt,
+                          const int64_t shDegreeToUse                          = -1,
```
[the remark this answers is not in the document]
fwilliams: This was introduced between releases and we don't need backward compatibility.

src/python/GaussianSplatBinding.cpp
```diff
@@ -20,14 +20,14 @@ bind_gaussian_splat3d(py::module &m) {
...
+        .value("OPENCV_RADTAN_5", fvdb::detail::ops::DistortionModel::OPENCV_RADTAN_5)
+        .value("OPENCV_RATIONAL_8", fvdb::detail::ops::DistortionModel::OPENCV_RATIONAL_8)
         .value("OPENCV_RADTAN_THIN_PRISM_9",
-               fvdb::detail::ops::CameraModel::OPENCV_RADTAN_THIN_PRISM_9)
-        .value("OPENCV_THIN_PRISM_12", fvdb::detail::ops::CameraModel::OPENCV_THIN_PRISM_12)
-        .value("ORTHOGRAPHIC", fvdb::detail::ops::CameraModel::ORTHOGRAPHIC)
+               fvdb::detail::ops::DistortionModel::OPENCV_RADTAN_THIN_PRISM_9)
+        .value("OPENCV_THIN_PRISM_12", fvdb::detail::ops::DistortionModel::OPENCV_THIN_PRISM_12)
+        .value("ORTHOGRAPHIC", fvdb::detail::ops::DistortionModel::ORTHOGRAPHIC)
         .export_values();
```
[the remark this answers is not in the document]
fwilliams: As above, we don't need backwards compatibility here.

fwilliams (commented):
Some questions and comments

src/fvdb/GaussianSplat3d.h
```diff
@@ -915,26 +915,28 @@ class GaussianSplat3d {
     /// geometry gradients through the rasterization step.
     ///
     /// Tile intersections are still computed using a (non-differentiable) projection step:
-    /// - For `cameraModel == CameraModel::PINHOLE` or `CameraModel::ORTHOGRAPHIC`, we reuse the
+    /// - For `cameraModel == DistortionModel::PINHOLE` or `DistortionModel::ORTHOGRAPHIC`, we reuse
+    /// the
     ///   classic projection path.
     /// - For OpenCV camera models, we use the Unscented Transform (UT) projection kernel to
```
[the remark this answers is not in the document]
fwilliams: Fixed in this branch: re-wrapped the doc sentence in `GaussianSplat3d.h` so the generated docs read cleanly.

src/fvdb/detail/ops/gsplat/GaussianProjectionJaggedBackward.h
fwilliams: Addressed in this PR update: reverted argument-order-only churn so signatures/calls keep the prior ordering unless behavior required a change.

src/fvdb/detail/ops/gsplat/GaussianProjectionJaggedForward.h
fwilliams: Addressed in this PR update: reverted argument-order-only churn so signatures/calls keep the prior ordering unless behavior required a change.

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
fwilliams: Addressed in this PR update: reverted argument-order-only churn so signatures/calls keep the prior ordering unless behavior required a change.

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
fwilliams: Addressed in this PR update: reverted argument-order-only churn so signatures/calls keep the prior ordering unless behavior required a change.

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
fwilliams: Addressed in this PR update: reverted argument-order-only churn so signatures/calls keep the prior ordering unless behavior required a change.

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.h
fwilliams: Addressed in this PR update: reverted argument-order-only churn so signatures/calls keep the prior ordering unless behavior required a change.

src/fvdb/detail/ops/gsplat/GaussianProjectionForward.h
fwilliams: Addressed in this PR update: reverted argument-order-only churn so signatures/calls keep the prior ordering unless behavior required a change.

src/fvdb/detail/ops/gsplat/GaussianProjectionJaggedBackward.cu
fwilliams: Addressed in this PR update: reverted argument-order-only churn so signatures/calls keep the prior ordering unless behavior required a change.

src/fvdb/detail/autograd/GaussianRasterizeFromWorld.cpp
fwilliams: Updated for consistency: renamed local/forwarded variables to `distortionModel` in the rasterize-from-world autograd path.

src/fvdb/detail/autograd/GaussianRasterizeFromWorld.h
fwilliams: Updated for consistency: renamed local/forwarded variables to `distortionModel` in the rasterize-from-world autograd path.

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
fwilliams: Added class/method documentation in `GaussianCameras.cuh` for the camera-op classes and key methods (shared-memory setup, visibility/depth checks, projection/unprojection helpers).

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
fwilliams: Added class/method documentation in `GaussianCameras.cuh` for the camera-op classes and key methods (shared-memory setup, visibility/depth checks, projection/unprojection helpers).

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
fwilliams: Added class/method documentation in `GaussianCameras.cuh` for the camera-op classes and key methods (shared-memory setup, visibility/depth checks, projection/unprojection helpers).

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
fwilliams: Added class/method documentation in `GaussianCameras.cuh` for the camera-op classes and key methods (shared-memory setup, visibility/depth checks, projection/unprojection helpers).

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
fwilliams: Looked at this again: we currently keep `ProjectWorldToPixelStatus` visible because `GaussianProjectionUT.cu` consumes the status (`BehindCamera`/`InImage`) outside the camera-op class. Keeping it exposed is intentional for now; if we later fold that logic into camera-op methods we can make it internal then.

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
fwilliams: Addressed with a low-risk API trim in this pass: removed `imageWidthPx()` / `imageHeightPx()` from the projection camera ops and replaced callsites with a narrower `projectedFootprintOutsideImage(...)` helper. We can continue API-surface reduction in follow-ups, but this keeps behavior unchanged and shrinks the exposed interface.

src/fvdb/detail/ops/gsplat/GaussianProjectionJaggedBackward.cu
fwilliams: Still agree with this direction. I left this thread unresolved on purpose and deferred full `R/t` encapsulation to a focused follow-up, since it requires moving both transform and VJP coupling into camera-op APIs rather than a localized edit.

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
fwilliams: Still agree with this direction. I left this thread unresolved on purpose and deferred full `R/t` encapsulation to a focused follow-up, since it requires moving both transform and VJP coupling into camera-op APIs rather than a localized edit.

src/fvdb/detail/ops/gsplat/GaussianRasterize.cuh
fwilliams: Good callout. For this PR, I kept the broader `mXXX`-state privatization out of scope (to avoid mixing with camera refactor behavior changes). I left this unresolved so we can track it in the upcoming RasterizeCommonArgs/tile-intersections cleanup PR.

src/tests/GaussianProjectionBackwardTest.cpp
```diff
@@ -177,14 +178,14 @@ TEST_F(GaussianProjectionBackwardTestFixture, DISABLED_GenerateOutputData) {
...
+                                                                               false,
                                                                                imageWidth,
                                                                                imageHeight,
                                                                                0.3,
                                                                                1e-2,
                                                                                1e10,
                                                                                0,
-                                                                               true,
-                                                                               false);
+                                                                               true);
```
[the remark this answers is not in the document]
fwilliams: Fixed in b54e44c. You were right: the disabled data-generator calls in `GaussianProjectionBackwardTest.cpp` had arguments shifted after the forward signature reorder. I updated both callsites to `..., imageWidth, imageHeight, eps2d, near, far, minRadius2d, calcCompensations, ortho` and re-scanned for similar `Ks, false, imageWidth` patterns; no other instances remain.

src/fvdb/detail/autograd/GaussianProjection.cpp
```diff
@@ -35,7 +35,6 @@ ProjectGaussians::forward(ProjectGaussians::AutogradContext *ctx,
     TORCH_CHECK(means.dim() == 2, "means must have shape (N, 3)");
     TORCH_CHECK(worldToCamMatrices.dim() == 3, "worldToCamMatrices must have shape (C, 4, 4)");
     TORCH_CHECK(projectionMatrices.dim() == 3, "projectionMatrices must have shape (C, 3, 3)");
-
```
fwilliams, line 38: The changes in this file are all review noise (whitespace). We should clean these up to keep the PR tighter.

src/fvdb/detail/ops/gsplat/GaussianProjectionJaggedForward.cu
```diff
@@ -135,7 +97,7 @@ jaggedProjectionForwardKernel(const uint32_t B,
     }
 
     // mask out gaussians outside the image region
-    if (isOutsideImageWithRadius(mean2d, radius, radius, imageWidth, imageHeight)) {
+    if (cameraOp.projectedFootprintOutsideImage(mean2d, radius, radius)) {
```
fwilliams: We might want to have predicates have symmetric naming on camera ops. Above, we use `isDepthVisible`. The name `isProjectedFootprintOutsideImage` indicates that this is also a predicate and is symmetric with `isDepthVisible`

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldBackward.cu
```diff
@@ -66,31 +67,14 @@ rasterizeFromWorld3DGSBackwardKernel(const RasterizeFromWorldBackwardArgs<NUM_CH
...
-                                                                 common.rollingShutterType,
-                                                                 common.cameraModel);
+    extern __shared__ char smem[];
+    CameraOp cameraOpLocal = args.cameraOp;
+    uintptr_t smemAddr     = reinterpret_cast<uintptr_t>(smem);
+    smemAddr               = alignUpAddress(smemAddr, alignof(nanovdb::math::Mat3<float>));
+    cameraOpLocal.loadSharedMemory(reinterpret_cast<void *>(smemAddr));
+    block.sync();
+
+    const nanovdb::math::Ray<float> ray = cameraOpLocal.projectToRay(camId, row, col);
```
fwilliams: Is `projectToRay` a good name for this method? It seems it's converting a pixel coordinate (row, col) for a given camera (camId) to a ray through that pixel.

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
fwilliams: What would folding the logic into the camera op look like code-wise?

src/fvdb/detail/ops/gsplat/GaussianProjectionJaggedForward.cu
fwilliams: Applied in 122ad74. Renamed the predicate to `isProjectedFootprintOutsideImage(...)` and updated callsites for symmetry with other predicate-style camera-op APIs (`isDepthVisible(...)`).

src/fvdb/detail/ops/gsplat/GaussianProjectionBackward.cu
```diff
@@ -347,64 +301,74 @@ dispatchGaussianProjectionBackward<torch::kCUDA>(
...
-                covars.has_value() ? nullptr : dLossDScales.data_ptr<float>(),
-                worldToCamMatricesRequiresGrad ? dLossDWorldToCamMatrices.data_ptr<float>()
-                                               : nullptr);
+            const auto cameraOp     = OrthographicCameraOp<float>{projectionMatrices,
+                                                                  worldToCamMatrices,
+                                                                  static_cast<int32_t>(C),
+                                                                  static_cast<int32_t>(imageWidth),
+                                                                  static_cast<int32_t>(imageHeight),
+                                                                  -1e10f,
+                                                                  1e10f};
```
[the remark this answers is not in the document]
fwilliams: Implemented in 122ad74. I replaced the repeated `-1e10f/1e10f` backward-only projection bounds with named constants (`kBackwardProjectionNearPlane`, `kBackwardProjectionFarPlane`) in `GaussianCameras.cuh`, and updated all callsites in both dense and jagged backward projection kernels.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorldBackward.cu
fwilliams: Updated in 122ad74. Renamed `projectToRay(...)` to `unprojectPixelToRay(...)` in distortion camera ops and rasterize-from-world callsites to make the pixel->ray direction explicit.

src/fvdb/detail/autograd/GaussianProjection.cpp
fwilliams: Fixed in 122ad74. I removed the formatting-only churn from this file to keep the PR diff focused on behavior/API changes.

src/fvdb/detail/autograd/GaussianRasterizeFromWorld.cpp
```diff
@@ -100,7 +100,7 @@ RasterizeGaussiansToPixelsFromWorld3DGS::forward(
     ctx->saved_data["imageOriginW"]       = (int64_t)imageOriginW;
     ctx->saved_data["imageOriginH"]       = (int64_t)imageOriginH;
     ctx->saved_data["tileSize"]           = (int64_t)tileSize;
-    ctx->saved_data["cameraModel"]        = (int64_t)cameraModel;
+    ctx->saved_data["distortionModel"]    = (int64_t)distortionModel;
```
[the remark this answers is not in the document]
fwilliams: Good point to call out. For this branch we are intentionally not preserving backward compatibility for pre-release saved autograd metadata keys, consistent with the earlier CameraModel->DistortionModel decision in this PR. I do not plan to dual-write legacy `"cameraModel"` in this branch.

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
fwilliams: Concretely, the fold-in path is what this PR already introduces for projection camera ops: kernels call `cameraOp.projectWorldGaussianTo2D(...)` / `projectWorldGaussianTo2DVJP(...)`, and camera ops own world->camera transform + covariance transform + projection + VJP chaining internally. For distortion/raster-from-world, this PR keeps the same architecture direction and additionally tightens naming (`unprojectPixelToRay`) and predicates, while full internalization of any remaining transform plumbing can continue in follow-ups without changing kernel contracts.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
```diff
@@ -125,6 +115,7 @@ struct RasterizeFromWorldCommonArgs {
     }
 };
 
+/// Vector-Jacobian product for y = normalizeSafe(x).
 template <typename T>
 inline __device__ nanovdb::math::Vec3<T>
 normalizeSafe(const nanovdb::math::Vec3<T> &v) {
```
harrism, line 123: **🚩 issue:** ‏Looks like you moved the docs from the VJP version here, which isn't correct? And now `normalizeSafeVJP` has no docs.

src/fvdb/detail/autograd/GaussianProjection.cpp
```diff
@@ -293,8 +293,7 @@ ProjectGaussiansJagged::backward(ProjectGaussiansJagged::AutogradContext *ctx,
     const int imageHeight = (int)ctx->saved_data["imageHeight"].toInt();
     const float eps2d     = (float)ctx->saved_data["eps2d"].toDouble();
     const bool ortho      = (bool)ctx->saved_data["ortho"].toBool();
-
-    auto variables       = FVDB_DISPATCH_KERNEL_DEVICE(means.device(), [&]() {
+    auto variables        = FVDB_DISPATCH_KERNEL_DEVICE(means.device(), [&]() {
```
harrism: **❓ question:** ‏Did you change clang-format versions? These seem like trivial alignment changes that should be automated.

src/fvdb/detail/autograd/GaussianRasterize.cpp
```diff
@@ -30,14 +32,13 @@ RasterizeGaussiansToPixels::forward(
     // const int N = means2d.size(1);
 
     auto variables          = FVDB_DISPATCH_KERNEL(means2d.device(), [&]() {
+        const auto renderWindow =
```
harrism: **🚩 issue:** ‏I don't really like using a generic vector for this. I think it would be less bug-prone to create a simple struct with named members.

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
```diff
@@ -0,0 +1,1383 @@
...
+template <typename T>
+inline __device__ T
+radiusFromCovariance2dDet(const nanovdb::math::Mat2<T> &covar2d,
+                          const T det,
+                          const T sigma,
+                          const T detClamp = T(0.01)) {
+    // Matches the classic FVDB/gsplat heuristic: use the largest eigenvalue of the 2x2 covariance
+    // (via trace/determinant) and take `sigma` standard deviations.
+    const T b  = T(0.5) * (covar2d[0][0] + covar2d[1][1]);
+    const T v1 = b + ::cuda::std::sqrt(nanovdb::math::Max(detClamp, b * b - det));
```
harrism, line 100: **💡 suggestion:** ‏Use `cuda::std::max` here for consistency.

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
```diff
@@ -0,0 +1,1383 @@
...
+                                     const T radiusY) const {
+        return isOutsideImageWithRadius(mean2d, radiusX, radiusY, imageWidth, imageHeight);
+    }
+
+    /// @brief Returns world-to-camera rotation and translation for camera `cid`.
+    ///
+    /// The returned transform maps a world-space point `p_w` to camera space via:
+    /// `p_c = R * p_w + t`.
+    inline __device__ std::tuple<Mat3, Vec3>
+    worldToCamRt(const int64_t cid) const {
```
harrism: **💡 suggestion:** ‏Call this `worldToCamTransform`

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
```diff
@@ -0,0 +1,1383 @@
...
+            Vec3(W[0][3], W[1][3], W[2][3]));
+    }
+
+    /// @brief Projects a camera-space 3D Gaussian to pixel-space `(cov2d, mean2d)`.
+    ///
+    /// Returns `(cov2d, mean2d)` where `cov2d` is the projected 2x2 covariance and `mean2d`
+    /// is the projected pixel mean.
+    inline __device__ std::tuple<nanovdb::math::Mat2<T>, nanovdb::math::Vec2<T>>
+    projectTo2DGaussian(const int64_t cid,
+                        const nanovdb::math::Vec3<T> &meansCamSpace,
```
harrism: **💡 suggestion:** ‏Singular mean. 
```suggestion
                        const nanovdb::math::Vec3<T> &meanCamSpace,
```

harrism (commented):
Factoring out the camera functionality is a nice simplification. I wonder if instead of CameraOp it should just be called Camera? Just a few other suggestions in comments.

src/fvdb/detail/autograd/GaussianProjection.cpp
fwilliams: Good catch — agreed. I removed the formatting-only churn in `GaussianProjection.cpp` so this file no longer carries review-noise-only edits in the PR diff. I also verified with a whitespace-insensitive diff (`git diff -w --ignore-blank-lines`) that no semantic change remains for that cleanup.

src/fvdb/detail/ops/gsplat/GaussianRasterizeFromWorld.cuh
fwilliams: Good catch. I fixed this in `f68a070`: `normalizeSafe(...)` now has its own doc comment, and `normalizeSafeVJP(...)` has explicit VJP documentation again.

src/fvdb/detail/autograd/GaussianProjection.cpp
fwilliams: No clang-format version change here. The alignment churn was noise; I removed it in `3894680` and verified `GaussianProjection.cpp` has no semantic diff for that cleanup (`git diff -w --ignore-blank-lines`).

src/fvdb/detail/autograd/GaussianRasterize.cpp
fwilliams: Agreed, and now done in `f68a070`: this path uses `RenderWindow2D` (named fields) instead of a generic vector, including kernel dispatch signatures.

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
fwilliams: Applied in `c088e5a`: renamed to `worldToCamTransform` and updated callsites.

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
fwilliams: Applied in `c088e5a`: renamed local/parameter usage to singular `meanCamSpace` for consistency.

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
fwilliams: Great suggestion on consistency. I tried switching this site to `cuda::std::max` and reran the branch build flow (`conda run -n fvdb ./build.sh install gtests`) in the `fvdb` env. In this toolchain, that fails with `namespace "cuda::std" has no member "max"` at this callsite, and first breaks in projection TUs (`GaussianProjectionForward.cu`, `GaussianProjectionUT.cu`, `GaussianProjectionBackward.cu`, `GaussianProjectionJaggedBackward.cu`).

So I kept `nanovdb::math::Max` here for compatibility. If you prefer, I can switch to one of these instead:
1) keep `nanovdb::math::Max` (current, builds),
2) use `cuda::std::fmax` for floating-point max, or
3) add a tiny local compatibility helper/wrapper and use that at callsites.

Happy to follow your preferred style.

harrism: A human would search and understand. https://nvidia.github.io/cccl/unstable/libcudacxx/standard_api/algorithms_library.html

It's in the CCCL `<cuda/std/algorithm>` header. However, that doesn't seem to exist until CCCL 3.2/CUDA 13.2. We may want to move to explicitly installing a specific CCCL in the future to expand its capabilities.

For now, just stick with what you have here. (although I think `std::max` or just `max` should work).

src/fvdb/detail/ops/gsplat/GaussianRasterize.cuh
```diff
@@ -183,19 +178,19 @@ template <typename ScalarType, size_t NUM_CHANNELS, bool IS_PACKED> struct Raste
 
     inline __host__ __device__ uint32_t
     renderWidth() const {
```
harrism, line 180: **🤓 nitpick:** ‏These functions seem a bit superfluous now.

fwilliams: Agreed, Will fix this up in the follow on PR, so I can merge this one without waiting on CI

PR #486, opened 2026-02-27
Fix datatype in backward projection test

matthewdcong:
The two tensors are integral so we should be checking to see if they're equal rather than close.


swahtz (approved):
Good catch, thanks

PR #488, opened 2026-03-03
Handle duplicate pixels in sparse pixel gaussian rendering

### LINKED ISSUE #106: Handle duplicate pixels in sparse pixel gaussian rendering
## 📋 Originally created by @fwilliams

> **Original author:** @fwilliams
> **Date:** 2025-02-12T00:03:49Z

---

We will squash duplicate pixels in the current code. Let's handle this properly.

harrism:
## Summary

- Deduplicate pixel coordinates at the binding layer (`deduplicatePixels`) before passing to `computeSparseInfo` and rasterization kernels, then scatter results back via `index_select`. This avoids any CUDA kernel modifications while correctly handling duplicate `(batchIdx, row, col)` entries that would otherwise cause incorrect tile bitmasks.
- Handle edge cases: single-list JaggedTensors with empty `jidx`, correct `jlidx` passthrough for `ldim()` consistency, and reducing `numContributingGaussians` to unique-pixel space for the contributing IDs kernel.
- Add comprehensive C++ unit tests for `deduplicatePixels` (20 tests covering empty, single, all-unique, duplicates, multi-batch, round-trip reconstruction, and per-batch offset validation) and Python end-to-end tests for all sparse render APIs with duplicate pixels (6 tests covering depths, images, backward gradients, num_contributing, contributing_ids, and multi-camera).

Fixes #106

## Test plan

- [x] C++ `DeduplicatePixelsTest`: 20/20 pass
- [x] C++ `GaussianComputeSparseInfoTest`: 14/14 pass
- [x] Python `TestGaussianRenderSparseDuplicatePixels`: 6/6 pass

### DIFF
--- a/fvdb/viz/_viewer_server.py
+++ b/fvdb/viz/_viewer_server.py
@@ -1,6 +1,7 @@
 # Copyright Contributors to the OpenVDB Project
 # SPDX-License-Identifier: Apache-2.0
 #
+import uuid
 import warnings
 import webbrowser
 
@@ -24,7 +25,12 @@ def _get_viewer_server_cpp() -> ViewerCpp:
     return _viewer_server_cpp
 
 
-def init(ip_address: str = "127.0.0.1", port: int = 8080, vk_device_id: int = 0, verbose: bool = False):
+def init(
+    ip_address: str = "127.0.0.1",
+    port: int = 8080,
+    vk_device_id: int = 0,
+    verbose: bool = False,
+):
     """
     Initialize the viewer web-server on the given IP address and port. You must call this function
     first before visualizing any scenes.
@@ -115,10 +121,45 @@ def show():
 
     try:
         from IPython import get_ipython
-        from IPython.display import IFrame, display
+        from IPython.display import HTML, display
 
         if get_ipython() is not None:
-            display(IFrame(src=url, width="100%", height="600px"))
+            # Generate a unique ID for this viewer instance
+            viewer_id = f"viewer-container-{uuid.uuid4().hex[:8]}"
+
+            html_content = f"""
+            <div id="{viewer_id}"></div>
+            <script>
+                (function() {{
+                    // Get the current page's protocol and hostname (without port)
+                    // This ensures we construct the URL correctly even when Jupyter
+                    // is running on a non-default port
+                    var protocol = window.location.protocol; // e.g., "http:" or "https:"
+                    var hostname = window.location.hostname; // e.g., "localhost" or "example.com"
+                    var viewerPort = "{viewer_server_port}";
+
+                    // Explicitly construct the viewer URL
+                    var viewerUrl = protocol + "//" + hostname + ":" + viewerPort;
+
+                    console.log("Viewer URL: " + viewerUrl);
+
+                    // Create and insert the iframe
+                    var iframe = document.createElement('iframe');
+                    iframe.src = viewerUrl;
+                    iframe.width = '100%';
+                    iframe.height = '600px';
+                    iframe.style.border = 'none';
+
+                    var container = document.getElementById('{viewer_id}');
+                    if (container) {{
+                        container.appendChild(iframe);
+                    }} else {{
+                        console.error('Could not find container element with id: {viewer_id}');
+                    }}
+                }})();
+            </script>
+            """
+            display(HTML(html_content))
             return
     except ImportError:
         pass
--- a/src/fvdb/GridBatch.cpp
+++ b/src/fvdb/GridBatch.cpp
@@ -243,12 +243,14 @@ GridBatch::max_pool(Vec3iOrScalar pool_factor,
                     const JaggedTensor &data,
                     Vec3iOrScalar stride,
                     std::optional<GridBatch> coarse_grid) const {
-    c10::DeviceGuard guard(device());
+    // TODO: ldim checks live here because the underlying ops accept plain torch::Tensor (via
+    // .jdata()). When these member functions are removed in favor of free-function ops, move checks
+    // to Python.
     TORCH_CHECK_VALUE(
         data.ldim() == 1,
-        "Expected data to have 1 list dimension, i.e. be a single list of coordinate values, but got",
+        "Expected data to have 1 list dimension, i.e. be a single list of coordinate values, but got ",
         data.ldim(),
-        "list dimensions");
+        " list dimensions");
 
     nanovdb::Coord pool_factor_coord = pool_factor.value();
     nanovdb::Coord stride_coord      = stride.value();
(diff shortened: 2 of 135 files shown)


src/fvdb/GaussianSplat3d.cpp
```diff
@@ -392,6 +390,89 @@ GaussianSplat3d::projectGaussiansImpl(const torch::Tensor &worldToCameraMatrices
...
+    // Build new JaggedTensor offsets for the unique pixels
+    auto uniqueBatchIdx = jidxLong.index_select(0, uniqueOrigIndices);
+    auto numLists       = pixelsToRender.num_outer_lists();
+    auto countsPerList  = torch::zeros({numLists}, longOpts);
+    countsPerList.scatter_add_(
+        0, uniqueBatchIdx.to(torch::kLong), torch::ones_like(uniqueBatchIdx, torch::kLong));
+    auto newOffsets = torch::zeros({numLists + 1}, longOpts);
+    newOffsets.slice(0, 1).copy_(countsPerList.cumsum(0));
+
+    auto newJidx = JaggedTensor::jidx_from_joffsets(newOffsets, numUnique);
```
swahtz: **💡 suggestion:** ‏I think `uniqueBatchIdx` already has the values you want for the new jagged indices.  You could just do this and save the binary search + allocation

```suggestion
    auto newJidx = uniqueBatchIdx.to(fvdb::JIdxScalarType);
```

src/fvdb/GaussianSplat3d.cpp
```diff
@@ -392,6 +390,89 @@ GaussianSplat3d::projectGaussiansImpl(const torch::Tensor &worldToCameraMatrices
...
+    auto firstInSorted     = isGroupStart.nonzero().squeeze(1);
+    auto uniqueOrigIndices = sortPerm.index_select(0, firstInSorted);
+    auto uniqueJData       = jdata.index_select(0, uniqueOrigIndices);
+
+    // Build new JaggedTensor offsets for the unique pixels
+    auto uniqueBatchIdx = jidxLong.index_select(0, uniqueOrigIndices);
+    auto numLists       = pixelsToRender.num_outer_lists();
+    auto countsPerList  = torch::zeros({numLists}, longOpts);
+    countsPerList.scatter_add_(
+        0, uniqueBatchIdx.to(torch::kLong), torch::ones_like(uniqueBatchIdx, torch::kLong));
```
swahtz: **💡 suggestion:** ‏`uniqueBatchIdx` is already a long because it came from `jidxLong` so you could save a dispatch.  `torch::ones` can also be used instead of  `ones_like` (save on querying the shape of `uniqueBatchIdx` which we already have in `numUnique`).

```suggestion
        0, uniqueBatchIdx, torch::ones({numUnique}, longOpts));
```

swahtz: **💡 suggestion:** ‏ On further thought, we could use `torch::bincount` to do these 2 lines in 1 call instead of 3.

```suggestion
    auto countsPerList = torch::bincount(uniqueBatchIdx, {}, numLists);
```

swahtz (approved):
Looks great I just had a few potential optimization suggestions since these will get called on every rendering iteration. 

I was also wondering for the unit tests if it would be worth having a test where all pixels in each batch are duplicates or perhaps that's redundant with the tests were just some pixels are duplicates.

src/fvdb/GaussianSplat3d.cpp
harrism: Great catch!

harrism: Took the other suggestion.

src/fvdb/GaussianSplat3d.cpp
harrism: Wow, this is what I get for trusting the LLM. Good catch.

harrism:
Great suggestions @swahtz . Implemented your two optimizations and found one other and a stale comment. Also added your suggested maximum compression test case.

src/fvdb/GaussianSplat3d.cpp
```diff
@@ -393,6 +393,87 @@ GaussianSplat3d::projectGaussiansImpl(const torch::Tensor &worldToCameraMatrices
...
+    const auto jidx                 = pixelsToRender.jidx();
+    const int64_t numPixelsPerImage = imageHeight * imageWidth;
+    const auto longOpts             = torch::TensorOptions().device(device).dtype(torch::kLong);
+    const auto boolOpts             = torch::TensorOptions().device(device).dtype(torch::kBool);
+
+    // Encode (batchIdx, row, col) into a single int64 key:
+    //   key = batchIdx * (H * W) + row * W + col
+    // For single-list JaggedTensors, jidx may be empty; synthesize zeros.
+    auto jidxLong =
+        jidx.size(0) == 0 ? torch::zeros({totalPixels}, longOpts) : jidx.to(torch::kLong);
```
fwilliams: why not use an empty jidxLong instead of zeros here?

src/fvdb/GaussianSplat3d.cpp
```diff
@@ -393,6 +393,87 @@ GaussianSplat3d::projectGaussiansImpl(const torch::Tensor &worldToCameraMatrices
...
+    // Sort keys and find group boundaries
+    auto [sortedKeys, sortPerm] = keys.sort();
+
+    auto isGroupStart = torch::ones({totalPixels}, boolOpts);
+    if (totalPixels > 1) {
+        isGroupStart.slice(0, 1).copy_(sortedKeys.slice(0, 1) != sortedKeys.slice(0, 0, -1));
+    }
+
+    // Assign a group ID (0-based) to each sorted position via cumsum
+    auto groupIds        = isGroupStart.cumsum(0) - 1;
```
fwilliams: you can do an in place cumsum instead of creating a new tensor.  you'll have to compute firstInSorted before you do this (it's currently below)

src/fvdb/GaussianSplat3d.cpp
harrism: Can do this. It requires some branching but that's OK. Saves a bit of memory and fill and it's easy.

src/fvdb/GaussianSplat3d.cpp
harrism: Good point.

harrism:
I decided against the custom kernel approach because by definition these are low-cost operations in terms of memory and computation. For sparse rendering, N (number of pixels) should be small -- say 5-50K. I estimate memory savings of a custom kernel to be at most ~1MB at that scale. Even if we render at the scale of 500K pixels (25% of a 1080p image) we're talking only 10MB savings. On the scale of 3DGS memory usage, this is peanuts.

The main reason for custom kernels here would be launch overhead and possibly synchronization reduction. But I don't think that should be addressed until it is determined to be a bottleneck.

harrism:
@fwilliams implemented the small optimizations you suggested. Torch cumsum_ doesn't work on bool tensors in CUDA, so had to cast it, so the memory savings is reduced a bit but there's still about 8N bytes savings (1 fewer allocation) for that one (rather than 16N).

PR #492, opened 2026-03-03
Feature/op consolidation

blackencino:
This is a largely automated PR which moves all of the OP dispatch and precondition code into each op's C++ implementation files, hiding the dispatch and template functions from view. This decreases the compile-time interconnectivity between components and prepares for simplification of the C++ side of GridBatch. The public API does not change in any way.

### DIFF
--- a/src/fvdb/detail/ops/BuildDilatedGrid.cu
+++ b/src/fvdb/detail/ops/BuildDilatedGrid.cu
@@ -40,20 +40,34 @@ dispatchDilateGrid<torch::kCUDA>(const GridBatchImpl &gridBatch,
     std::vector<nanovdb::GridHandle<TorchDeviceBuffer>> handles;
     for (int i = 0; i < gridBatch.batchSize(); i += 1) {
         nanovdb::GridHandle<TorchDeviceBuffer> handle;
-        nanovdb::OnIndexGrid *grid =
-            gridBatch.nanoGridHandleMut().deviceGrid<nanovdb::ValueOnIndex>(i);
-        TORCH_CHECK(grid, "Grid is null");
 
-        for (auto j = 0; j < dilationAmount[i]; j += 1) {
-            nanovdb::tools::cuda::DilateGrid<nanovdb::ValueOnIndex> dilateOp(grid, stream);
-            dilateOp.setOperation(nanovdb::tools::morphology::NN_FACE_EDGE_VERTEX);
-            dilateOp.setChecksum(nanovdb::CheckMode::Default);
-            dilateOp.setVerbose(0);
+        if (dilationAmount[i] == 0) {
+            const auto &srcHdl = gridBatch.nanoGridHandle();
+            uint8_t *srcData   = srcHdl.buffer().deviceData();
+            for (int g = 0; g < i; ++g) {
+                srcData += srcHdl.gridSize(g);
+            }
+            const uint64_t gridBytes = srcHdl.gridSize(i);
+            TorchDeviceBuffer dstBuf(gridBytes, gridBatch.device());
+            cudaMemcpyAsync(
+                dstBuf.deviceData(), srcData, gridBytes, cudaMemcpyDeviceToDevice, stream.stream());
+            handle = nanovdb::GridHandle<TorchDeviceBuffer>(std::move(dstBuf));
+        } else {
+            nanovdb::OnIndexGrid *grid =
+                gridBatch.nanoGridHandleMut().deviceGrid<nanovdb::ValueOnIndex>(i);
+            TORCH_CHECK(grid, "Grid is null");
 
-            handle = dilateOp.getHandle(guide);
-            C10_CUDA_KERNEL_LAUNCH_CHECK();
+            for (auto j = 0; j < dilationAmount[i]; j += 1) {
+                nanovdb::tools::cuda::DilateGrid<nanovdb::ValueOnIndex> dilateOp(grid, stream);
+                dilateOp.setOperation(nanovdb::tools::morphology::NN_FACE_EDGE_VERTEX);
+                dilateOp.setChecksum(nanovdb::CheckMode::Default);
+                dilateOp.setVerbose(0);
 
-            grid = handle.deviceGrid<nanovdb::ValueOnIndex>();
+                handle = dilateOp.getHandle(guide);
+                C10_CUDA_KERNEL_LAUNCH_CHECK();
+
+                grid = handle.deviceGrid<nanovdb::ValueOnIndex>();
+            }
         }
 
         handles.push_back(std::move(handle));
@@ -130,6 +144,13 @@ dilateGrid(const GridBatchImpl &gridBatch, const std::vector<int64_t> &dilationA
                                   dilationAmount.end(),
                                   [](int64_t amount) { return amount >= 0; }),
                       "dilation amount must be non-negative.");
+
+    if (std::all_of(dilationAmount.begin(), dilationAmount.end(), [](int64_t amount) {
+            return amount == 0;
+        })) {
+        return gridBatch.clone(gridBatch.device());
+    }
+
     std::vector<nanovdb::Vec3d> voxS, voxO;
     gridBatch.gridVoxelSizesAndOrigins(voxS, voxO);
     auto hdl = FVDB_DISPATCH_KERNEL_DEVICE(gridBatch.device(), [&]() {


src/fvdb/detail/autograd/AvgPoolGrid.cpp
```diff
@@ -51,12 +48,10 @@ AvgPoolGrid::backward(AvgPoolGrid::AutogradContext *ctx, AvgPoolGrid::variable_l
     const int64_t strideZ        = ctx->saved_data["stride_z"].toInt();
     const nanovdb::Coord poolingFactor(poolingFactorX, poolingFactorY, poolingFactorZ);
     const nanovdb::Coord stride(strideX, strideY, strideZ);
-    Variable gradOut = grad_output.at(0).contiguous(); // [#coarse_voxels | #coarse_corners, *]
+    Variable gradOut = grad_output.at(0); // [#coarse_voxels | #coarse_corners, *]
```
fwilliams, line 51: why did we make this continuous to begin with?

src/fvdb/detail/autograd/ReadIntoDense.cpp
```diff
@@ -176,47 +114,33 @@ ReadIntoDenseCmajor::forward(AutogradContext *ctx,
...
     torch::Tensor retShape =
         torch::empty({(int64_t)sparseData.dim()}, torch::TensorOptions().dtype(torch::kLong));
     auto acc = retShape.accessor<int64_t, 1>();
     for (int i = 0; i < sparseData.dim(); i++) {
         acc[i] = sparseData.size(i);
     }
-    ctx->saved_data["final_shape"]  = retShape;
-    ctx->saved_data["first_dim"]    = sparseDataReshape.size(0);
-    ctx->saved_data["last_dim"]     = sparseDataReshape.size(1);
-    ctx->saved_data["dummy_tensor"] = torch::empty({0}, sparseData.options());
```
fwilliams, line 195: why did we have a dummy tensor here? what happened to last dim?

src/fvdb/detail/autograd/ReadIntoDense.cpp
```diff
@@ -176,47 +114,33 @@ ReadIntoDenseCmajor::forward(AutogradContext *ctx,
...
 
 ReadIntoDenseCmajor::variable_list
 ReadIntoDenseCmajor::backward(AutogradContext *ctx, variable_list grad_output) {
-    // Use data saved in forward
     torch::Tensor denseOrigins = ctx->saved_data["dense_origins"].toTensor(); // [B, 3]
-    int64_t firstDim           = ctx->saved_data["first_dim"].toInt();
-    int64_t lastDim            = ctx->saved_data["last_dim"].toInt();
     std::vector<int64_t> finalShapeTensor =
         intTensor1DToStdVector(ctx->saved_data["final_shape"].toTensor());
-    torch::TensorOptions sparseDataOpts = ctx->saved_data["dummy_tensor"].toTensor().options();
```
fwilliams, line 209: same question as above? what happened to the dummy tensor?

src/fvdb/detail/ops/BuildGridFromNearestVoxelsToPoints.h
```diff
@@ -6,25 +6,19 @@
...
 namespace detail {
 namespace ops {
 
-template <torch::DeviceType>
-nanovdb::GridHandle<TorchDeviceBuffer>
-dispatchBuildGridFromNearestVoxelsToPoints(const JaggedTensor &points,
-                                           const std::vector<VoxelCoordTransform> &txs);
+c10::intrusive_ptr<GridBatchImpl>
+buildGridFromNearestVoxelsToPoints(const JaggedTensor &points,
+                                   const std::vector<nanovdb::Vec3d> &voxelSizes,
```
fwilliams, line 18: is VoxelCoordTransform now internal to ops?

src/fvdb/GridBatch.cpp
```diff
@@ -243,13 +243,6 @@ GridBatch::max_pool(Vec3iOrScalar pool_factor,
...
                     Vec3iOrScalar stride,
                     std::optional<GridBatch> coarse_grid) const {
-    c10::DeviceGuard guard(device());
-    TORCH_CHECK_VALUE(
-        data.ldim() == 1,
-        "Expected data to have 1 list dimension, i.e. be a single list of coordinate values, but got",
-        data.ldim(),
-        "list dimensions");
-
     nanovdb::Coord pool_factor_coord = pool_factor.value();
```
harrism, line 255: The `data.ldim() == 1` check that was here (and in `avg_pool`, `refine`, `write_to_dense_cminor/cmajor`, and all `sample_*`/`splat_*` functions) can't be replicated downstream because `.jdata()` strips the JaggedTensor structure before the op sees it. I'd recommend keeping these specific `ldim` checks in `GridBatch.cpp` — they're the last point where list-structure validation is possible.

src/fvdb/detail/ops/IjkToInvIndex.cu
```diff
@@ -84,20 +85,15 @@ IjkToInvIndex(const GridBatchImpl &batchHdl, const JaggedTensor &ijk, bool cumul
...
-                                   const JaggedTensor &ijk,
-                                   bool cumulative) {
-    return IjkToInvIndex<torch::kCPU>(batchHdl, ijk, cumulative);
+ijkToInvIndex(const GridBatchImpl &batchHdl, const JaggedTensor &ijk, bool cumulative) {
+    TORCH_CHECK_VALUE(
+        ijk.ldim() == 1,
+        "Expected ijk to have 1 list dimension, i.e. be a single list of coordinate values, but got",
+        ijk.ldim(),
+        "list dimensions");
+    return FVDB_DISPATCH_KERNEL_DEVICE(
```
harrism, line 95: Nit: this uses `FVDB_DISPATCH_KERNEL_DEVICE` (no PrivateUse1) while the symmetric `ijkToIndex` uses `FVDB_DISPATCH_KERNEL` (with PrivateUse1). The old code had this same asymmetry (IjkToInvIndex only had kCPU/kCUDA specializations), so this is a faithful reproduction — but if the original omission was accidental, now would be a natural time to align them.

harrism (commented):
**Disclosure: AI-assisted review** (Cursor + Claude, requested by @harrism). I verified the findings but the heavy lifting was automated analysis of the full 8k-line diff.

## Overall

This is a clean, well-executed mechanical transformation. The pattern is consistent across all 129 files, the code reads much better with self-contained ops, and the dead code cleanup is welcome. A few items below, roughly in order of importance.

## 1. Lost `ldim == 1` Precondition Checks

~15 `data.ldim() == 1` checks were removed from `GridBatch.cpp` but **cannot be re-added** in the op `.cu` files because the JaggedTensor is unwrapped via `.jdata()` into a plain `torch::Tensor` before reaching the ops. Without these, a user passing a malformed nested JaggedTensor gets a cryptic kernel error instead of a clear precondition message.

Affected functions in `GridBatch.cpp`: `max_pool`, `avg_pool`, `refine` (data check only — the mask ldim check was correctly moved to `BuildFineGridFromCoarse.cu`), `write_to_dense_cminor`, `write_to_dense_cmajor`, `sample_trilinear`, `sample_trilinear_with_grad`, `sample_bezier`, `sample_bezier_with_grad`, `splat_trilinear`, `splat_bezier`.

Recommendation: keep these specific `ldim` checks in `GridBatch.cpp` since they can only be validated before `.jdata()` extraction.

## 2. `IjkToIndex` vs `IjkToInvIndex` Dispatch Asymmetry

`ijkToIndex` uses `FVDB_DISPATCH_KERNEL` (CPU + CUDA + PrivateUse1) while the symmetric `ijkToInvIndex` uses `FVDB_DISPATCH_KERNEL_DEVICE` (CPU + CUDA only). The old code had kCPU, kCUDA, and kPrivateUse1 specializations for `IjkToIndex` but only kCPU and kCUDA for `IjkToInvIndex`, so the dispatch macros faithfully reproduce the original behavior — but if that original asymmetry was unintentional, now is a good time to fix it. (See inline comment.)

## 3. Widened Device Support (Intentional?)

Three ops changed from `FVDB_DISPATCH_KERNEL_DEVICE` to `FVDB_DISPATCH_KERNEL`, adding PrivateUse1: `ActiveGridCoords`, `SplatIntoGridTrilinear`, `UpsampleGridNearestBackward`. Worth confirming this is intentional. Copilot's review also flagged related PrivateUse1 routing concerns in `JaggedTensorIndex.cu` and `JCat0.cu` that are worth looking at in this context.

## 4. Behavioral Change in ReadFromDense / ReadIntoDense (Improvement)

Changed from asserting `is_contiguous()` (throws on non-contiguous input) to calling `.contiguous()` (makes a copy if needed). This is strictly more permissive and better UX — just noting it as an intentional contract change.

## Good Things

- Dead code cleanup: `JOffsetsFromJIdx.cu` dropped a 60-line commented-out `cub::DeviceRunLengthEncode` block and consolidated 3 per-device specializations into 1 function.
- `JOffsetsFromJIdx.cu` device guard improved from `CUDAGuard` to `OptionalDeviceGuard` (now also covers PrivateUse1).
- Dead `dLdDepthSq` / `outDepthSq` references cleaned up in VolumeRender.
- `dummy_tensor` hack eliminated from ReadIntoDense/ReadFromDense autograd (see replies to @fwilliams).
- Inject preconditions are actually *stricter* than before: two new `size(0) == totalVoxels()` checks were added.

src/fvdb/detail/autograd/AvgPoolGrid.cpp
harrism: The `.contiguous()` call was moved into `DownsampleGridAvgPool.cu` — the new `downsampleGridAvgPoolBackward` function does `torch::Tensor gradOutContig = coarseGradOut.contiguous()` before dispatching. Not lost, just relocated. (Same pattern in `DownsampleGridMaxPool.cu` for the max_pool backward.)

src/fvdb/detail/autograd/ReadIntoDense.cpp
harrism: The `dummy_tensor` (`torch::empty({0}, sparseData.options())`) existed solely to preserve `TensorOptions` (device + dtype) across the autograd context so the backward could allocate its output on the right device. The ops now allocate output tensors internally using the gradient input's `.options()`, which autograd guarantees will match the original forward input. So the hack is no longer needed.

`last_dim` and `first_dim` are now computed internally by the ops from the gradient tensor shape. `grid_size` in saved_data was actually dead data — it was saved in forward but never read in backward. All safely eliminated.

src/fvdb/detail/autograd/ReadIntoDense.cpp
harrism: Same answer as above — the `dummy_tensor` was only needed to shuttle `TensorOptions` through the autograd context. The new ops handle output allocation internally, so it's gone.

src/fvdb/detail/ops/BuildGridFromNearestVoxelsToPoints.h
harrism: Partially. `VoxelCoordTransform` is removed from the public API of `BuildGridFromNearestVoxelsToPoints`, `BuildGridFromMesh`, and `BuildGridFromPoints` — these now take decomposed `std::vector<nanovdb::Vec3d>` for `voxelSizes` and `origins` separately, and construct `VoxelCoordTransform` internally in the `.cu` files.

However, it's still exposed in `IjkForMesh.h` and `NearestIjkForPoints.h`, so the internalization is partial.

src/fvdb/detail/autograd/AvgPoolGrid.cpp
blackencino: I have no idea! It's weird that the automated process made these kinds of changes. I wonder if it sorta does a "clean up and simplify" pass as part of its outer agentic system context. I remember reading that claude's prompt has something to that effect in it.

src/fvdb/detail/autograd/ReadIntoDense.cpp
blackencino: Was it just a place to store options?

src/fvdb/GridBatch.cpp
blackencino: I think the solution to all of these is just to pass in the JaggedTensor to the op, rather than the jdata.

blackencino: Though, I also think we should redesign JaggedTensor to not have these two forms. It is a conceptual confusion minefield.

src/fvdb/detail/ops/BuildFineGridFromCoarse.h
```diff
@@ -6,29 +6,25 @@
 
 #include <fvdb/JaggedTensor.h>
 #include <fvdb/detail/GridBatchImpl.h>
-#include <fvdb/detail/TorchDeviceBuffer.h>
 
-#include <nanovdb/GridHandle.h>
 #include <nanovdb/NanoVDB.h>
 
-#include <torch/types.h>
+#include <c10/util/intrusive_ptr.h>
```
matthewdcong: Are these `intrusive_ptr` includes needed throughout? `GridBatchImpl` inherits from `CustomClassPointer` which should already include `intrusive_ptr` as a transitive header. If it's not needed, we should just remove it throughout.

src/fvdb/detail/ops/BuildGridFromIjk.h
```diff
@@ -6,19 +6,21 @@
 
 #include <fvdb/JaggedTensor.h>
 #include <fvdb/detail/GridBatchImpl.h>
-#include <fvdb/detail/TorchDeviceBuffer.h>
 
-#include <nanovdb/GridHandle.h>
 #include <nanovdb/NanoVDB.h>
 
-#include <torch/types.h>
+#include <vector>
```
matthewdcong, line 10: Also wondering if this `vector `include (and similar ones throughout this PR) are needed since ivalue.h already uses `vector.`

src/fvdb/detail/ops/IjkToInvIndex.cu
matthewdcong: The original omission was intentional. I believe we only got as far as implementing the forward pass for a sparse U-Net and switched focus before we could tackle the backwards pass.

src/fvdb/detail/ops/MarchingCubes.cu
```diff
@@ -14,6 +15,8 @@ namespace fvdb {
 namespace detail {
 namespace ops {
 
+namespace {
+
 template <typename ScalarT>
 __hostdev__ static inline int
```
matthewdcong: I don't think this needs to be static any longer because it's now in an anonymous namespace and therefore restricted to this TU.

src/fvdb/detail/ops/RayImplicitIntersection.cu
```diff
@@ -4,18 +4,20 @@
...
 
 #include <c10/cuda/CUDAException.h>
 
 namespace fvdb {
 namespace detail {
 namespace ops {
+namespace {
 
-static const int INVALID_SIGN = 10;
+const int INVALID_SIGN = 10;
```
matthewdcong: I think this can be a `constexpr` now that the `static` keyword is removed.

src/fvdb/detail/ops/ReadIntoDense.h
```diff
@@ -6,23 +6,23 @@
 
 #include <fvdb/detail/GridBatchImpl.h>
 
+#include <nanovdb/NanoVDB.h>
```
matthewdcong: Not needed since nanovdb/NanoVDB.h is include in GridBatchImpl.h

src/fvdb/detail/ops/VoxelNeighborhood.h
```diff
@@ -6,20 +6,17 @@
 
 #include <fvdb/JaggedTensor.h>
 #include <fvdb/detail/GridBatchImpl.h>
-#include <fvdb/detail/utils/Utils.h>
 
-#include <torch/types.h>
+#include <cstdint>
```
matthewdcong, line 10: Not sure if needed

src/fvdb/detail/autograd/AvgPoolGrid.cpp
blackencino: The .contiguous() call was removed from the autograd backward because the responsibility for ensuring contiguity was moved into the op itself during consolidation. If you look at downsampleGridAvgPoolBackward in DownsampleGridAvgPool.cu (line 305), the op now does:
```
torch::Tensor gradOutContig = coarseGradOut.contiguous();
```
before passing it to the kernel. The inner DownsampleGridAvgPoolBackward template then asserts coarseGradOut.is_contiguous() as a precondition (line 211). So the contiguity guarantee is preserved -- it just moved from the autograd layer into the op, which is consistent with the consolidation goal of making ops self-contained. The same pattern was applied to MaxPoolGrid's backward.

The original .contiguous() call was there because the old dispatchDownsampleGridAvgPoolBackward template function required contiguous input but didn't enforce it internally -- so the caller (autograd) had to do it. Now the op handles it.

src/fvdb/detail/autograd/ReadIntoDense.cpp
blackencino: Answered in response to next question at lines 206:209.

src/fvdb/detail/autograd/ReadIntoDense.cpp
blackencino: Both questions are about the same thing. In the original code, the backward pass manually allocated the output tensor:
```
torch::Tensor ret = torch::zeros({firstDim, lastDim}, sparseDataOpts);
```
and then passed it as a mutable output to dispatchReadFromDenseCminor, which wrote into it in-place. This required saving three things through the autograd context:

* first_dim -- the voxel count (N), i.e. sparseDataReshape.size(0)
* last_dim -- the coalesced feature dimension, i.e. sparseDataReshape.size(1)
* dummy_tensor -- a zero-element tensor saved purely to preserve TensorOptions (dtype + device), since saved_data stores IValues and can't hold a TensorOptions directly

After consolidation, the readFromDenseCminor / readFromDenseCmajor ops are self-contained -- they create their own output tensor internally (see ReadFromDense.cu line 333-334: torch::zeros({batchHdl.totalVoxels(), denseDataReshape.size(4)}, denseData.options())). So the backward no longer needs to pre-allocate the output, which means firstDim, lastDim, and the dummy tensor are all unnecessary and were removed.

The dummy tensor trick was always a workaround for PyTorch's C++ autograd saved_data not supporting TensorOptions as an IValue type. It's nice that we no longer need it.

src/fvdb/detail/ops/BuildGridFromNearestVoxelsToPoints.h
blackencino: Yes, effectively. The consolidated `buildGridFromNearestVoxelsToPoints` op now takes `voxelSizes` and `origins` directly (matching the `GridBatchImpl::createFromNeighborVoxelsToPoints` signature) and constructs the `VoxelCoordTransform` vector internally before passing it down to the lower-level `dispatchBuildGridFromNearestVoxelsToPoints` template and `nearestNeighborIJKForPoints`.

Previously, the caller (`GridBatchImpl::createFromNeighborVoxelsToPoints`) was responsible for constructing the transforms from voxel sizes and origins, then passing them to the templated dispatch function. The consolidation moved that construction into the op itself so the public-facing op signature is simpler and doesn't expose `VoxelCoordTransform`.

`VoxelCoordTransform` still appears in headers of lower-level internal ops (`NearestIjkForPoints.h`, `IjkForMesh.h`) that the consolidated ops call internally, but it's no longer part of the top-level op API surface.

src/fvdb/detail/ops/JaggedTensorIndex.cu
```diff
@@ -692,64 +692,43 @@ jaggedTensorIndexIntPrivateUse1(const JaggedTensor &jt, int64_t idxVal) {
...
+    }
+    if (end <= at::indexing::INDEX_MIN) {
+        end = 0;
+    }
+    if (jt.device().is_cuda()) {
+        c10::cuda::CUDAGuard deviceGuard(jt.device());
+        return jaggedTensorIndexSliceCuda(jt, start, end, step);
+    } else {
+        return jaggedTensorIndexSliceCpu(jt, start, end, step);
+    }
```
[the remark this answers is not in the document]
blackencino: This has been fixed in incoming PR, which reverts to old support surface for kPrivateUse1

src/fvdb/detail/ops/JCat0.cu
```diff
@@ -234,6 +232,15 @@ dispatchJCat0<torch::kCPU>(const std::vector<JaggedTensor> &vec) {
...
 }
 
+JaggedTensor
+jCat0(const std::vector<JaggedTensor> &tensors) {
+    if (tensors[0].device().is_cuda()) {
+        return jCat0CUDA(tensors);
+    } else {
+        return jCat0CPU(tensors);
+    }
+}
```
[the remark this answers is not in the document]
blackencino: This has been fixed in incoming PR which reverted to previous kPrivateUse1 support surface.

src/fvdb/detail/ops/JaggedTensorIndex.cu
```diff
@@ -692,64 +692,43 @@ jaggedTensorIndexIntPrivateUse1(const JaggedTensor &jt, int64_t idxVal) {
...
-                                                    const JaggedTensor &idx) {
-    c10::cuda::CUDAGuard deviceGuard(jt.device());
-    return jaggedTensorIndexJaggedTensorImpl(jt, idx);
+jaggedTensorIndexJaggedTensor(const JaggedTensor &jt, const JaggedTensor &idx) {
+    if (jt.device().is_cuda()) {
+        c10::cuda::CUDAGuard deviceGuard(jt.device());
+        return jaggedTensorIndexJaggedTensorImpl(jt, idx);
+    } else {
+        return jaggedTensorIndexJaggedTensorImpl(jt, idx);
+    }
```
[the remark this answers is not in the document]
blackencino: This has been resolved in incoming PR (commit) that reverts to the previous kPrivateUse1 support surface.

src/fvdb/detail/ops/BuildDilatedGrid.cu
```diff
@@ -112,4 +119,23 @@ dispatchDilateGrid<torch::kCPU>(const GridBatchImpl &gridBatch,
...
+dilateGrid(const GridBatchImpl &gridBatch, const std::vector<int64_t> &dilationAmount) {
+    TORCH_CHECK_VALUE(static_cast<int64_t>(dilationAmount.size()) == gridBatch.batchSize(),
+                      "dilationAmount should have same size as batch size, got ",
+                      dilationAmount.size(),
+                      " != ",
+                      gridBatch.batchSize());
+    TORCH_CHECK_VALUE(std::all_of(dilationAmount.begin(),
+                                  dilationAmount.end(),
+                                  [](int64_t amount) { return amount > 0; }),
+                      "dilation amount must be strictly positive.");
```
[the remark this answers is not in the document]
blackencino: Suggestion implemented, python tests expanded slightly to check that 0 dilation is allowed.

src/fvdb/GridBatch.cpp
blackencino: These precondition checks have all been added back into GridBatch.cpp because we can't easily have JaggedTensors as Autograd variables without making them into TorchScript types. This will just go away by itself when we move autograd into python in a subsequent PR. But all the ldim==1 checks are back.

src/fvdb/detail/ops/IjkToInvIndex.cu
blackencino: This is the one and only place where I let the new IjkToInvIndex keep its kPrivateUse1 support, since it is structurally identical to IjkToIndex, and may as well support them both.

src/fvdb/detail/ops/BuildFineGridFromCoarse.h
blackencino: I feel like IWYU says we should keep it there.

src/fvdb/detail/ops/BuildGridFromIjk.h
blackencino: I try to adhere to IWYU coarsely, so I think this is ok. We explicitly mention the type in the API, it should have its antecedent.

src/fvdb/detail/ops/MarchingCubes.cu
blackencino: Fixed all of these, removed the static.

src/fvdb/detail/ops/RayImplicitIntersection.cu
blackencino: Fixed

src/fvdb/detail/ops/ReadIntoDense.h
blackencino: Removed this, because although we do reference nanovdb::Coord, that's transitive enough that we can not be pedantic. There's no Grid or Tree type in the public API, so this makes sense.

src/fvdb/detail/ops/VoxelNeighborhood.h
blackencino: the API mentions int32_t, so it is IWYU practice to have cstdint (and I usually also add cstddef). The header guards mean it isn't parsed more than once, and the std library files are hit so many times during compilation, they're in whatever filesystem caches they need to be.

matthewdcong (approved):
Looks good, thanks for the changes.

PR #499, opened 2026-03-04
Improve/optimize mGPU scaling via batched prefetching and sorting changes

matthewdcong:
1. Coalesce consecutive `cudaMemPrefetchAsync` calls into a single `cudaMemPrefetchBatchAsync` call in order to amortize OS overhead
2. Decouple the mGPU radix sort into trivially parallel per-camera radix sorts. Effectively, the data corresponding to each batch can be sorted independently on each GPU instead of one large mGPU radix sort.
3. Switch from `DeviceRadixSort` to `DeviceMergeSort` for mGPU. In mGPU, the performance advantage of radix sort is outweighed by the additional cost of allocating separate input and output buffers.


src/fvdb/detail/ops/gsplat/FusedSSIM.cu
```diff
@@ -609,10 +654,19 @@ fusedSSIMPrivateUse1(
     auto img1_ = img1.contiguous();
     auto img2_ = img2.contiguous();
 
-    const auto globalBlockCount = ((W + BLOCK_X - 1) / BLOCK_X) * ((H + BLOCK_Y - 1) / BLOCK_Y) * B;
+    std::vector<cudaEvent_t> events(c10::cuda::device_count());
     for (const auto deviceId: c10::irange(c10::cuda::device_count())) {
         C10_CUDA_CHECK(cudaSetDevice(deviceId));
         auto stream = c10::cuda::getCurrentCUDAStream(deviceId);
+        C10_CUDA_CHECK(cudaEventCreate(&events[deviceId], cudaEventDisableTiming));
```
harrism, line 661: **🎵 note (non-blocking):** ‏In case you are interested, a while back I wrote benchmarks of CUDA events. When I measured it, create (no timing) + record cost about a millisecond. Using a simple event pool can be a lot cheaper.
https://github.com/harrism/cuda_event_benchmark

src/fvdb/detail/ops/gsplat/GaussianRasterizeBackward.cu
```diff
@@ -1221,10 +1222,49 @@ callRasterizeBackwardPrivateUse1(
...
+        auto stream = c10::cuda::getStreamFromPool(false, deviceId);
+        C10_CUDA_CHECK(cudaStreamWaitEvent(stream, events[deviceId]));
 
         uint32_t deviceTileOffset, deviceTileCount;
         std::tie(deviceTileOffset, deviceTileCount) = deviceChunk(tileCount, deviceId);
+        uint32_t cameraOffset                       = deviceTileOffset / (tileCount / C);
+        uint32_t cameraCount =
+            cuda::ceil_div(deviceTileOffset + deviceTileCount, tileCount / C) - cameraOffset;
+        if (deviceTileCount) {
+            std::vector<torch::Tensor> tensors = {// means2d,
```
harrism: **❓ question:** ‏Why not fetch means2d, conics, features, etc?

src/fvdb/detail/ops/gsplat/GaussianUtils.h
```diff
@@ -0,0 +1,31 @@
...
+                            uint32_t cameraOffset,
+                            uint32_t cameraCount,
+                            int deviceId,
+                            cudaStream_t stream);
+
+void perCameraPrefetchBatchAsync(const torch::TensorList &tensors,
+                                 uint32_t cameraOffset,
+                                 uint32_t cameraCount,
+                                 int deviceId,
+                                 cudaStream_t stream);
```
harrism, line 30: **☑️ todo:** ‏Docs.

harrism (approved):
Looks good. I'd like you to document the two new utility functions before merging.

src/fvdb/detail/ops/gsplat/FusedSSIM.cu
matthewdcong: Yup, those benchmarks have been super helpful (and taught me the importance of timing vs non-timing performance wise). An event pool has been on my to-do list for a while, but the question is where to store it (whether in PyTorch or fVDB, or just have two), and I haven't come up with a good answer yet.

src/fvdb/detail/ops/gsplat/GaussianRasterizeBackward.cu
matthewdcong: Those pages are already resident on the GPU from the forward pass. Prefetching them is idempotent, but it does incur a noticeable overhead. Avoiding prefetching tensors already resident from the forward pass throughout the code gives a 10-15% speedup on an 8 GPU reconstruction.

src/fvdb/detail/ops/gsplat/GaussianUtils.h
matthewdcong: good suggestion, added

src/fvdb/detail/ops/gsplat/GaussianRasterizeBackward.cu
harrism: Maybe add a comment?

PR #508, opened 2026-03-05
Added PredGatherIGemm sparse conv backend.

blackencino:
# PredGatherIGemm: Alternative Sparse Convolution Backend

## Summary

This PR adds a new sparse convolution backend -- **PredGatherIGemm** -- that uses
CUTLASS/CuTe implicit-GEMM (IGEMM) with predicated `cp.async` gather loads on
SM80+ (Ampere and later) GPUs. It processes one output NanoVDB leaf node per CTA,
using TF32 tensor-core arithmetic for the computation.

The backend is integrated into the `ConvolutionPlan` framework as a selectable
backend (`expert_config={"backend": "pred_gather_igemm"}`), with the existing
GatherScatterDefault backend remaining the default.

## Constraints

The PredGatherIGemm backend is intentionally limited in scope compared to the
default GatherScatterDefault backend:

- **CUDA only**, requires SM80+ (Ampere or later)
- **Float32 only** (internally promoted to TF32)
- **Forward pass only** -- no transpose, no analytical backward (backward falls
  back to GatherScatterDefault when used via autograd)
- **Uniform kernel sizes** only: 3, 5, or 7 (x=y=z)
- **Uniform strides** only: 1 or 2 (x=y=z)
- **Channel counts** must be multiples of 32
- **Batch size 1** only

Kernel size and stride are dispatched at compile time using the project's
`dispatch` framework, giving 6 total template instantiations.

## Performance Characteristics

Benchmarked on SM120 with Cin=64, Cout=128, kernel 3x3x3, stride 1:

| Scenario | PredGatherIGemm | GS + topology | GS (topology cached) |
|---|---|---|---|
| 1M dense (75% leaf occ) | **5.2 ms** | 45.8 ms | 31.2 ms |
| 2M dense (75% leaf occ) | **10.2 ms** | 89.1 ms | 64.9 ms |
| 4M sparse (25% leaf occ) | 32.0 ms | 21.1 ms | **14.8 ms** |
| 8M sparse (10% leaf occ) | 43.6 ms | 8.1 ms | **5.0 ms** |

The IGEMM backend is significantly faster for **dense or near-dense** grids
(high leaf-node occupancy), where its one-leaf-per-CTA approach keeps the GPU
fully occupied. At low occupancy the per-CTA work becomes sparse and the
GatherScatterDefault backend -- which operates on compacted index pairs -- wins
decisively.

### New files

- `src/fvdb/detail/ops/convolution/PredGatherIGemm.h` -- public header
- `src/fvdb/detail/ops/convolution/PredGatherIGemm.cu` -- CUTLASS IGEMM kernel,
  CuTe layouts, dispatch table, and entry point
- `src/tests/PredGatherIGemmTest.cu` -- C++ gtests: correctness validation
  against GatherScatterDefault across all 6 kernel/stride combinations, plus
  speed comparison benchmarks
- `tests/unit/test_conv_pred_gather_igemm.py` -- Python tests: forward-pass
  validation against dense PyTorch conv3d ground truth and cross-backend
  comparison with GatherScatterDefault

### Modified files

- `src/fvdb/GridBatch.h` / `src/fvdb/GridBatch.cpp` -- added static
  `predGatherIGemmConv` method
- `src/python/Bindings.cpp` -- pybind11 binding for `pred_gather_igemm_conv`
- `fvdb/_fvdb_cpp.pyi` -- type stub for the new binding
- `fvdb/convolution_plan.py` -- `_PredGatherIGemmBackend`, autograd wrapper
  (`_PredGatherIGemmConvFn`), backend selection logic in `_build_backend`
- `src/CMakeLists.txt` / `src/tests/CMakeLists.txt` -- added new source and test
  files to the build

## Test Plan

- `ninja PredGatherIGemmTest && ./src/tests/PredGatherIGemmTest` --
  runs the C++ gtest suite (correctness + benchmarks)
- `python -m pytest tests/unit/test_conv_pred_gather_igemm.py -v` --
  runs the Python test suite (forward-only, TF32-tolerant comparisons against
  dense ground truth and GatherScatterDefault)

### DIFF
--- a/src/fvdb/detail/ops/convolution/PredGatherIGemm.cu
+++ b/src/fvdb/detail/ops/convolution/PredGatherIGemm.cu
@@ -330,13 +330,13 @@ namespace gather_util {
 
 using namespace cute;
 
-template <class Index> struct IndexedGather {
+template <class Index, int Offset = 0> struct IndexedGather {
     CUTE_HOST_DEVICE constexpr IndexedGather(Index const *indices = {}) : indices_(indices) {}
 
     template <typename I>
     CUTE_HOST_DEVICE constexpr Index
     operator()(I i) const {
-        return indices_[i];
+        return indices_[i] + Index(Offset);
     }
 
     CUTE_HOST_DEVICE friend void
@@ -444,13 +444,15 @@ using namespace cute;
 // Templatized convolution geometry
 // ---------------------------------------------------------------------------
 
-template <int T_, int R_, int S_, int STx_, int STy_, int STz_> struct Geometry {
+template <int T_, int R_, int S_, int STx_, int STy_, int STz_, int KoutDenom_> struct Geometry {
     static constexpr int T   = T_;
     static constexpr int R   = R_;
     static constexpr int S   = S_;
     static constexpr int STx = STx_;
     static constexpr int STy = STy_;
     static constexpr int STz = STz_;
+
+    static_assert(KoutDenom_ == 32 || KoutDenom_ == 128, "KoutDenom must be 32 or 128");
 
     static constexpr int Z = 4;
     static constexpr int P = 2;
@@ -481,7 +483,7 @@ template <int T_, int R_, int S_, int STx_, int STy_, int STz_> struct Geometry 
     }
 
     static constexpr int TC = 32;
-    static constexpr int TK = 32;
+    static constexpr int TK = KoutDenom_;
 
     static constexpr int ZZ = 1;
     static constexpr int PP = 2;
@@ -516,9 +518,6 @@ template <int T_, int R_, int S_, int STx_, int STy_, int STz_> struct Geometry 
         return CHx * CHy * CHz;
     }
 };
-
-// Alias for the currently supported configuration
-using Geometry3x3x3 = Geometry<3, 3, 3, 1, 1, 1>;
 
 // ---------------------------------------------------------------------------
 // CuTe layout construction for the IGEMM
@@ -562,10 +561,12 @@ template <class SettingsT> struct Layouts {
                             STz * EG),
                 make_stride(EC, CHy * CHz * EG, CHz * EG, EG)));
 
-        auto xformed_act_gather_outer = make_layout(
-            make_shape(_1{}, _1{}),
-            make_stride(gather_util::CustomStride{gather_util::IndexedGather{gather_idx_buf}, C},
-                        _1{}));
+        auto xformed_act_gather_outer =
+            make_layout(make_shape(_1{}, _1{}),
+                        make_stride(
+                            gather_util::CustomStride{
+                                gather_util::IndexedGather<uint64_t, -1>{gather_idx_buf}, C},
+                            _1{}));
 
         return composition(
             xformed_act_gather_outer, make_arithmetic_tuple(_0{}, _0{}), xformed_act_logical_inner);
@@ -635,10 +636,12 @@ template <class SettingsT> struct Layouts {
                                                 _64{} * ES,
                                                 _8{} * ES,
                                                 ES)));
-        auto xformed_out_scatter_outer = make_layout(
-            make_shape(_1{}, _1{}),
-            make_stride(gather_util::CustomStride{gather_util::IndexedGather{scatter_idx_buf}, K},
-                        _1{}));
+        auto xformed_out_scatter_outer =
+            make_layout(make_shape(_1{}, _1{}),
+                        make_stride(
+                            gather_util::CustomStride{
+                                gather_util::IndexedGather<uint64_t, -1>{scatter_idx_buf}, K},
+                            _1{}));
         return composition(xformed_out_scatter_outer,
                            make_arithmetic_tuple(_0{}, _0{}),
                            xformed_out_logical_inner);
@@ -952,6 +955,7 @@ namespace ops {
 
 enum class conv_kernel_size : int { k3 = 3, k5 = 5, k7 = 7 };
 enum class conv_stride : int { s1 = 1, s2 = 2 };
+enum class conv_kout_denom : int { K32 = 32, K128 = 128 };
 
 } // namespace ops
 } // namespace detail
@@ -973,6 +977,13 @@ template <> struct type_label<fvdb::detail::ops::conv_stride> {
     }
 };
 
+template <> struct type_label<fvdb::detail::ops::conv_kout_denom> {
+    static consteval auto
+    value() {
+        return fixed_label("conv.kout_denom");
+    }
+};
+
 } // namespace dispatch
 
 // ============================================================================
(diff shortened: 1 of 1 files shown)


src/fvdb/detail/ops/convolution/PredGatherIGemm.cu
```diff
@@ -0,0 +1,1121 @@
...
+        const int64_t C    = features.size(1);
+        const int64_t K    = weights.size(0);
+
+        const int64_t N_out            = output_grid.totalVoxels();
+        const uint32_t outputLeafCount = output_grid.numLeavesAt(0);
+
+        auto filter_igemm = weights.permute({2, 3, 4, 0, 1}).contiguous();
+
+        auto opts         = torch::dtype(torch::kFloat32).device(features.device());
+        auto padded_input = torch::zeros({N_in + 1, C}, opts);
```
sifakis: This operation should be able to work directly on the `features` tensor, without the extra padding; see the kernel launch below for details (and caveats)

src/fvdb/detail/ops/convolution/PredGatherIGemm.cu
```diff
@@ -0,0 +1,1121 @@
...
+        padded_input.slice(0, 1).copy_(features);
+
+        auto padded_output = torch::zeros({N_out + 1, K}, opts);
```
sifakis: Similarly, we should be able to avoid padding

src/fvdb/detail/ops/convolution/PredGatherIGemm.cu
```diff
@@ -0,0 +1,1121 @@
...
+        cudaFuncSetAttribute(kernelEntrypoint<ConvOp, nanovdb::ValueOnIndex, decltype(tFilter)>,
+                             cudaFuncAttributeMaxDynamicSharedMemorySize,
+                             smem_size);
+
+        kernelEntrypoint<ConvOp, nanovdb::ValueOnIndex, decltype(tFilter)>
+            <<<outputLeafCount, ConvOp::MaxThreadsPerBlock, smem_size, stream>>>(
+                tFilter,
+                nanoInputGrid,
+                nanoOutputGrid,
+                padded_input.data_ptr<float>(),
```
sifakis: The IGEMM kernel (per NanoVDB convention) treats sidecars as being 1-based, in the sense that the first active voxel will be at the entry(row) = 1, whereas entry/row 0 is the background value.
However, the kernel will never actually touch the 0-th entry (this is where the Predication code kicks in), so we can just offset the data base pointer to adjust for this. 
This is presuming that the activation tensor is stored in voxel-major layout (the stride from one voxel to the next is equal to C). 

In this case, we should be able to use 
`unpadded_input.data_ptr<float>()-C`
and
`unpadded_output.data_ptr<float>()-C`
in this call, while getting the same result (and avoiding the copies for purpose of padding)

src/fvdb/detail/ops/convolution/PredGatherIGemm.cu
```diff
@@ -0,0 +1,1121 @@
...
+    Dy() const {
+        return dy;
+    }
+    __hostdev__ int
+    Dz() const {
+        return dz;
+    }
+
+    static constexpr int TC = 32;
+    static constexpr int TK = 32;
```
sifakis: Mental note for future specialization ... TK=32/64/128 are supported and have been tested in this context. The TK = 128 case can provide a 40-50% acceleration over the TK=32 case. It might be premature to specialize now, but we may consider having a template argument TK_ for this class, and selectively launch the best specialization based on whether the (run-time) output channel K is a multiple of 32/64 or 128 (picking the highest applicable alignment/multiplier)

tests/unit/test_conv_pred_gather_igemm.py
```diff
@@ -0,0 +1,323 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+"""
+Test the PredGatherIGemm (CUTLASS IGEMM) sparse convolution backend.
+
+The PredGatherIGemm backend has strict constraints compared to the default:
+  - CUDA only (SM80+)
+  - float32 only (uses TF32 tensor-core arithmetic internally)
+  - Forward pass only (no backward, no transpose)
+  - Uniform kernel sizes: 3, 5, 7
```
sifakis, line 11: Non-uniform sizes should actually work (they have been tested in the standalone development harness).
There might be an upper limit of ~9 in the kernel size (roughly, in terms of geom.average) due to the impact to SMEM utilization.

tests/unit/test_conv_pred_gather_igemm.py
```diff
@@ -0,0 +1,323 @@
...
+  - Uniform strides: 1, 2
```
sifakis, line 12: Nonuniform strides (but only chosen from 1 or 2) should also work, although this has not been thoroughly tested in standalone benchmarks.

blackencino: I chose uniform so that the instantiation burden was minimized. Since this is a "special case" backend, we can afford to limit it to the cases where it is most likely to be used.

tests/unit/test_conv_pred_gather_igemm.py
blackencino: Yep, I just limited it to uniform sizes to minimize the instantiation burden.

src/fvdb/detail/ops/convolution/PredGatherIGemm.cu
blackencino: I'll add a TK=128 specialization.

blackencino: Done in incoming commit to this PR. It does indeed speed things up quite a bit.

blackencino:
```
SpeedComparison (Cin=64, Cout=128, kernel 3x3x3):
  [1M dense] N_in=786432  N_out=786432  leaves=2048  leaf_occ=75%
    PredGatherIGemm: 5.20076 ms  |  GS+topo: 45.9111 ms  |  GS only: 31.2607 ms
  [2M dense] N_in=1572864  N_out=1572864  leaves=4096  leaf_occ=75%
    PredGatherIGemm: 10.2267 ms  |  GS+topo: 88.6112 ms  |  GS only: 64.1556 ms
  [4M sparse-25] N_in=1048576  N_out=1048576  leaves=8192  leaf_occ=25%
    PredGatherIGemm: 32.2555 ms  |  GS+topo: 21.0543 ms  |  GS only: 14.8498 ms
  [8M sparse-10] N_in=838860  N_out=838860  leaves=16384  leaf_occ=9.99999%
    PredGatherIGemm: 43.6757 ms  |  GS+topo: 8.00768 ms  |  GS only: 4.96998 ms
    ```

src/fvdb/detail/ops/convolution/PredGatherIGemm.cu
blackencino: This is fixed by modifing IndexGather with an offset, so we never bother with a padded copy.

src/fvdb/detail/ops/convolution/PredGatherIGemm.cu
blackencino: Fixed via an offset in IndexeGather, avoiding padded copy.

src/fvdb/detail/ops/convolution/PredGatherIGemm.cu
blackencino: Fixed via an offset in IndexedGather, padded copy removed.

PR #509, opened 2026-03-05
Fix CCCL version check macro

matthewdcong:
This macro incorrectly compiles to false with CUDA 13 due to the `&&` w/ the minor version check.


blackencino (approved):
LGTM!

PR #512, opened 2026-03-05
Add release branching docs and automation scripts

harrism:
## Summary

- Add `docs/release-process.md` documenting a simplified OneFlow-based release branching model with burndown, code freeze, and release phases
- Add `devtools/start-release.sh` to automate burndown start: creates release branch, bumps versions, opens merge PR via `gh` CLI
- Add `devtools/finish-release.sh` to automate release day: tags the release, merges PR, creates GitHub Release
- Add `devtools/test-release-scripts.sh` with 23 end-to-end assertions against a disposable temp git repo
- Update `CONTRIBUTING.md` to reference the release process docs
- Scripts detect target repo from working directory (`git rev-parse --show-toplevel`) so they work with both fvdb-core and fvdb-reality-capture

## Test plan

- [x] `./devtools/test-release-scripts.sh` passes (23/23 assertions)
- [x] `./devtools/start-release.sh 0.4.0 --dry-run` prints expected plan
- [x] `./devtools/finish-release.sh 0.4.0 --dry-run` prints expected plan
- [x] Review `docs/release-process.md` for accuracy and completeness

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -505,7 +505,7 @@ jobs:
     if: ${{ !cancelled() && needs.start-validation-runner.result == 'success' }}
     runs-on: val-${{ matrix.python-version }}-pt${{ matrix.torch-version }}-cu${{ matrix.cuda-version }}-${{ github.run_id }}
     container:
-      image: nvidia/cuda:12.8.1-cudnn-runtime-ubuntu22.04
+      image: nvidia/cuda:${{ matrix.cuda-image-tag }}-cudnn-runtime-ubuntu22.04
       env:
         PYTHONPATH: ""
       options: --rm


docs/release-process.md
```diff
@@ -0,0 +1,166 @@
...
+Use `--dry-run` to preview without making changes.
+
+### During Burndown
+
+Once the release branch exists, PR targeting rules change:
+
+- **New PRs opened during burndown** target `main` by default and will ship in
+  the *following* release. If a PR is needed in the current release, it must
+  target the release branch directly and requires maintainer approval.
+- **PRs opened before burndown** that have not yet merged can still land on
```
harrism: On reflection, I don't think this is a good approach. We should make burndown *planning* an active task, which means we should decide which open PRs are going into the release and retarget them to the release branch, rather than merging into main and cherry-picking them back.  If they are needed immediately on main, they can then be cherry-picked back to main as well. This may cause a minor conflict when the release branch is merged back in but the content will be nearly identical and easy to resolve, as stated below.

docs/release-process.md
```diff
@@ -0,0 +1,168 @@
...
+and are merged back into `main` at release time.
+
+```
+main:    ──A──B──C─────────────F──G──H── ...
+               \               /
+release/v0.4:   D──E──────────T (tag v0.4.0)
+```
+
+- **A, B, C**: normal development on `main`
+- **C**: burndown begins; `release/v0.4` branches off, `main` version bumped
```
swahtz: B is where burndown begins I believe.

harrism: I agree.

harrism: Good catch -- fixed in 334e70f. The branch forks from B (current HEAD of main), then C is the version bump commit on main. Also added a commit D on main between the branch point and the merge to show a new feature landing during burndown that is not included in the release.

swahtz (approved):
lgtm

PR #533, opened 2026-03-11
Restore fix for dLossDQuat missing warpSum

matthewdcong:
Originally added in #435 but accidentally clobbered in #444


swahtz:
Just want to catchup with @harrism about whether we merge this into the release/v0.4 branch instead.

swahtz (approved):
Thanks for catching this!

PR #545, opened 2026-03-13
publish: dual S3 + PyPI publish on release (with GPU tests)

swahtz:
On GitHub Release events, upload the full wheel matrix to the S3 production index (simple/ prefix) AND publish an sdist to PyPI. Previously these paths were mutually exclusive.

- Add `pypi` and `release` routing flags to pr-flags job
- Handle `release` event: set s3=true, pypi=true, release=true
- Rewrite publish-dist to build and upload an sdist instead of downloading pre-built wheels, matching what PyPI hosts today
- Extend auditwheel repair condition to also run for release events
- Enable GPU validation (smoke test + unit tests) on releases
- Add `release` as a workflow_dispatch publish_target option so a failed release can be re-triggered manually
- Replace github.event_name checks with the release flag so manual dispatch with publish_target=release correctly targets production S3 (simple/) and PyPI

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -88,6 +88,7 @@ jobs:
         with:
           sparse-checkout: env/test_environment.yml
           sparse-checkout-cone-mode: false
+          ref: ${{ inputs.branch || github.ref }}
       - name: Parse test environment versions
         id: test-env
         run: |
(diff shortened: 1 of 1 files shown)


harrism (approved):
Approve, but the combinations of release targets and options are a getting more complicated, do we need more documentation?

PR #546, opened 2026-03-13
Improve mGPU partitioning for SH operators

matthewdcong:
Prior to this change, mGPU spherical harmonics calculations were partitioned across the camera-Gaussian pairs, i.e. C * N. This means that each GPU would need to write to each of the N derivatives associated with the spherical harmonics coefficients which necessitated the use of system-scope atomics across GPUs therefore bottlenecking the interconnect.

Instead, we partition the calculations across Gaussians. Each mGPU processes a distinct segment of Gaussians for all cameras which means that atomics only need to be device scoped and never need to traverse the interconnect. Furthermore, each mGPU only needs to access the memory that's physically resident on itself further speeding up calculations.

Naturally, the forward and backward passes should match in order to prevent ping-ponging of shared tensors between the forward and backwards pass. On an 8x A100 reconstruction, this improves performance by about 5%.


blackencino (approved):
I'd eventually love to see this stuff work with the dispatch helpers that were created, particularly with the gpuAtomicAdd replacements here. However, that's clearly an optimization for a later day.

PR #547, opened 2026-03-13
Improve mGPU partitioning for Gaussian projection operators

matthewdcong:
Similar to #546 . Prior to this change, mGPU projection calculations were partitioned across the camera-Gaussian pairs, i.e. C * N. This means that each GPU would need to write to each of the N derivatives associated with the projection outputs which necessitated the use of system-scope atomics across GPUs therefore bottlenecking the interconnect.

Instead, we partition the calculations across Gaussians. Each mGPU processes a distinct segment of Gaussians for all cameras which means that atomics only need to be device scoped (with the exception of the camera pose derivative).

In addition, we load the cameras into shared memory for the backwards pass following the example of the forwards pass. Overall, these changes provide a >5% speedup on 2x RTX 3090s.


blackencino (approved):
As with #546, I'd love to see this be able to take advantage of some of the nice additions in the dispatch framework, but that's an optimization for a later day. This looks great.

PR #554, opened 2026-03-18
Shared memory optimizations for Gaussian rasterization

matthewdcong:
1. Forward rasterization does not currently store features in shared memory. As the problem size becomes larger (more intersections with each Gaussian), the cost of an unconditional global load is outweighed by the shared memory reuse.
2. In addition, we cull loads for Gaussians with an opacity less than the threshold necessary for a Gaussian to be valid in the volume rendering pass. This optimization applies to the forward and backwards pass.

In profiling, this reduces a 17m 20s single-GPU reconstruction to 16m and 48s, leading to an approximately >3% speedup.

### DIFF
--- a/fvdb/_gaussian_autograd.py
+++ b/fvdb/_gaussian_autograd.py
@@ -384,6 +384,7 @@ class _RasterizeScreenSpaceGaussiansFn(torch.autograd.Function):
             tile_size,
             tile_offsets,
             tile_gaussian_ids,
+            -1,
             backgrounds,
             masks,
         )
@@ -530,6 +531,7 @@ class _RasterizeScreenSpaceGaussiansSparseFn(torch.autograd.Function):
             tile_pixel_mask,
             tile_pixel_cumsum,
             pixel_map,
+            -1,
             backgrounds,
             masks,
         )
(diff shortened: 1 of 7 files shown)


src/fvdb/detail/ops/gsplat/GaussianRasterizeForward.cu
```diff
@@ -548,7 +578,7 @@ launchRasterizeForwardKernels(
                                                                                   pixelMap);
 
             // Thread blocks cooperatively cache a tile of Gaussians in shared memory
-            const uint32_t sharedMem = getSharedMemRequirements<ScalarType>(tileSize);
+            const uint32_t sharedMem = getSharedMemRequirements<ScalarType>(NUM_CHANNELS, tileSize);
```
harrism: **🚩 issue:** ‏Shouldn't this be NUM_SHARED_CHANNELS? Also what happens if the number of channels is too large to fit all the features in shared memory?

harrism (changes requested):
One concern.

src/fvdb/detail/ops/gsplat/GaussianRasterizeForward.cu
matthewdcong: There's no `NUM_SHARED_CHANNELS` in the forwards pass (it's just `NUM_CHANNELS`) because there isn't chunking implemented. So assuming unlimited shared memory, this would be correct as written.

That being said, the lack of chunking is likely why the tests are failing for large feature depths so I'll have to add that.

harrism (approved):

PR #555, opened 2026-03-18
Reduce shared memory usage in pinhole projection by re-arranging blocks

matthewdcong:
Currently, each block loads all C cameras. Furthermore, the number of blocks scales with C and therefore our shared memory bandwidth scales as O(C^2). To address this, we parameterize the blocks with the cameras in the Y direction ensuring that each block only requires one camera. This enables us to load just the single camera into shared memory and have aggregate shared memory bandwidth that scales with C.

Note that this optimization doesn't apply to the non-pinhole case. In that scenario, the rasterization pass needs access to all the cameras since it is parameterized by tiles though it's possible this could be changed at a future point with more extensive refactoring.

### DIFF
--- a/src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
+++ b/src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
@@ -194,28 +194,41 @@ template <typename T> struct PerspectiveCamera {
     T nearPlane         = T(kBackwardProjectionNearPlane);
     T farPlane          = T(kBackwardProjectionFarPlane);
 
-    Mat3 *__restrict__ projectionMatsShared        = nullptr; // [C,3,3], optional
-    Mat3 *__restrict__ worldToCamRotMatsShared     = nullptr; // [C,3,3], optional
-    Vec3 *__restrict__ worldToCamTranslationShared = nullptr; // [C,3], optional
+    Mat3 *__restrict__ projectionMatShared         = nullptr; // [3,3], optional
+    Mat3 *__restrict__ worldToCamRotMatShared      = nullptr; // [3,3], optional
+    Vec3 *__restrict__ worldToCamTranslationShared = nullptr; // [3], optional
 
   public:
     /// @brief Returns dynamic shared-memory bytes required to cache camera matrices.
-    inline __host__ __device__ size_t
+    constexpr __host__ __device__ size_t
     numSharedMemBytes() const {
-        return static_cast<size_t>(2 * numCameras) * sizeof(Mat3) +
-               static_cast<size_t>(numCameras) * sizeof(Vec3);
+        return 2 * sizeof(Mat3) + sizeof(Vec3);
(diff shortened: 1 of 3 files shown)


src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
```diff
@@ -265,8 +278,8 @@ template <typename T> struct PerspectiveCamera {
         using Vec2   = nanovdb::math::Vec2<T>;
 
         T fx, fy, cx, cy;
-        if (projectionMatsShared != nullptr) {
-            const Mat3 &K = projectionMatsShared[cid];
+        if (projectionMatShared != nullptr) {
+            const Mat3 &K = *projectionMatShared;
```
fwilliams, line 286: If you want, this could be a nice opportunity to rename K to projectionMatrix or something more in line with the conventions of the codebase.

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
```diff
@@ -194,28 +194,41 @@ template <typename T> struct PerspectiveCamera {
...
     inline __device__ void
-    loadSharedMemory(void *sharedMemory) {
-        const int64_t C             = numCameras;
-        projectionMatsShared        = reinterpret_cast<Mat3 *>(sharedMemory);
-        worldToCamRotMatsShared     = projectionMatsShared + C;
-        worldToCamTranslationShared = reinterpret_cast<Vec3 *>(worldToCamRotMatsShared + C);
-        copyMat3Accessor<T>(C, projectionMatsShared, projectionMatricesAcc);
-        copyMat3Accessor<T>(C, worldToCamRotMatsShared, worldToCamMatricesAcc);
-        copyWorldToCamTranslation<T>(C, worldToCamTranslationShared, worldToCamMatricesAcc);
+    loadSharedMemory(const int64_t cid, void *sharedMemory) {
```
fwilliams, line 214: Can we add a comment here or somewhere else on the axis ordering in shared memory with an  explanation as to why we chose that ordering. We missed this problem before, so it might be helpful to call out the problem and solution for future readers of the code.

fwilliams (approved):
Looks good and great catch! I have a few small comments but none that are blocking. Approved!

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
matthewdcong: That's a good suggestion, but we already have `projectionMatrix` as a function in the same class (and other variants elsewhere) so I'm just going to keep it as is for now since it's just a temp local variable.

src/fvdb/detail/ops/gsplat/GaussianCameras.cuh
matthewdcong: done

PR #560, opened 2026-03-25
Fuse computeGradientState into projectionBackwardsKernel

matthewdcong:
`computeGradientState` calculates quantities that inform whether or not Gaussians need to be refined. However, it requires reading `dLossDMeans2d` and `radii` from global memory. The preceeding kernel, `projectionBackwardKernel` already calculates the former and reads the latter making it a good candidate for fusion.

In order to account for the differences in launch parameters, we need to use atomic adds and max functions. However, these have relatively small contention (equal to C the number of cameras) so the performance hit is minimal compared to the cost of re-reading from global memory. On a 2x GPU reconstruction, the fused `projectionBackwardsKernel` is 1.25x faster than the non-fused `projectionBackwardsKernel` followed by `computeGradientState`.


fwilliams (approved):
Looks good and makes the code simpler too which is nice.

PR #563, opened 2026-03-25
Hotfix release process improvements

swahtz:
This pull request introduces comprehensive support for hotfix releases (PATCH > 0) in the release tooling, including both `start-release.sh` and `finish-release.sh` scripts. It updates the release workflow to handle hotfixes differently from minor releases, ensures proper versioning and branch checks, and adds automated documentation version syncing via a new GitHub Actions workflow. Extensive tests have been added to validate hotfix scenarios.

**Hotfix Release Support in Release Scripts:**

* `devtools/start-release.sh`, `devtools/finish-release.sh`: Added logic to detect hotfix releases (PATCH > 0), ensuring that hotfixes are prepared on existing release branches, require the branch to match the previous tag, and skip creating adopt branches. The scripts now provide clear instructions and enforce correct branch/tag states for hotfixes. [[1]](diffhunk://#diff-e4f587213e30a3c92eb2ffd1236ee2783b3c94e7829018900dfc6b4bdfdd59d6R6-R10) [[2]](diffhunk://#diff-e4f587213e30a3c92eb2ffd1236ee2783b3c94e7829018900dfc6b4bdfdd59d6L27-R34) [[3]](diffhunk://#diff-e4f587213e30a3c92eb2ffd1236ee2783b3c94e7829018900dfc6b4bdfdd59d6R99-R114) [[4]](diffhunk://#diff-e4f587213e30a3c92eb2ffd1236ee2783b3c94e7829018900dfc6b4bdfdd59d6R141-R223) [[5]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73R6-R11) [[6]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73L26-R37) [[7]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73L107-R133) [[8]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73R158-R163) [[9]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73L187-R208) [[10]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73L199-R228) [[11]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73R245-R257) [[12]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73R286-R295) [[13]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73L274-R315)

* [`devtools/finish-release.sh`](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73L107-R133): For hotfix releases, skips adopt branch creation and updates PR closing comments to clarify that changes are not merged back into main. Also triggers the new doc version sync workflow after creating a GitHub Release. [[1]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73L107-R133) [[2]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73R158-R163) [[3]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73L187-R208) [[4]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73L199-R228) [[5]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73R245-R257) [[6]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73R286-R295) [[7]](diffhunk://#diff-6736c27a9c15d3054c2158ee99fba6596fd45659941928cc0eccc4860091ac73L274-R315)

**Automated Documentation Version Sync:**

* [`.github/workflows/sync-doc-version.yml`](diffhunk://#diff-250abcf09d61581e29e71e3ed9e981c2705f72fce19224a5d0e9efab202ebb4cR1-R70): Added a new workflow to automatically update the stable documentation version in `docs/conf.py` to match the latest GitHub release. This workflow is triggered on a schedule or manually, and is also invoked by the release scripts.

* [`.github/workflows/docs.yml`](diffhunk://#diff-9cf2000c53760d837a449f874e53f792819108d3a4bf346336d0f7d082deae2cR4-R6): Modified to trigger on pushes to `main` that affect files in the `docs/` directory, ensuring documentation is rebuilt when the stable version is updated.

**Testing and Validation:**

* [`devtools/test-release-scripts.sh`](diffhunk://#diff-211d3295395fdceec176c31214f4cbf6c6204f969e2787fdbf5e158d22c2657cR63-R71): Added extensive tests for hotfix release scenarios, including dry runs, validation of branch/tag state, correct version bumping, and ensuring no adopt branch is created for hotfixes. [[1]](diffhunk://#diff-211d3295395fdceec176c31214f4cbf6c6204f969e2787fdbf5e158d22c2657cR63-R71) [[2]](diffhunk://#diff-211d3295395fdceec176c31214f4cbf6c6204f969e2787fdbf5e158d22c2657cR348-R483)

These changes make the release process more robust and automate the synchronization of documentation with releases, while ensuring hotfixes are handled safely and correctly.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -98,6 +98,13 @@ algorithms for 3D reconstruction from sensor data.
    api/nn
    api/utils
 
+.. toctree::
+   :maxdepth: 1
+   :caption: Development Process
+
+   release-process
+   markdown/clangd
+
 .. raw:: html
 
    <hr>


docs/release-process.md
```diff
@@ -9,33 +9,42 @@ branching model with short-lived release branches.
...
 
 ```
-main:           ──A──B──C──D──────────────────────G──H──I── ...
-                       \                          /
-release/v0.4:           E──F────────T (tag v0.4.0)
-                                     \          /
-adopt/v0.4:                           V────────  (version fixup)
+main:           ──A──B──C──D──────────G──H──────────────K── ...
+                       \             /       \          /
+release/v0.4:           E──F──T     /     P──Q──U      /
```
harrism: **🚩 issue:** ‏P seems to be out in space -- no line connecting before it. (Probably just an alignment issue?)

devtools/start-release.sh
```diff
@@ -118,6 +138,89 @@ done
...
+
+    echo ""
+    log "Done. Hotfix v${VERSION} prepared on $RELEASE_BRANCH."
+    echo ""
+    log "Next steps:"
+    log "  1. Cherry-pick or apply fixes to $RELEASE_BRANCH"
+    log "  2. Push $RELEASE_BRANCH to $REMOTE"
+    log "  3. Wait for publish workflow to pass"
+    log "  4. Run: ./devtools/finish-release.sh $VERSION"
+    exit 0
```
harrism, line 222: **👏 praise:** ‏I like that the start-release.sh finishes with next steps instructions. The finish-release.sh should do this too!

devtools/finish-release.sh
```diff
@@ -230,7 +243,24 @@ else
...
+\`${ADOPT_BRANCH}\` branch. The version in \`pyproject.toml\` has been set to
+the current \`main\` development version to avoid conflicts.
+
+**Tag:** \`${TAG}\`
+
+### To merge
+
+Merge this PR once CI passes. Use a **merge commit** (not squash) to preserve
+the release branch history on \`main\`.
+EOF
```
harrism: **💡 suggestion:** ‏I think finish-release.sh should finish with output that tells the user next steps, including this really important point to use a merge commit (not squash).

devtools/finish-release.sh
```diff
@@ -273,4 +307,5 @@ log "Done. Release $TAG tagged."
 if ! $NO_PR && ! $DRY_RUN; then
     log "The GitHub Release will trigger the publish workflow."
     log "Merge the adopt PR from $ADOPT_BRANCH once CI passes."
```
harrism: **💡 suggestion:** ‏Ahh here it is. Just clarify that this should be a merge commit.

harrism (changes requested):
Looks great! Couple of suggestions.

devtools/finish-release.sh
swahtz: Yes I'm wondering if we need to advise the user that they should decide whether to use a merge or squash commit.  

A hotfix is, I think in most cases, going to be comprised of cherry-picks from `main` to release a fix of something that needs to be backported so the history of that commit will already be on `main`.  I imagine in those cases the user would want to squash the merge but in cases where they've had to be more involved in making a bespoke fix or merging other branches, they would want to have a merge commit.

docs/release-process.md
swahtz: I reckoned it was just hard to read with the crossing lines but perhaps it's more confusing than not having the lines there.  P is on release/v0.4, just chronologically after T, G, H, etc.

devtools/finish-release.sh
swahtz: I've updated this message, let me know what you think @harrism

docs/release-process.md
swahtz: I've added some line connections between T and P

devtools/finish-release.sh
harrism: Most commits to main will be PRs, and you never want to squash multiple PRs in the history. Also, if you squash them, and they are also in main, then when merging back into main you will get conflicts (trivial conflicts, but conflicts nonetheless).

I don't see hotfix as different from regular release here. In fact it's just a matter of patch vs. major/minor release, right?

devtools/finish-release.sh
```diff
@@ -305,7 +305,11 @@ run git checkout main
 echo ""
 log "Done. Release $TAG tagged."
 if ! $NO_PR && ! $DRY_RUN; then
-    log "The GitHub Release will trigger the publish workflow."
-    log "Merge the adopt PR from $ADOPT_BRANCH once CI passes."
-    log "After merging, re-deploy docs: gh workflow run docs.yml --ref main"
+    log "Next steps:"
+    log "  1. Wait for the publish workflow (triggered by the GitHub Release)"
+    log "  2. Merge the adopt PR from $ADOPT_BRANCH once CI passes"
+    log "     - Use 'Squash and merge' if the release only cherry-picked commits from main"
```
harrism: I don't think you should squash the commits from main if they are all PRs (each already squashed).

devtools/finish-release.sh
swahtz: Hotfix is going to be different than a major/minor release.  When we make a branch for major/minor release, we do it off `main` and all the commits on `main` and this release branch are going to be largely the same at that point.

When we release a hotfix/patch version, we have commits on `main` we want to bring over to the release branch.  We cherry-pick those commits and we get a new commit hash on that other branch that has the same content as the commit on `main`, so if we were to merge the adopt branch from this into `main` with a merge, we have two commits with the same changes.  Sidenote: GH doesn't seem to do a great job visualizing these changes when creating a PR like this… e.g. https://github.com/openvdb/fvdb-core/pull/564 which should look like a no-op besides the docs version change.

However, when we merge an adopt branch from a major/minor release into `main`, the only new commits will be any we had to put in there to fix last-minute release bugs (perhaps a fix to `publish.yml`, etc.) and it makes sense to retain that kind of history.  There won't be any duplicate commit content; all the other commits on the release branch will be shared history with `main`.

When we merge an adopt branch from a hotfix/patch release (further down the `release/vMINOR` branch), we've cherry-picked commits over to the release branch.   Even though they have the same content as something else in `main`'s history, they have different hashes because they have different ancestors and so will show as a 'duplicate' commit in the `main` history.

Our purpose for making the hotfix release, in most cases, is going to be to adopt bugfixes into a minor release that are already on `main`, and that was my argument for squashing the PR of a hotfix into `main` because all those changes will already exist in `main` and have the correct history there.

harrism: Ah, OK, I was a little confused. Now I think I see it. 

Actually I think if you ONLY cherry pick commits from main into the hotfix release, then there is NOTHING to merge back into main. Main will have all of those commits and will have a newer version than the hotfix. In this case the PR should just be closed rather than merged.

If the hotfix branch has some additional commits / PRs that are not in main, then it should be merged with a merge commit, and conflicts resolved.

Agree?

swahtz: Well the only thing that should different between the hotfix adopt branch and `main` is that the version of the package to install changed in the docs (which is important).  And docs can only be deployed from `main`.  Maybe we should change that and the docs should be deployed from the latest release branch?

On the second point, yes, if there's additional commits or PRs from other branches, then yes, merge commit preserving the history.

harrism: > Well the only thing that should different between the hotfix adopt branch and main is that the version of the package to install changed in the docs (which is important). 

But that difference is a conflict if you merge it, right? 

> Maybe we should change that and the docs should be deployed from the latest release branch?

This would be the right thing to do. Ideally our docs site would let the user select the version they are using.  RAPIDS, for example, deploys 3 versions of docs: previous version, current stable release, and next release (from main).  See screenshot. I think two would be fine (stable and dev). But this is not crucial for this PR.

image

devtools/finish-release.sh
swahtz: Done!  Good catch

devtools/finish-release.sh
swahtz: Yep, I've changed these docs now that we're not going to merge hotfix branches back into `main`

devtools/finish-release.sh
swahtz: I made an issue out of the versioned docs comment (something we've talked about casually quite a bit).  

The other comment we've obviated by just not merging a hotfix release branch back into `main` (in favour of using a GH action to update the docs stable fvdb-core version which _should_ be the only meaningful difference in a hotfix release compared to `main`).

Issue #459, opened 2026-02-11
Proposal: Functional Refactor of the GridBatch Pipeline


blackencino:
# Proposal: Functional Refactor of the GridBatch Pipeline

> **TL;DR** — Eliminate two entire C++ abstraction layers (~2,600 lines) by moving validation and ergonomic logic to Python — where it already largely exists — and replacing the wrapper code with ~25 thin, strict free functions that do only what C++ must do: autograd tape registration and CPU/CUDA template dispatch. **The public-facing Python API does not change. Existing Python tests should pass without modification.**

---

**Contents**

- [Motivation](#motivation)
- [What Changes](#what-changes)
- [What the Thin C++ Functions Look Like](#what-the-thin-c-functions-look-like)
- [What Python Gains](#what-python-gains)
- [What GridBatchImpl Becomes](#what-gridbatchimpl-becomes)
- [Immutability Cleanup in C++](#immutability-cleanup-in-c)
- [Functional Python API](#functional-python-api)
- [Design Principle: Strict in C++, Ergonomic in Python](#design-principle-strict-in-c-ergonomic-in-python)
- [Why This Is Safe](#why-this-is-safe)
- [Why This Can Be Largely Automated](#why-this-can-be-largely-automated)
- [Phasing](#phasing)
- [Metrics](#metrics)

---

### The current pipeline has four layers for every operation

```
Python  (fvdb/grid_batch.py, fvdb/grid.py)
  -> pybind binding  (src/python/GridBatchBinding.cpp)
    -> C++ wrapper class  (src/fvdb/GridBatch.h, src/fvdb/GridBatch.cpp)
      -> C++ core  (src/fvdb/detail/GridBatchImpl.h, src/fvdb/detail/GridBatchImpl.cu)
        -> dispatch / ops / autograd  (actual kernels)
```

Most operations pass through all four layers with each one doing progressively less work. By the time you reach `GridBatchImpl`, the actual computation is typically a single dispatch call. The two middle layers — `GridBatch` and the binding — exist primarily to convert types and re-validate inputs that Python already validated.

### The wrapper layer is already bypassed by the code that matters

[...]

PR #582, opened 2026-04-01
Retire C++ GridBatch wrapper; add functional API and Grid class

blackencino:
Implements #459 — the functional refactor of the GridBatch pipeline.

## What this does

Eliminates the C++ `GridBatch` wrapper class and `Vec3*OrScalar` type system (~2,600 lines), replacing them with:

- **`GridBatchData`** — the former `GridBatchImpl`, now a frozen struct bound directly to Python. No public mutation after construction.
- **`fvdb.functional`** — a `torch.nn.functional`-style module with explicit `_batch` and `_single` variants for every operation (~120 functions).
- **`Grid`** — an immutable single-grid class (`grid_count == 1`) with plain `torch.Tensor` I/O, complementing `GridBatch` which uses `JaggedTensor`.

`GridBatch` and `Grid` are now pure Python classes whose methods are one-liner delegations to the functional layer.

## How to review this

This is a large diff (213 files, ~11k lines each way) but is structurally simple. The kernel code, ops, and dispatch layer are essentially untouched — this change reorganizes the machinery *around* the kernels.

**Start here:**
1. `fvdb/functional/__init__.py` — the export list shows the full public surface
2. `fvdb/functional/_transforms.py` — a small, representative file showing the `_batch`/`_single` split pattern
3. `fvdb/grid.py` — the new `Grid` class; every method is a one-liner to `functional.*_single`
4. `fvdb/grid_batch.py` — same pattern, delegating to `functional.*_batch`

**Then skim:**
5. `fvdb/functional/_dense.py` — the `_InjectFn` autograd function, where `mark_dirty` was replaced with clone+contiguous (see commit `288b04d` for rationale)
6. `src/fvdb/detail/GridBatchData.h` — the frozen struct (formerly `GridBatchImpl`)
7. `src/python/GridBatchOps.cpp` — the thin C++ binding layer

**What you can skip:**
- `src/fvdb/detail/ops/` — kernel code, unchanged in logic (only `GridBatchImpl` → `GridBatchData` type renames)
- `src/fvdb/detail/autograd/` — C++ autograd functions are still present but most ops now use Python-side `torch.autograd.Function` equivalents
- `.github/workflows/` — merged from main, not part of this work

## What is NOT touched

- **Gaussian splatting** — 5 lines changed (added `Grid` to a type check). The gsplat kernels, rasterizer, projection, and SH code are completely untouched.
- **CUDA kernels** — no kernel logic changed. The ops layer (`src/fvdb/detail/ops/`) only has `GridBatchImpl` → `GridBatchData` type renames.
- **Viewer** — untouched.
- **nn modules** — untouched (they call Grid/GridBatch methods, which delegate to functional).

## Key design decisions

- **`_batch` vs `_single` naming** — every functional op has two explicit variants rather than polymorphic dispatch. `GridBatch` methods call `_batch`; `Grid` methods call `_single`.
- **`Grid` uses singular params** — `voxel_size`/`origin` (not plural), matching its single-grid semantics.
- **Inject autograd** — `mark_dirty` on non-contiguous views disconnects PyTorch's autograd graph. Replaced with clone+contiguous, matching the pattern the C++ autograd already used (`dst.set_jdata(ret[1])`).
- **Rank-1 validation** — single-grid constructors enforce `NumericMaxRank1` so users can't accidentally pass batch-shaped params.

## Test results

- **Python**: 2546 passed, 6 skipped (full suite excluding viz/WSL)
- **C++ gtests**: 356 passed (all suites including GatherScatterDefaultConv)
- **C++ benchmarks**: all 5 compile and run

## Known pre-existing issues (not introduced by this PR)

- `test_gaussian_splat_3d.py::test_to_other` had a `gridbatch_from_dense` scalar broadcast bug — fixed in this PR
- `test_viz.py` tests require a display (skip under WSL)


fvdb/functional/_constructors.py
```diff
@@ -0,0 +1,529 @@
...
+        _fvdb_cpp.create_from_empty(str(resolved_device), _to_vec3d_batch(voxel_sizes_t), _to_vec3d_batch(origins_t))
+    )
+
+
+# ---------------------------------------------------------------------------
+#  Concatenation
+# ---------------------------------------------------------------------------
+
+
+def concatenate_grids(grids: Sequence[GridBatch]) -> GridBatch:
```
swahtz: /
**💭 thought (non-blocking):** Do we want to have a `concatenate_grids` that takes a sequence of `Grid`s?

fvdb/functional/__init__.py
```diff
@@ -0,0 +1,310 @@
...
+    "morton_zyx_batch",
+    "morton_zyx_single",
+    "hilbert_batch",
+    "hilbert_single",
+    "hilbert_zyx_batch",
+    "hilbert_zyx_single",
+    "edge_network_batch",
+    "edge_network_single",
+    # Indexing
+    "index_grid",
```
swahtz: **❓ question:** Should this name really be `index_grid_batch` since it inputs/outputs `GridBatch`?  'Single' indexing doesn't make semantic sense but just to keep the taxonomy consistent

fvdb/functional/_constructors.py
```diff
@@ -0,0 +1,529 @@
...
+        voxel_sizes (NumericMaxRank2): Voxel size per grid, broadcastable to ``(B, 3)``.
+        origins (NumericMaxRank2): Origin per grid, broadcastable to ``(B, 3)``.
+        device (DeviceIdentifier | None): Device to create on. Defaults to ijk's device.
+
+    Returns:
+        result (GridBatch): A new grid batch.
+
+    .. seealso:: :func:`grid_from_ijk`
+    """
+    resolved_device = resolve_device(device, inherit_from=ijk)
```
swahtz: **❓ question:** There's this function and a few more below where `resolved_device` is built but not used, is it meant to be used?  If not, can we remove?

fvdb/jagged_tensor.py
```diff
@@ -48,7 +48,7 @@
 from .types import DeviceIdentifier, LShapeSpec, RShapeSpec
 
 if TYPE_CHECKING:
-    from .grid import Grid
+    pass
```
swahtz: **💅 polish:** ‏ Let's just remove this if statement

fvdb/types.py
```diff
@@ -1077,6 +1077,15 @@ def to_Vec3fBroadcastable(
     return to_FloatingTensorBroadcastableRank1(x, (3,), dtype, value_constraint=value_constraint, do_broadcast_to=False)
 
 
+def validate_rank1_voxel_params(voxel_sizes: NumericMaxRank1, origins: NumericMaxRank1) -> None:
+    """Validate that voxel_sizes and origins are rank-1 for single-grid operations.
+
```
swahtz, line 1082: **☑️ todo:** ‏ Just missing the docs for the args

fvdb/functional/_transforms.py
```diff
@@ -0,0 +1,128 @@
...
+# ---------------------------------------------------------------------------
+
+
+def voxel_to_world_batch(grid: GridBatch, ijk: JaggedTensor) -> JaggedTensor:
+    """Transform voxel-space coordinates to world-space positions for a grid batch.
+
+    Supports backpropagation.
+
+    Args:
+        grid (GridBatch): The grid batch whose transforms to use.
```
swahtz: **💡 suggestion:** Can we change the phrasing of all the 'grid' arguments docs in these functions because it's a bit awkward due to the dangling infinitive… I'd suggest:  "The grid batch supplying the voxel-to-world transforms.", or phrasing as appropriate for the othe functions.


```suggestion
        grid (GridBatch): The grid batch supplying the voxel-to-world transforms.
```

fvdb/functional/_transforms.py
```diff
@@ -0,0 +1,128 @@
...
+
+from .. import _fvdb_cpp
+from ..jagged_tensor import JaggedTensor
+
+if TYPE_CHECKING:
+    from ..grid import Grid
+    from ..grid_batch import GridBatch
+
+
+class _VoxelToWorldFn(torch.autograd.Function):
```
swahtz, line 19: **👏 praise:** ‏ Just loving moving all this autograd machinery to python.

fvdb/grid.py
```diff
@@ -188,1207 +150,1064 @@ def from_dense_axis_aligned_bounds(
...
 
+        Returns:
+            voxel_to_world_matrix (torch.Tensor): Shape ``(4, 4)``.
         """
-        resolved_device = resolve_device(device)
-        voxel_size = to_Vec3fBatch(voxel_size, value_constraint=ValueConstraint.POSITIVE)
-        origin = to_Vec3fBatch(origin)
-        grid_impl = GridBatchCpp(voxel_sizes=voxel_size, grid_origins=origin, device=resolved_device)
-        return cls(impl=grid_impl)
+        return self.data.voxel_to_world_matrix_at(0)
```
swahtz, line 449: **❓ question:** ‏This function doesn't include a `.to(self.device)` like  `voxel_size()` or `origin()`, though I wonder if `voxel_size` or `origin` actually should?  These data are on the CPU and perhaps we're assuming too much of the user's intent to force a copy to the device; perhaps they only need this on the host.  Either way, I think we should make it consistent and I reckon I'd vote that we leave these metadata on the host and not assume the user wants them copied to the same device as the `Grid`.

fvdb/functional/_constructors.py
```diff
@@ -0,0 +1,529 @@
...
+
+
+def _wrap_single_grid(cpp_impl):
+    from ..grid import Grid
+
+    return Grid(data=cpp_impl)
+
+
+def grid_from_dense(
+    dense_dims: NumericMaxRank1,
```
swahtz, line 340: **🤓 nitpick (non-blocking):** ‏ All the arguments for the `grid_` constructors have plural names where they only accept singular items.  For example dense_dims, origins, voxel_sizes are only ever for a singular grid.  We should just make these all singular.

fvdb/grid.py
```diff
@@ -1397,1038 +1216,508 @@ def ray_implicit_intersection(
...
+        return functional.inject_from_dense_cminor_single(self, dense_data, dense_origin)
 
     def inject_from_dense_cmajor(self, dense_data: torch.Tensor, dense_origin: NumericMaxRank1 = 0) -> torch.Tensor:
-        """
-        Inject values from a dense :class:`torch.Tensor` into a :class:`torch.Tensor` associated with this :class:`Grid`.
+        """Inject values from a dense tensor (CXYZ order) into sparse voxel data.
 
-        This is the "C Major" (channels major) version, which assumes the ``dense_data`` is in CXYZ order. *i.e* the
-        dense tensor has shape ``[channels*, dense_size_x, dense_size_y, dense_size_z]``.
+        ``dense_data`` has shape ``(C*, X, Y, Z)`` or ``(1, C*, X, Y, Z)``.
```
swahtz: **⌨️ typo:** ‏. Docs say that dense_data can be unbatched ``(C*, X, Y, Z)`` but `_InjectFromDenseCminorFn.backward` always indexes dense_data like this to get the grid size:

`grid_size = list(ctx.dense_shape[1:4])`

If dense_data is passed as 4D (not batched), the backward pass will extract the wrong `grid_size` values.

fvdb/grid_batch.py
```diff
@@ -142,13 +127,15 @@ class GridBatch:
...
-        """
-        Constructor for internal use only. - use the GridBatch.from_* classmethods instead.
+    def __init__(self, *, data: "_fvdb_cpp.GridBatchData") -> None:
+        """Internal constructor -- use ``GridBatch.from_*`` classmethods instead.
+
+        Args:
+            data (_fvdb_cpp.GridBatchData): The underlying C++ grid batch data object.
         """
-        self._impl = impl
+        self.data = data
```
swahtz: **💡 suggestion:** ‏ We should be able to enforce the immutability of GridBatch the same way we do with Grid with a `__slots__` and `__setattr__` override

fvdb/gaussian_splatting.py
```diff
@@ -2903,15 +2652,7 @@ def render_contributing_gaussian_ids(
...
             far (float): The far clipping plane distance for the projection.
-            camera_model (CameraModel): Semantic camera model for projection. Default is
-                :attr:`fvdb.CameraModel.PINHOLE`.
-            projection_method (ProjectionMethod): Projection implementation selector. Default is
-                :attr:`fvdb.ProjectionMethod.AUTO`.
-            distortion_coeffs (torch.Tensor | None): Distortion coefficients with shape ``(C, 12)``.
-                Required for :class:`CameraModel.OPENCV_*` camera models. For
-                :class:`CameraModel.PINHOLE` and :class:`CameraModel.ORTHOGRAPHIC`, pass
-                ``None`` or a ``(C, 12)`` tensor, which is ignored. To represent no
-                distortion with an OpenCV camera model, pass a zero-filled tensor.
```
swahtz, line 2914: **❓ question:** ‏Why remove this overload of `to`?  We have an overload where the `other` argument is a `GridBatch`, why get rid of the `Grid` overload?

fvdb/grid_batch.py
```diff
@@ -1896,103 +1343,72 @@ def refine(
...
+
+        .. seealso:: :meth:`Grid.refined_grid`
         """
-        subdiv_factor = to_Vec3iBroadcastable(subdiv_factor, value_constraint=ValueConstraint.POSITIVE)
-        return GridBatch(impl=self._impl.refined_grid(subdiv_factor, None if mask is None else mask._impl))
+        from . import functional
+
+        return functional.refined_grid_batch(self, subdiv_factor, mask)
 
     def to(self, target: "str | torch.device | torch.Tensor | JaggedTensor | GridBatch") -> "GridBatch":
```
swahtz: **❓ question:** ‏Should `to` be able to take as a `target` a `Grid`? (and vice-verse should `Grid.to` be able to take as a target argument a `GridBatch`?

fvdb/viz/_utils.py
```diff
@@ -3,14 +3,13 @@
 import torch
 
 from .._fvdb_cpp import JaggedTensor
-from ..grid import Grid
 from ..grid_batch import GridBatch
 
 
-def grid_edge_network(grid: Grid) -> tuple[torch.Tensor, torch.Tensor]:
+def grid_edge_network(grid: GridBatch) -> tuple[torch.Tensor, torch.Tensor]:
```
swahtz: **⌨️ typo:** ‏`grid_edge_network` takes a `GridBatch` but `gridbatch_edge_network` (below) also takes a `GridBatch`.  I think this input unintentionally changed to be a `GridBatch`

fvdb/convolution_plan.py
```diff
@@ -3,7 +3,7 @@
 #
 """
 Black-box encapsulation of configuration structures for sparse convolution using
-fVDB Grid and GridBatch. Design is intended to be reminiscent of the "plan" concept from FFT
+fVDB GridBatch. Design is intended to be reminiscent of the "plan" concept from FFT
```
swahtz, line 6: **❓ question:** ‏Is the removal of `Grid` support from `ConvolutionPlan` so that `GridBatch` is only supported intentional?  I don't think I saw a mention of that in the PR

src/fvdb/detail/ops/ActiveGridCoords.h
swahtz, line 1: **👏 praise:** ‏Can't believe we called this Goords for so long… 🤦‍♂️

fvdb/utils/tests/grid_utils.py
```diff
@@ -110,22 +81,17 @@ def make_grid_and_point_data(
...
-        include_boundary_points: If True, create points both inside and outside the grid
-        expand: Number of times to replicate points when creating boundary samples
 
-    Returns:
-        tuple: (primal_grid_batch, dual_grid_batch, jagged_points)
-    """
+def make_grid_batch_and_jagged_point_data(
+    device: DeviceIdentifier, dtype: torch.dtype, include_boundary_points: bool = False, expand: int = 10
+) -> tuple[GridBatch, GridBatch, JaggedTensor]:
+    """Create a grid batch (batch size 1) and jagged point data for testing."""
```
swahtz: **🧹 chore:** ‏ I don't think we should get rid of the args/returns docstrings here.

fvdb/utils/tests/grid_utils.py
```diff
@@ -87,18 +63,13 @@ def make_grid_and_point_data(
     if not include_boundary_points:
         return grid, grid_d, p
 
-    # Ensure some samples land inside and outside the voxel grid
-    # We create samples by sampling in a radius roughly the size of a voxel around each
-    # voxel center so that some points will land outside but only slightly
     found = False
```
swahtz, line 69: **🧹 chore:** ‏ I don't think we should remove the useful comments in this function, could we please restore those?

src/fvdb/detail/GridBatchData.h
```diff
@@ -23,26 +23,22 @@
...
-        VoxelCoordTransform mDualTransform;   // Dual Transform of this grid (i.e. transform which
-                                              // aligns origin with voxel corner)
-        nanovdb::Vec3d mVoxelSize;            // Size of a single voxel in world space
-        uint32_t mNumLeaves;                  // Number of leaf nodes in this grid
-        int64_t mNumVoxels;                   // Number of voxels in this grid
-        uint64_t mNumBytes;                   // Number of bytes in the buffer of this grid
-        nanovdb::CoordBBox mBBox;             // Bounding box of this grid
+        uint32_t version = 1;
+
+        int64_t mCumLeaves = 0;
```
swahtz: **🧹 chore:** These changes get rid of a fair amount of useful comments in this file which I think are useful explanations to new developers.  I think we should bring them back unless there's a reason why we removed them.

swahtz (commented):
I haven't been through all of what I wanted to review this evening but wanted to get the first parts out.  Generally think this is a great improvement.

I'm wondering a bit about whether we want to put the JaggedTensor functional components into their own module which we currently have at the top-level of `fvdb` such as `fvdb.relu/sigmoid` in future work.  Thinking as a new user of the library, they might discover all these pure functions related to grids in `fvdb.functional` and pure functions on JaggedTensors in `fvdb.*` (some of these functions like relu/sigmoid they'd [expect to find by convention in `fvdb.nn.functional`, or similar to `torch.nn.functional`](https://github.com/pytorch/pytorch/issues/22117)).

One side general question.  Now that `Grid` drops in most places that `GridBatch` can be used, do we want to change the nn modules to I/O Grid _or_ GridBatch?  I'm sure using `Grid` currently works at runtime, but our typing across the nn modules only traffics in `GridBatch`/`JaggedTensor`.  Shall we change that in future work?

fvdb/convolution_plan.py
blackencino: Convolution never supported Grid, only GridBatch, because the assumption is that in training you're always batching.

fvdb/functional/_constructors.py
blackencino: Fixed! Changed both the cat function and the from_cat constructor for GridBatch. There only needs to be one variation of this because it always produces a GridBatch.

fvdb/functional/__init__.py
blackencino: Fixed!

fvdb/functional/_constructors.py
blackencino: Fixed!

fvdb/jagged_tensor.py
blackencino: Fixed

fvdb/types.py
blackencino: Fixed

fvdb/functional/_transforms.py
blackencino: Fixed!

fvdb/functional/_transforms.py
blackencino: Me too! Thanks

fvdb/functional/_constructors.py
blackencino: Done!

fvdb/grid.py
blackencino: Fixed.

fvdb/grid_batch.py
blackencino: Done.

fvdb/gaussian_splatting.py
blackencino: Fixed

fvdb/grid_batch.py
blackencino: Fixed.

src/fvdb/detail/ops/ActiveGridCoords.h
swahtz: https://en.wikipedia.org/wiki/Gord%27s_Gold

blackencino: Goords!

fvdb/utils/tests/grid_utils.py
blackencino: Fixed.

fvdb/viz/_utils.py
blackencino: Fixed.

fvdb/utils/tests/grid_utils.py
blackencino: Fixed

src/fvdb/detail/GridBatchData.h
blackencino: Fixed

fvdb/grid.py
blackencino: I agree with this change, but it's a difference in behavior from what it did before, and will require some refactoring of tests and usage. I'd prefer to make it in a separate PR.

fwilliams (approved):
Discussed issues with Chris offline, and this is in good shape. Great refactor!

PR #588, opened 2026-04-07
Fix weighted average in TSDF integration to apply `pixelWeight` to new samples

jinhwanlazy:
The TSDF and feature blending formulas divided by `oldWeight + pixelWeight` but only weighted the old value, giving the new sample an implicit weight of 1. This produced incorrect results when per-pixel weights differ from 1.

### DIFF
--- a/src/fvdb/detail/ops/IntegrateTSDF.cu
+++ b/src/fvdb/detail/ops/IntegrateTSDF.cu
@@ -304,8 +304,8 @@ integrateTSDFKernel(const ScalarDataType truncationMargin,
             }
             const ScalarType newWeight =
                 oldWeight + pixelWeight; // ScalarType(1) + oldWeight * pixelWeight;
-            const ScalarType newTsdf        = (oldWeight * oldTsdf + pixelWeight * tsdf) / newWeight;
-            outTsdfAcc[unionWriteOffset]    = ScalarDataType(newTsdf);
+            const ScalarType newTsdf     = (oldWeight * oldTsdf + pixelWeight * tsdf) / newWeight;
+            outTsdfAcc[unionWriteOffset] = ScalarDataType(newTsdf);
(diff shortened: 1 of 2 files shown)


fwilliams (commented):
Great catch! Do you mind adding a Python unit test for the case where you have pixel weights not equal to one? It should be good to merge after that

jinhwanlazy:
@fwilliams I added a test. Hopefully I got it right!

harrism:
I think you need to merge the latest from main into your branch.

fwilliams (approved):
Looks great! Thank you for the contribution!

jinhwanlazy:
@fwilliams @harrism Thanks for the review and guidance!

harrism:
We have one remaining problem blocking merging -- @jinhwanlazy your commits are not gpg/ssh signed. For OSS security and trust on all projects, I recommend setting this up for yourself. To sign your commits with SSH, you must first tell Git to use SSH for signing, add your SSH key to GitHub as a "Signing Key," and then update your local Git configuration with that key.  This gives the little "Verified" green tag on your commits here in GH, like you can see on my commit.

There's a guide here: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

You would need to amend both of your commits on this PR and then force-push in order to fix it. Let us know if you need help.

jinhwanlazy:
Thanks for pointing this out. I had misunderstood and thought using --signoff during commits was sufficient.

In reality, I hadn’t properly set up commit signing, and I also hadn’t registered a “Signing Key” on GitHub before.

I’ve just amended the commits with proper signing and force-pushed. Hopefully everything looks correct now!

harrism:
Thanks @jinhwanlazy -- yes, we have both requirements. Sorry for the confusion. All merged!

PR #595, opened 2026-04-10
Move Gaussian splatting autograd and pipeline logic from C++ to Python

fwilliams:
## Summary

- Port all C++ autograd functions (projection, rasterization, SH evaluation) to pure Python `torch.autograd.Function` implementations in a new `_gaussian_autograd.py` module, making the differentiable pipeline easier to extend and debug.
- Eliminate the C++ `GaussianSplat3d` class entirely: the Python `GaussianSplat3d` class now directly manages `torch.Tensor` attributes instead of delegating to a C++ `_impl` object.
- Rewrite `deduplicatePixels` as a pure Python function using PyTorch ops, removing its C++ binding and implementation.
- Refactor PLY I/O to return a flat tuple of tensors (no intermediate struct), and update the viewer to accept individual tensors directly instead of a C++ object.
- Remove ~7,600 lines of dead C++ code (`GaussianSplat3d.{h,cpp}`, `detail/autograd/*`, `DeduplicatePixelsTest.cpp`, `GaussianSplat3dCameraApiTest.cpp`).
- Simplify pybind bindings to only expose low-level dispatch functions; all autograd wiring now lives in Python.

## Test plan

- [x] All 135 Python unit tests pass (`cd tests && pytest unit/test_gaussian_splat_3d.py -v`)
- [x] All 21 C++ gtests pass (`ctest --test-dir build/.../src/tests`)
- [x] CI passes (codestyle, DCO, full test matrix)


matthewdcong:
Thanks, looks good. Do we want to add a Python test for deduplicate pixels since we've removed the C++ test (and implementation)?

matthewdcong (approved):

fwilliams:
Good call — added a `TestDeduplicatePixels` class with 22 parameterized tests (11 cases x int32/int64 dtypes) covering: empty input, single pixel, all unique, some/all duplicates, multi-batch with and without duplicates, round-trip reconstruction, and jagged tensor offset verification. These mirror the deleted C++ `DeduplicatePixelsTest` suite.

swahtz:
I didn't have a chance to review this PR because this was over the weekend, but here are some notes on things that changed in this port we should address.  It's also useful to know we don't have coverage to catch some of these situations and we should get an LLM friend to write us some pytests to make sure we have coverage going forward.

###  Python autograd backward() asserts on None gradients

This is pointed out by Copilot above, but all 5 Python `torch.autograd.Function.backward()` methods in `_gaussian_autograd.py` hard-assert that incoming gradient tensors are not `None`:

```python
assert grad_means2d is not None
assert grad_depths is not None
assert grad_conics is not None
```

In PyTorch autograd, these grads **can be `None`** when a particular output isn't used in the loss (e.g., if you only backprop through rendered images but not depths). The C++ autograd handled undefined tensors gracefully. The Python version will crash with an `AssertionError` in these edge cases.


###  No input shape/contiguity validation in `gaussian_render_jagged`

The C++ validated 6 tensor shapes and 6 contiguity requirements:

```cpp
TORCH_CHECK(means.rsizes() == torch::IntArrayRef({ggz, 3}), "means must have shape (ggz, 3)");
TORCH_CHECK(quats.rsizes() == torch::IntArrayRef({ggz, 4}), "quats must have shape (ggz, 4)");
// ... etc for scales, opacities, viewmats, Ks
TORCH_CHECK(means.is_contiguous(), "means must be contiguous");
// ... etc
```

The Python `gaussian_render_jagged` in `__init__.py` has no input validation -- no shape checks, no contiguity checks.

### No shape/device/dtype validation in `from_tensors`

The C++ constructor's `checkState()` validated:
- `means` shape `(N, 3)`, `quats` shape `(N, 4)`, `logScales` shape `(N, 3)`, `logitOpacities` shape `(N,)`, `sh0` shape `(N, 1, D)`, `shN` shape `(N, K-1, D)`
- All tensors on the same device
- All tensors are floating-point
- All tensors have the same scalar type

The Python `from_tensors()` does none of this -- raw tensors are stored without validation, so errors surface inside CUDA kernels.

### `sh0` 2D-to-3D auto-reshape removed

The C++ constructor handled `sh0` being passed as `(N, D)` by automatically unsqueezing to `(N, 1, D)`:

```cpp
if (mSh0.dim() == 2) {
    mSh0 = mSh0.unsqueeze(1);
}
```

The Python `from_tensors` does not handle this, so passing 2D `sh0` will cause downstream failures.

### No SH degree validation in jagged path

The C++ checked:

```cpp
TORCH_CHECK(K >= (actualShDegree + 1) * (actualShDegree + 1),
            "K must be at least (shDegreeToUse + 1)^2");
```

The Python `gaussian_render_jagged` does not validate that `K` (number of SH basis functions) is sufficient for the requested SH degree. This could produce silently incorrect SH evaluation.

### No w2c/K shape checks or C > 0 check in `_do_projection`

The C++ `validateCameraProjectionArgs` checked `worldToCameraMatrices.sizes() == {C, 4, 4}`, `projectionMatrices.sizes() == {C, 3, 3}`, and `C > 0`. The Python `_do_projection` only checks contiguity, not shapes or non-emptiness.

### `from_state_dict` validation

The C++ `loadStateDict` had ~25 `TORCH_CHECK_VALUE` statements validating: all required keys present, gradient accumulation tensor shapes `(N,)`, devices match, dtypes are correct (`int32` for step counts/radii), consistency between gradient norms and step counts. The Python `from_state_dict` does none of this -- it just calls `state_dict.get()`.

### `cat()` method missing

The C++ had a `cat()` static method with comprehensive validation for concatenating splat objects. There is no Python equivalent.

### Deleted C++ tests not fully replaced

**GaussianSplat3dCameraApiTest** (5 test cases covering AUTO resolution, validation rejection, empty camera batches, distortion ignored for pinhole/ortho, etc.) -- No Python equivalents were added for these camera model dispatch and validation scenarios.

Issue #590, opened 2026-04-08
Functional rewrite of Gaussian Splat API


fwilliams:
# Functional Gaussian Splatting API Redesign

## Unified Decision Flowchart

All pipelines share Stages 1-2. At Stage 3, the user picks dense or sparse tile intersection. At Stage 4, the user picks their output: rendered images, world-space-gradient rendering, or analysis (counts/IDs).

```mermaid
flowchart TD
    Input["3D Gaussians + Camera Params"]

    subgraph S1 ["Stage 1: Projection"]
        proj["project_gaussians()"]
    end

    subgraph S2 ["Stage 2: Features"]
        sh["evaluate_gaussian_sh()"]
    end

    subgraph S3 ["Stage 3: Tile Intersection"]
        dense_tiles["intersect_gaussian_tiles()"]
        sparse_tiles["intersect_gaussian_tiles_sparse()"]
    end

    subgraph S4 ["Stage 4: Output"]
        rast_ss["rasterize_screen_space_gaussians()"]
        rast_ws["rasterize_world_space_gaussians()"]
        rast_sparse["rasterize_screen_space_gaussians_sparse()"]
        count_dense["count_contributing_gaussians()"]
        id_dense["identify_contributing_gaussians()"]
        count_sparse["count_contributing_gaussians_sparse()"]
        id_sparse["identify_contributing_gaussians_sparse()"]
    end

    Input --> proj
    proj -->|"ProjectedGaussians"| sh
    proj --> dense_tiles
    proj --> sparse_tiles

    dense_tiles -->|"GaussianTileIntersection"| rast_ss
    dense_tiles --> rast_ws
    dense_tiles --> count_dense
    dense_tiles --> id_dense

    sparse_tiles -->|"SparseGaussianTileIntersection"| rast_sparse
    sparse_tiles --> count_sparse
    sparse_tiles --> id_sparse

    sh -->|"features"| rast_ss
    sh --> rast_ws
    sh --> rast_sparse
```

**Decision points at each stage:**

- **Stage 1** -- One function, two internal paths: `projection_method=ANALYTIC` (differentiable) or `UNSCENTED` (forward-only). Dispatched automatically via `AUTO` based on `camera_model`.
- **Stage 2** -- One function, three `GaussianRenderMode` values: `FEATURES` (SH evaluation), `DEPTH` (passthrough), `FEATURES_AND_DEPTH` (concatenated).
[...]

PR #596, opened 2026-04-10
Refactor Gaussian splatting ops and extract utility functions

fwilliams:
## Summary

This PR is primarily code migration and renaming for consistency. It introduces no functional changes and removes some code duplication. 

What is included:
- Renames 21 Gaussian splatting operator file pairs in `ops/gsplat/` with descriptive names (e.g. `GaussianProjectionForward` -> `ProjectGaussiansAnalyticForward`, `GaussianRasterizeForward` -> `RasterizeScreenSpaceGaussiansForward`)
- Extracts shared CUDA utilities into `utils/cuda/` (BinSearch, CubWrapper, Prefetch, WarpReduce, Alignment, CopyCoords, OpType) and `utils/cuda/math/` (AffineTransform, Rotation)
- Extracts Gaussian-specific utilities into `utils/gsplat/` (Gaussian2D, GaussianCameraAccessorCopy, GaussianCameras, GaussianMath, GaussianRasterize, GaussianRasterizeFromWorld, GaussianRasterizeOptionalInputs)
- Removes the template dispatch pattern (`dispatchFoo<DeviceTag>`) from op headers in favor of plain functions with internal device dispatch
- Deletes dead gsplat files: GaussianUtils.cpp/.h/.cuh, GaussianVectorTypes, GaussianMacros, GaussianRenderSettings, GaussianRigidTransform, GaussianCameraMatrixUtils, GaussianWarpUtils
- Renames `GaussianSplatBinding.cpp` -> `GaussianSplatOps.cpp`
- Updates all include paths in bindings, viewer, PLY I/O, and 16 C++ test files
- Removes `OpType` struct in favor of PyTorch's built-in `at::opmath_type`
- Removes `GSPLAT_PRAGMA_UNROLL` struct in favor of `#pragma unroll`

Part of #590 (PR 2: refactor gsplat directory).

## Test plan

- [x] C++ build succeeds
- [x] C++ gtests: 39/41 pass (2 pre-existing flaky failures unrelated to this change: `GaussianRasterizeForwardMaskedEdgeTile.NoDeadlock` deadlock timeout, `PredGatherIGemmTest` CUDA launch failure)
- [x] Python unit tests (`cd tests && pytest unit -v`)
- [x] CI passes


src/fvdb/detail/utils/cuda/Prefetch.h
```diff
@@ -0,0 +1,130 @@
+// Copyright Contributors to the OpenVDB Project
```
matthewdcong, line 1: This file can be a .h file as opposed to a .cuh file because it does not have any device code.

src/fvdb/detail/utils/cuda/OpType.cuh
```diff
@@ -0,0 +1,41 @@
+// Copyright Contributors to the OpenVDB Project
```
matthewdcong: Can we just use `at::opmath_type` instead of declaring our own?

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -1,15 +1,15 @@
 // Copyright Contributors to the OpenVDB Project
```
matthewdcong, line 1: Should we call this EvaluateSphericalHarmonicsForward/`evaluate_spherical_harmonics_fwd` instead? There's nothing in this file that specifically deals with Gaussians.

src/fvdb/detail/utils/cuda/CubWrapper.cuh
```diff
@@ -0,0 +1,23 @@
+// Copyright Contributors to the OpenVDB Project
```
matthewdcong: Could we just use `CUB_WRAPPER` in ATen/cuda/cub.cuh?

If not, we should keep this local to the two translation units that use it in order to avoid conflicts due to macros with the same time.

matthewdcong (changes requested):

src/fvdb/detail/utils/cuda/Prefetch.h
fwilliams: Fixed: renamed `Prefetch.cuh` to `Prefetch.h` and updated include guards and all three include sites.

src/fvdb/detail/utils/cuda/OpType.cuh
fwilliams: Fixed: replaced custom `OpType<T>::type` with `at::opmath_type<T>` from `<ATen/OpMathType.h>` in the two files that use it (`IntersectGaussianTiles.cu`, `IntegrateTSDF.cu`). Removed the unused include from the remaining 8 files and deleted `OpType.cuh`.

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
fwilliams: Fixed: renamed files to `EvaluateSphericalHarmonics{Forward,Backward}.{h,cu}` and functions to `evaluate_spherical_harmonics_{fwd,bwd}`. Updated all includes, CMakeLists.txt, pybind11 bindings, Python call sites, and C++ tests.

src/fvdb/detail/utils/cuda/CubWrapper.cuh
fwilliams: Fixed: replaced custom `CubWrapper.cuh` with ATen's `CUB_WRAPPER` from `<ATen/cuda/cub.cuh>` in both `IntersectGaussianTiles.cu` and `BuildSparseGaussianTileLayout.cu`. Deleted `CubWrapper.cuh`. The ATen macro is functionally identical (same probe + allocate + call pattern using the caching allocator).

fwilliams: I ended up using a translation unit local version of the cub wrapper since the aten one seems to depend on being a debug build.

matthewdcong:
I think we also have some Python function naming leaking into the C++ side. For example, `dispatch_rasterize_screen_space_gaussians_bwd` in RasterizeScreenSpaceGaussiansBackward.cu. Unless that's intentional?

fwilliams:
> I think we also have some Python function naming leaking into the C++ side. For example, `dispatch_rasterize_screen_space_gaussians_bwd` in RasterizeScreenSpaceGaussiansBackward.cu. Unless that's intentional?

It's intentional. I'm trying to have the operator names and the binding names align. `rasterize_screen_space_gaussians` is more precise and constrasts against `rasterize_world_space_gaussians`.

matthewdcong:
> > I think we also have some Python function naming leaking into the C++ side. For example, `dispatch_rasterize_screen_space_gaussians_bwd` in RasterizeScreenSpaceGaussiansBackward.cu. Unless that's intentional?
> 
> It's intentional. I'm trying to have the operator names and the binding names align. `rasterize_screen_space_gaussians` is more precise and constrasts against `rasterize_world_space_gaussians`.

Sorry I mean the conventions -- In the C++ code, it should be named `rasterizeScreenSpaceGaussians` in order to match with the rest of the code.

harrism:
> * Moves 21 Gaussian op file pairs from `ops/gsplat/` into `ops/` with descriptive names (e.g. `GaussianProjectionForward` -> `ProjectGaussiansAnalyticForward`, `GaussianRasterizeForward` -> `RasterizeScreenSpaceGaussiansForward`)

If we move all 3DGS ops into the top-level ops directory, it would be helpful to start all 3DGS files with a common prefix. Previously we had "Gaussian".

fwilliams:
> > * Moves 21 Gaussian op file pairs from `ops/gsplat/` into `ops/` with descriptive names (e.g. `GaussianProjectionForward` -> `ProjectGaussiansAnalyticForward`, `GaussianRasterizeForward` -> `RasterizeScreenSpaceGaussiansForward`)
> 
> If we move all 3DGS ops into the top-level ops directory, it would be helpful to start all 3DGS files with a common prefix. Previously we had "Gaussian".

They are standalone ops now with a consistent naming that matches their python counterpart names. e.g. `rasterize_screen_space_gaussians_fwd` matches `RasterizeScreenSpaceGaussiansForward`. 

I'd push back a bit on adding `Gaussian` as a prefix to each op file here since we don't do this for e.g. `JaggedTensor `or GridBatch (its `SampleGridTrilinear not GridBatchSampleGridTrilinear`). My medium term goal is to have the names of files and operators match from the `_fvdb_cpp` layer down to the ops directory. The gsplat directory was large with a lot common utilities that got moved out. Now its slimmed down to about 20 ops. 

What are your thoughts?

swahtz:
> > * Moves 21 Gaussian op file pairs from `ops/gsplat/` into `ops/` with descriptive names (e.g. `GaussianProjectionForward` -> `ProjectGaussiansAnalyticForward`, `GaussianRasterizeForward` -> `RasterizeScreenSpaceGaussiansForward`)
> 
> If we move all 3DGS ops into the top-level ops directory, it would be helpful to start all 3DGS files with a common prefix. Previously we had "Gaussian".

Why do we want to move the Gaussian-related ops out of `ops/gsplat` (or whatever ops subdirectory name we might want)?  From a purely code organization standpoint I find it beneficial to have all the interdependent Gaussian-related ops together.  What's the benefit of moving them?

fwilliams:
> > > * Moves 21 Gaussian op file pairs from `ops/gsplat/` into `ops/` with descriptive names (e.g. `GaussianProjectionForward` -> `ProjectGaussiansAnalyticForward`, `GaussianRasterizeForward` -> `RasterizeScreenSpaceGaussiansForward`)
> > 
> > 
> > If we move all 3DGS ops into the top-level ops directory, it would be helpful to start all 3DGS files with a common prefix. Previously we had "Gaussian".
> 
> Why do we want to move the Gaussian-related ops out of `ops/gsplat` (or whatever ops subdirectory name we might want)? From a purely code organization standpoint I find it beneficial to have all the interdependent Gaussian-related ops together. What's the benefit of moving them?

For the same reason we haven't split up GridBatch and all the JaggedTensor ops into their own folders. If we do this, we can segment out different ops into different directories based on their type but I think its not always a clean mapping. E.g. here spherical harmonics evaluation is a generic operator that we happen to use a lot in conjunction with Gaussian splatting. 

We can certainly try and group .cu/.h pairs into folders for code organization purposes but I wonder if we shouldn't do this all in one go. 

Personally, I prefer to have all the ops be in one big folder but I don't feel that strongly one way or another. If we're going to partition op files, I prefer to do it consistently.

swahtz:
> > > > * Moves 21 Gaussian op file pairs from `ops/gsplat/` into `ops/` with descriptive names (e.g. `GaussianProjectionForward` -> `ProjectGaussiansAnalyticForward`, `GaussianRasterizeForward` -> `RasterizeScreenSpaceGaussiansForward`)
> > > 
> > > 
> > > If we move all 3DGS ops into the top-level ops directory, it would be helpful to start all 3DGS files with a common prefix. Previously we had "Gaussian".
> > 
> > 
> > Why do we want to move the Gaussian-related ops out of `ops/gsplat` (or whatever ops subdirectory name we might want)? From a purely code organization standpoint I find it beneficial to have all the interdependent Gaussian-related ops together. What's the benefit of moving them?
> 
> For the same reason we haven't split up GridBatch and all the JaggedTensor ops into their own folders. If we do this, we can segment out different ops into different directories based on their type but I think its not always a clean mapping. E.g. here spherical harmonics evaluation is a generic operator that we happen to use a lot in conjunction with Gaussian splatting.
> 
> We can certainly try and group .cu/.h pairs into folders for code organization purposes but I wonder if we shouldn't do this all in one go.
> 
> Personally, I prefer to have all the ops be in one big folder but I don't feel that strongly one way or another. If we're going to partition op files, I prefer to do it consistently.

But the reality is that there is no functionality that uses the spherical harmonics evaluation operators outside of Gaussian splatting and the arguments to the SH evaluation ops are specific to the Gaussian splatting toolchain (e.g. the means and projected radii arguments).  Until there's a reason to move them into a more generic folder, I don't see why we would change that from where they stand today.

harrism:
Unless it's a public interface that we want to simplify #include lines for, I prefer to have things more organized rather than one big dump folder.

fvdb/__init__.py
```diff
@@ -287,6 +287,11 @@ def evaluate_spherical_harmonics(
     """
     if sh_degree > 0 and view_directions is None:
         raise ValueError("view_directions is required when sh_degree > 0")
+    if view_directions is None:
+        view_directions = radii.new_empty(0)
+    if shN is None:
+        N, _, D = sh0.shape
```
swahtz, line 314: **💡 suggestion:** ‏ Since we're making assumptions here about the shape of the input, some input checking should be performed on sh0 like checkings its dimensionality and checking to see if its N is the same size as radii's N

fwilliams:
> Unless it's a public interface that we want to simplify #include lines for, I prefer to have things more organized rather than one big dump folder.



> Unless it's a public interface that we want to simplify #include lines for, I prefer to have things more organized rather than one big dump folder.

Sounds good - I moved it into a folder for better organization. Let me know if you guys are happy with the PR otherwise!

src/python/GaussianSplatOps.cpp
```diff
@@ -0,0 +1,572 @@
...
+          &ops::buildSparseGaussianTileLayout,
+          py::arg("tile_side_length"),
+          py::arg("num_tiles_w"),
+          py::arg("num_tiles_h"),
+          py::arg("pixels_to_render"));
+
+    // ------- UT projection forward (non-differentiable) -------
+
+    m.def(
+        "project_gaussians_ut_fwd",
```
swahtz: **💅 polish:** ‏ In this refactor,  the abbreviation 'sh' changed to `spherical_harmonics` (in functions like `evaluate_spherical_harmonics_fwd`), I think we should also change ut to `unscented_transform` for consistency/clarity

fvdb/__init__.py
swahtz, line 1: **☑️ todo:** ‏I can't comment on it because it's not part of the review, but it looks like the `_fvdb_cpp.pyi` was left out of these changs and needs to be updated based on these changes.  I wonder if there's a way to automatically generate this.

src/python/GaussianSplatOps.cpp
```diff
@@ -0,0 +1,572 @@
...
+
+#include <torch/extension.h>
+
+void
+bind_gaussian_splat_ops(py::module &m) {
+    namespace ops         = fvdb::detail::ops;
+    using DistortionModel = fvdb::detail::ops::DistortionModel;
+
+    // -----------------------------------------------------------------------
+    // Enum types (moved from GaussianSplatBinding.cpp)
```
swahtz: **🧹 chore:** ‏ The "(moved from …)" comment is more of a comment about this PR and won't be relevant to the new code, I think we should just remove this.

src/python/GaussianSplatOps.cpp
```diff
@@ -0,0 +1,572 @@
...
+    // Data types needed by the functional ops
+    // -----------------------------------------------------------------------
+
+    // -----------------------------------------------------------------------
+    // Kernel-level bindings (for Python autograd and composition)
+    // -----------------------------------------------------------------------
+
+    // -----------------------------------------------------------------------
+    // Analysis operations (call raw tensor dispatch functions directly)
+    // -----------------------------------------------------------------------
```
swahtz, line 72: **🧹 chore:** ‏ Are the first two comment blocks left-overs?  If so, those could be removed.  Also the "(call raw tensor dispatch…)" expression seems like an implementaiton detail and could be removed

src/python/GaussianSplatOps.cpp
```diff
@@ -0,0 +1,572 @@
...
+          py::arg("means"),
+          py::arg("log_scales"),
+          py::arg("logit_opacities"),
+          py::arg("quats"),
+          py::arg("noise_scale"),
+          py::arg("t"),
+          py::arg("k"));
+
+    // -----------------------------------------------------------------------
+    // PLY I/O (wraps C++ PLY functions directly with raw tensors)
```
swahtz: **💅 polish:** ‏Can just removed the parenthetical here too

src/python/GaussianSplatOps.cpp
```diff
@@ -0,0 +1,572 @@
...
+          py::arg("metadata") = py::none());
+
+    m.def("load_gaussian_ply",
+          &fvdb::detail::io::loadGaussianPly,
+          py::arg("filename"),
+          py::arg("device") = torch::kCPU);
+
+    // ------- Raw forward/backward dispatch (for Python autograd) -------
+
+    // 1. project_gaussians_analytic_fwd
```
swahtz: **❓ question:** ‏ Why are all these functions numbered?  Can we just remove the numbers in the comments?

fvdb/gaussian_splatting.py
```diff
@@ -1591,9 +1592,9 @@ def _eval_sh(
         if sh_degree_to_use > 0:
             cam_to_world, info = torch.linalg.inv_ex(w2c)
             view_dirs = means[None, :, :] - cam_to_world[:, None, :3, 3]
-            return _EvaluateGaussianSHFn.apply(sh_degree_to_use, C, view_dirs, sh0, shN, radii)
         else:
-            return _EvaluateGaussianSHFn.apply(sh_degree_to_use, C, None, sh0, None, radii)
+            view_dirs = means.new_empty(0)
```
swahtz, line 1601: **💡 suggestion:** ‏ Like in the `fvdb.evaluate_spherical_harmonics` function, we could also use an empty tensor for shN so that autograd won't unnecessarily hold the shN tensor during degree=0 evaluation.  Alternatively, I wonder if we couldn't just call `fvdb.evaluate_spherical_harmonics` here and de-duplicate some logic?
```suggestion
            view_dirs = means.new_empty(0)
            shN = sh0.new_empty(sh0.new_empty(self.num_gaussians, 0, sh0.size(2)))
```

src/fvdb/detail/ops/gsplat/IdentifyTopContributingGaussians.h
```diff
@@ -0,0 +1,58 @@
...
+#include <tuple>
+
+namespace fvdb {
+namespace detail {
+namespace ops {
+
+/// @brief Performs deep image rasterization to render the IDs and weighted alpha values of the
+/// top-K most visible Gaussians for each pixel (dispatch wrapper).
+std::tuple<torch::Tensor, torch::Tensor>
+identifyTopContributingGaussians(const torch::Tensor &means2d,
```
swahtz, line 21: **❓ question:** ‏ The form of these names seems to have changed.  Can these be changed back to rasterizeTopContributingGaussianIds, etc.

src/fvdb/detail/utils/cuda/math/AffineTransform.cuh
```diff
@@ -0,0 +1,64 @@
...
+
+#include <tuple>
+
+namespace fvdb {
+namespace detail {
+
+/// Transform a point from world to camera coordinates: p_cam = R * p_world + t.
+template <typename T>
+inline __device__ nanovdb::math::Vec3<T>
+transformPointWorldToCam(nanovdb::math::Mat3<T> const &worldToCamRotation,
```
swahtz, line 23: **🧹 chore:** ‏ Most of the new files here in `utils/cuda` are lacking doxygen documentation comments.  `BinSearch.cuh` has them but the other new headers should get doxygen comments to match those.

src/fvdb/detail/utils/gsplat/GaussianMath.cuh
```diff
@@ -0,0 +1,131 @@
...
+
+#include <nanovdb/math/Math.h>
+
+#include <cstdint>
+#include <tuple>
+
+namespace fvdb {
+namespace detail {
+
+/// @brief VJP for quaternion-and-scale to covariance matrix transformation.
```
swahtz, line 18: **☑️ todo:** ‏ This function (and several below) lack their param/return docs

src/fvdb/detail/utils/gsplat/GaussianMath.cuh
```diff
@@ -0,0 +1,131 @@
...
+
+/// Transmittance threshold below which a pixel is considered fully opaque.
+/// Matches the 3DGS convention (exclusive comparison).
+constexpr float kTransmittanceThreshold = 1e-4f;
+
+using tilePixelMaskAccessor                       = fvdb::TorchRAcc64<uint64_t, 2>;
+static constexpr uint32_t sTileBitmaskBitsPerWord = 64;
+
+inline uint32_t
+numWordsPerTileBitmask(const uint32_t tileSideLength) {
```
swahtz, line 141: **☑️ todo:** ‏ This function and the rest below are lacking docs.

src/fvdb/detail/utils/gsplat/GaussianRasterizeFromWorld.cuh
```diff
@@ -94,65 +96,12 @@ struct RasterizeFromWorldCommonArgs {
...
-/// Load quaternion in [w,x,y,z] order.
-template <typename T>
-inline __device__ nanovdb::math::Vec4<T>
-quatLoadWxyz(const T *q) {
-    return nanovdb::math::Vec4<T>(q[0], q[1], q[2], q[3]);
-}
-
 /// Build S^{-1} R^T from quaternion + scale.
 template <typename T>
 inline __device__ nanovdb::math::Mat3<T>
```
swahtz, line 116: **💡 suggestion:** ‏ Perhaps this function and `isclRotVectorJacobianProduct` move to `cuda/math/Rotation.cuh`?  Now that they've been moved from living directly alongside the code that used these functions, we should give them unabbreviated names and full doxygen documentation in the comments.

src/fvdb/detail/utils/gaussian/GaussianRasterizeOptionalInputs.h
```diff
@@ -19,13 +19,13 @@ struct PreparedRasterOptionalInputs {
...
 
 inline PreparedRasterOptionalInputs
-prepareRasterOptionalInputs(const torch::Tensor &features,
-                            const int64_t C,
-                            const int64_t tileExtentH,
-                            const int64_t tileExtentW,
-                            const int64_t numChannels,
-                            const at::optional<torch::Tensor> &backgrounds,
-                            const at::optional<torch::Tensor> &masks) {
+prepare_raster_optional_inputs(const torch::Tensor &features,
```
swahtz: **❓ question:** ‏ My understanding is that the intention was to use snake case for functions that had python bindings (to align their names).  I don't think this function is bound to python in these changes, I think this function should remain camel-case unless I'm misundesstanding the intent?

swahtz (changes requested):
Loving the removal of the templated dispatch functions everywhere.  Just a few docs/naming and other improvement comments.  Thanks

matthewdcong:
If we're going to keep the splatting operators in a subdirectory, can we just keep the gsplat folder name instead of introducing a new gaussian_splatting folder? It would make the changes easier to follow through the history in the short term, and we can rename it in a straightforward follow up PR.

fvdb/__init__.py
matthewdcong: I looked in my notes and `stubgen -m fvdb._Cpp` was what I used back in the day to get most of the way there.

src/python/GaussianSplatOps.cpp
fwilliams: Good thought on consistency. The distinction I'm drawing is that "SH" was expanded to `spherical_harmonics` because spherical harmonics is the *subject* of `evaluate_spherical_harmonics` -- it's what the function computes. In `project_gaussians_ut_fwd`, "UT" is describing the *method* used for projection, not the subject. Expanding it to `project_gaussians_unscented_transform_fwd` makes an already long name quite unwieldy. UT is also a well-established abbreviation in this domain (the Julier & Uhlmann Unscented Transform). I'd prefer to keep it short here -- happy to add a doc comment on the binding explaining the abbreviation if that helps clarity.

src/fvdb/detail/ops/gsplat/IdentifyTopContributingGaussians.h
fwilliams: Good question -- you're right that the implementation walks the full tile-based rasterization pipeline internally (the kernel is literally called `rasterizeTopContributingGaussianIdsForward`). But I went through a lot of design iterations on the naming for the functional API (see issue #590), and the naming here is deliberate.

The functional pipeline splits into four conceptual stages: **projection**, **tile intersection**, **rasterization**, and **analysis**. The distinction between rasterization and analysis isn't "per-pixel output vs. not" -- both produce per-pixel results. The distinction is:

1. **Rasterize ops synthesize a new signal** -- they alpha-composite Gaussian features into images that didn't exist before. **Analysis ops report metadata** about the compositing process -- which Gaussians participated and by how much. It's synthesis vs. introspection.
2. **Rasterize ops are differentiable** (they all have backward passes). **Analysis ops are non-differentiable by design** -- they're diagnostic/query tools. This is a fundamental API-level distinction.
3. **The output types are structurally different.** Rasterize ops return dense `Tensor` images (float pixel grids). Analysis ops return `JaggedTensor` of integer IDs and weights -- variable-length per-pixel lists. Categorically different result types.

The verb "identify" describes what the operation does for the caller -- it answers "which Gaussians matter at each pixel?" The fact that rasterization is the mechanism used to answer that is an implementation detail (and the internal kernel names still reflect that). The family reads naturally as a coherent set of analysis operations: `countContributingGaussians`, `identifyContributingGaussians`, `identifyTopContributingGaussians`.

The stale names in `_fvdb_cpp.pyi` are from before this rename and are being cleaned up in this push.

fvdb/__init__.py
fwilliams: Fixed in latest push.

fvdb/__init__.py
fwilliams: Fixed in latest push.

src/python/GaussianSplatOps.cpp
fwilliams: Fixed in latest push.

src/python/GaussianSplatOps.cpp
fwilliams: Fixed in latest push.

src/python/GaussianSplatOps.cpp
fwilliams: Fixed in latest push.

src/python/GaussianSplatOps.cpp
fwilliams: Fixed in latest push.

fvdb/gaussian_splatting.py
fwilliams: Fixed in latest push.

src/fvdb/detail/utils/cuda/math/AffineTransform.cuh
fwilliams: Fixed in latest push.

src/fvdb/detail/utils/gsplat/GaussianMath.cuh
fwilliams: Fixed in latest push.

src/fvdb/detail/utils/gsplat/GaussianMath.cuh
fwilliams: Fixed in latest push.

src/fvdb/detail/utils/gsplat/GaussianRasterizeFromWorld.cuh
fwilliams: Fixed in latest push.

src/fvdb/detail/utils/gaussian/GaussianRasterizeOptionalInputs.h
fwilliams: Fixed in latest push.

fwilliams:
Done -- renamed back to `gsplat/` in the latest push.

src/fvdb/detail/ops/gsplat/FusedSSIM.cu
```diff
@@ -580,50 +580,7 @@ fusedSSIMBackwardCUDA(double C1,
...
-                                             prefetchLocations.data(),
-                                             prefetchLocationIndices.data(),
-                                             prefetchLocations.size(),
-                                             0,
-                                             stream));
-#endif
-}
-
-} // namespace
+namespace {} // namespace
```
matthewdcong: this can be removed

src/fvdb/detail/ops/gsplat/FusedSSIM.cu
```diff
@@ -1,3 +1,6 @@
+// Copyright Contributors to the OpenVDB Project
```
matthewdcong: Is this the right thing to do for vendored code?

fwilliams: I'm not sure - deferring to @harrism who added this code in

src/fvdb/detail/ops/gsplat/FusedSSIM.cu
fwilliams: Fixed in latest push.

src/fvdb/detail/ops/gsplat/FusedSSIM.cu
harrism: We already have the SPDX below the original MIT license below. So this addition duplicates it. I think you can remove the lower one. But definitely retain the OSRB reference and the original license.

src/fvdb/detail/utils/cuda/Prefetch.h
matthewdcong, line 1: These prefetch functions used to be implemented in the .cpp file as opposed to being inlined in a header file. Can we keep it that way? From a development standpoint, prefetching requires quite a bit of tuning and keeping the implementation in a separate .cpp file significantly reduces incremental compile times.

src/fvdb/detail/utils/gsplat/Gaussian2D.cuh
```diff
@@ -1,10 +1,8 @@
 // Copyright Contributors to the OpenVDB Project
```
matthewdcong, line 1: These utils in utils/gaussian are almost used exclusively in the gsplat ops (with the exception of one include) in the viewer. Should we just move them to the gsplat directory to keep things more self-contained?

src/fvdb/detail/ops/gsplat/FusedSSIM.cu
```diff
@@ -841,8 +796,6 @@ fusedSSIMBackwardPrivateUse1(double C1,
                 H,
                 W,
                 CH,
-                static_cast<float>(C1),
```
matthewdcong, line 844: I know these parameters are possibly redundant but can we please keep them? In order to stay aligned with the upstream implementation.

matthewdcong (commented):
Sorry for the continued comments. I have some more feedback after checking out your branch and playing around with it.

src/fvdb/detail/utils/cuda/Prefetch.h
```diff
@@ -0,0 +1,130 @@
...
+    cudaMemLocation location                       = {cudaMemLocationTypeDevice, deviceId};
+    std::vector<cudaMemLocation> prefetchLocations = {location};
+
+    for (size_t i = 0; i < tensors.size(); ++i) {
+        const auto &tensor = tensors[i];
+        TORCH_CHECK(tensor.is_contiguous(), "Tensor to prefetch is not contiguous");
+        prefetchPointers.emplace_back(tensor.data_ptr<float>() + localElementOffset);
+        prefetchSizes.emplace_back(localElementCount * sizeof(float));
+    }
+    std::vector<size_t> prefetchLocationIndices(prefetchPointers.size(), 0);
```
matthewdcong: Yikes! That's a bug I think. `prefetchLocationIndices` needs to match the length of `prefetchLocations`, not `prefetchPointers`.

src/fvdb/detail/utils/cuda/Prefetch.h
```diff
@@ -0,0 +1,130 @@
...
+        const auto &tensor = tensors[i];
+        TORCH_CHECK(tensor.is_contiguous(), "Tensor to prefetch is not contiguous");
+        TORCH_CHECK(cameraOffset + cameraCount <= tensor.size(0),
+                    "Tensor does not have a batched first dimension");
+        size_t scalarSize = c10::elementSize(tensor.scalar_type());
+        prefetchPointers.emplace_back(static_cast<uint8_t *>(tensor.data_ptr()) +
+                                      cameraOffset * tensor.stride(0) * scalarSize);
+        prefetchSizes.emplace_back(cameraCount * tensor.stride(0) * scalarSize);
+    }
+    std::vector<size_t> prefetchLocationIndices(prefetchPointers.size(), 0);
```
matthewdcong: Same bug

src/fvdb/detail/ops/gsplat/FusedSSIM.cu
matthewdcong: I remember now. I was the one that added it in, and yeah we already have it below. So I'm going to remove this duplicate addition.

src/fvdb/detail/utils/cuda/Prefetch.h
matthewdcong: I moved the implementations to a .cpp file in my commit

src/fvdb/detail/utils/cuda/Prefetch.h
matthewdcong: I fixed this one but I'm still concerned how a subtle bug arose. What motivated the change?

src/fvdb/detail/ops/gsplat/FusedSSIM.cu
matthewdcong: Pushed up a commit to restore them

src/python/GaussianSplatOps.cpp
swahtz: I think the distinction between whether to expand these abbreviations for clarity for subjects or methods is somewhat ad hoc; one could equally argue that for clarity not abbreviating methods compared to subjects is more important because what a method does is less obvious compared to the 'what' it operates on (which also usually has a descriptive argument name).  Regardless, we don't really have a consistent style across the codebase (we've got abbreviated subjects like TSDF, PLY and abbreviated method descriptions like SSIM, MCMC, GS [gather-scatter]).

What about `project_gaussians_unscented_fwd`?  If not, I'm happy with `ut` and a explanation of the abbreviation to clarify for the reader.

swahtz: We have been trying to remove abbreviations across the codebase recently but if we do want to come up with a consistent paradigm for abbreviations, I think we should all chat and document it in our CODESTYLE.  It might be hard to use the distinction of method/subject as a means because it would make some of our existing abbreviations be written out as very long clauses.  Otherwise, we might want to just have a glossary and indicate we always abbreviate a concept with a specific abbreviation (e.g. is 'GS' Gaussian splat or gather-scatter?)

src/fvdb/detail/ops/gsplat/IdentifyTopContributingGaussians.h
swahtz: 3 I'm not sure is the strongest distinction; when doing deep image rasterization we'd be producing variable-length per-pixel quantities (for the use case of doing proper volume rendering where samples can be correctly composited with other volume renders without artefacs) but points 1 and 2 would both be satisfied.

I'm fine with the analytic/rasterize distinction moving forward, though, those are fine names.

src/python/GaussianSplatOps.cpp
fwilliams: You're making a good point about the inconsistency across the codebase -- you're right that the subject/method distinction I was drawing is ad hoc and doesn't hold up well against all our existing abbreviations (TSDF, SSIM, MCMC, etc.).

I like `project_gaussians_unscented_fwd` as a compromise. It expands the abbreviation enough to be clear without getting unwieldy. I'll make the rename.

And I think your idea of documenting an abbreviation policy in CODESTYLE is the right next step. We should have an explicit glossary so there's no ambiguity about what GS means, for instance. Happy to help draft that as a follow-up.

src/fvdb/detail/utils/gsplat/Gaussian2D.cuh
fwilliams: I'd prefer to keep these in `utils/gaussian/` rather than moving them into `ops/gsplat/`. The design in issue #590 intentionally separates three layers:

- **`utils/cuda/`** -- general-purpose CUDA utilities (nothing Gaussian-specific)
- **`utils/gaussian/`** -- Gaussian-domain shared infrastructure
- **`ops/gsplat/`** -- per-op kernel code

The practical reason is that `GaussianCameras.cuh` is already consumed outside of `ops/gsplat/` -- specifically by the viewer (`Viewer.h`, `ViewerBinding.cpp`) and by tests (`GaussianProjectionForwardTest.cpp`, `GaussianCamerasTest.cu`, `GaussianUtilsTest.cu`). If we move these files into `ops/gsplat/`, those external consumers would need to reach into the ops directory, which inverts the dependency direction.

The intent is that `utils/gaussian/` holds infrastructure that's Gaussian-specific but shared across consumers (ops, viewer, tests), while `ops/gsplat/` holds the actual kernel dispatch code. I think this keeps the boundaries clean. It's also consistent with the rest of the utils directory being common utilities for ops in the library,

src/fvdb/detail/utils/cuda/Prefetch.h
fwilliams: Good catch -- this was my mistake when extracting the inline prefetch code into the shared `Prefetch.h` header. Your original code correctly used `std::vector<size_t> prefetchLocationIndices = {0}` (a single-element vector matching the single-element `prefetchLocations`). When I refactored it into a generic utility function, I changed the initializer to `std::vector<size_t> prefetchLocationIndices(prefetchPointers.size(), 0)` -- I think the intent was to make it scale with the number of tensors, but that's wrong because `prefetchLocationIndices` needs to match `prefetchLocations`, not `prefetchPointers`.

The bug is benign at runtime since `numPrefetchLocs = 1` means CUDA only reads `prefetchLocationIndices[0]`, but it violates the API contract and allocates more memory than needed. Thanks for catching and fixing it.

Also agree on moving the implementations back to a `.cpp` to keep incremental compile times down -- thanks for doing that too.

src/fvdb/detail/utils/cuda/Prefetch.h
fwilliams: Same root cause as above -- the `= {0}` initializer was incorrectly changed to `(size, 0)` during the extraction into the shared header. Thanks for the fix.

src/fvdb/detail/ops/gsplat/IdentifyTopContributingGaussians.h
fwilliams: Thanks Jonathan, appreciate the thoughtful discussion.

src/fvdb/detail/utils/cuda/Prefetch.h
fwilliams: Thanks for moving this back -- agree that keeping the implementation in a `.cpp` is better for incremental compile times.

src/fvdb/detail/ops/gsplat/FusedSSIM.cu
fwilliams: Makes sense, thanks for restoring them.

src/fvdb/detail/utils/gsplat/Gaussian2D.cuh
matthewdcong: Can we name it utils/gsplat then to keep it consistent?

fwilliams: Makes sense -- renamed `utils/gaussian/` to `utils/gsplat/` for consistency with `ops/gsplat/`. Done in the latest push.

swahtz (approved):
Looking good, thanks for taking on board my notes and explaining the changes.

matthewdcong:
Can we update the PR and title description to reflect the latest since it ends up going into the final squashed MR? Also, please add that `GSPLAT_PRAGMA_UNROLL` has been replaced by `#pragma unroll`.

matthewdcong:
Modified the description and title. Lmk if anything is off

matthewdcong (approved):

PR #600, opened 2026-04-13
Materialize repeated opacities for compatibility with multiple splatting implementations

matthewdcong:
Currently, we unsqueeze and expand the opacities (following the optimization in #451) creating a non-contiguous view. However, the image space mGPU rasterization and world space rasterization implementations require contiguous expanded opacities. This PR changes opacities to always be expanded for consistent behavior across multiple splatting implementations. Note that we can't simply call `contiguous()` in the specific operator impl because the shared autograd code caches the opacities for the backwards pass.

This undoes the optimization in #451 and results in a small (~0.5%) decrease in performance in the single GPU case. However, this is outweighed by the corresponding speedups in the mGPU and world space case due to one fewer allocation and more efficient paging.

### DIFF
--- a/fvdb/gaussian_splatting.py
+++ b/fvdb/gaussian_splatting.py
@@ -1625,7 +1625,10 @@ class GaussianSplat3d:
     ) -> torch.Tensor:
         """Sigmoid of logit_opacities, optionally scaled by antialias compensations."""
 
-        # TODO(fvdb): Avoid materializing a repeated [C,N] tensor when opacities are shared across cameras (similar to PR #451).
+        # Ideally, we would like to avoid materializing the repeated [C,N] tensor when opacities
+        # are shared across cameras by replacing the repeat call with .unsqueeze(0).expand(C, -1).
+        # However, a non-contiguous opacities tensor is not currently supported in world space
+        # rasterization and mGPU image space rasterization.
         opacities = torch.sigmoid(self._logit_opacities).repeat(C, 1)
         if antialias and compensations is not None:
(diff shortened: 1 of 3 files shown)


fwilliams (approved):

fvdb/gaussian_splatting.py
```diff
@@ -1624,7 +1624,9 @@ def _make_opacities(
         antialias: bool,
     ) -> torch.Tensor:
         """Sigmoid of logit_opacities, optionally scaled by antialias compensations."""
-        opacities = torch.sigmoid(self._logit_opacities).unsqueeze(0).expand(C, -1)
+
+        # TODO(fvdb): Avoid materializing a repeated [C,N] tensor when opacities are shared across cameras (similar to PR #451).
```
swahtz: **❓ question:** ‏Why leave this as a TODO?  Aren't we explicitly undoing what PR #451 was doing to not have diverging paths for single GPU camera-space and mGPU camera-space and (any-device config) world-space?  Should we leave that as a note instead in case someone naively wants to fix this TODO with #451's approach?

matthewdcong: I think ideally we do want to avoid materializing the full tensor once all codepaths support it. I'll improve the TODO to make this more clear.

swahtz: Great, thanks, I think having an explanation is much more clear.  Prevents an eager developer or over-zealous LLM from redoing the undo.

swahtz (approved):
Looks good, thanks!

Issue #602, opened 2026-04-14
Support rendering arbitrary per-Gaussian features without SH evaluation


swahtz:
NOTE:  This is documenting an Issue for the code comment from https://github.com/openvdb/fvdb-core/blob/421bc31da1f43a4c08aadafe0a02d76b54f0e8a7/src/fvdb/GaussianSplat3d.cpp#L1100

The Gaussian projection pipeline currently always evaluates spherical harmonics on the input SH/feature tensors. This means that functions like `renderContributingGaussianIds`, which need to render a non-SH quantity (integer IDs), must go through `projectGaussiansImpl` and pay for a full SH evaluation whose results are then discarded.

**Desired behavior:**

- Allow rendering of arbitrary per-Gaussian feature tensors (any dtype/shape) without requiring SH evaluation — i.e. a "passthrough" shading model.
- Longer-term, support pluggable shading models during projection so that SH is just one option alongside raw feature passthrough or future alternatives.

**Current workaround:**

`renderContributingGaussianIdsImpl` calls the existing projection path (which evaluates SH) and then ignores the SH output, passing hard-coded Gaussian IDs to a dedicated "deep IDs" rasterize kernel. This wastes compute on the unused SH evaluation.

**Challenges / scope:**

- Much of the existing code assumes `sh0`/`shN` tensors are the primary per-Gaussian features. Supporting an additional feature tensor alongside (or instead of) the SH tensors would require refactoring that assumption in multiple places.

**Reference:** `src/fvdb/GaussianSplat3d.cpp`, `renderContributingGaussianIdsImpl` (~line 1100)

PR #603, opened 2026-04-14
Restore useful comments from autograd/pipeline refactor

swahtz:
Several useful comments/todos/fixmes that were part of GaussianSplat3d.cpp were not carried over during PR #595, this attempts to restore those back to appropriate locations

### DIFF
--- a/fvdb/__init__.py
+++ b/fvdb/__init__.py
@@ -79,6 +79,7 @@ from ._gaussian_autograd import (
 )
 
 
+# TODO: Make a batched class to encapsulate this jagged rendering pipeline.
 def gaussian_render_jagged(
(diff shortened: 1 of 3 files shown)


harrism (approved):
My only question is whether some or all of these would be better as issues than as comments. Thoughts?

swahtz:
> My only question is whether some or all of these would be better as issues than as comments. Thoughts?

Yes so definitely this one I turned into an issue https://github.com/openvdb/fvdb-core/issues/602

I think some of the other 'TODOs' are perhaps more like explanations of the implementation and could be downgraded to 'NOTE'.  I at least wanted to preserve their intent alongside the code in the same places.

PR #608, opened 2026-04-15
Initialize gradient accumulation tensors before UT projection path

harrism:
## Summary
- Move lazy initialization of gradient accumulation tensors (`_accumulated_gradient_step_counts`, `_accumulated_mean_2d_gradient_norms`, `_accumulated_max_2d_radii`) to before the Unscented Transform early return in `_do_projection`
- Add test proving the UT path now initializes gradient state when `accumulate_mean_2d_gradients=True`

## Root Cause

The UT projection path in `_do_projection` returned early (line ~1500) without lazily initializing the gradient accumulation tensors. The lazy init block (lines ~1524-1547) only ran for the ANALYTIC path. This left `accumulated_gradient_step_counts` and `accumulated_mean_2d_gradient_norms` as `None` for the entire training run when using OpenCV camera models, causing downstream consumers to crash.

## Fix

Move the lazy init block before the `_use_ut()` check. The UT kernel (`_C.project_gaussians_ut_fwd`) does not accumulate into these tensors, so they remain zero — a correct "no gradient data" representation. Full UT gradient accumulation support can follow as a separate enhancement.

See: https://github.com/openvdb/fvdb-reality-capture/issues/279

## Test plan
- [x] New test `test_ut_projection_initializes_gradient_accumulation` fails before fix (asserting None), passes after
- [x] All 8 `TestGaussianCameraApi` tests pass with zero regressions

### DIFF
--- a/fvdb/gaussian_splatting.py
+++ b/fvdb/gaussian_splatting.py
@@ -1497,30 +1497,6 @@ class GaussianSplat3d:
             if distortion_coeffs is None:
                 raise RuntimeError("distortionCoeffs must be provided for OpenCV camera models")
 
-        if self._use_ut(camera_model, projection_method):
-            if distortion_coeffs is None:
-                distortion_coeffs = torch.empty(C, 0, device=means.device, dtype=means.dtype)
-            result = _C.project_gaussians_ut_fwd(
-                means,
-                quats,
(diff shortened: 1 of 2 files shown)


swahtz (approved):
lgtm

PR #610, opened 2026-04-16
Versioned Documentation:  Documentation integration with Read the Docs

### LINKED ISSUE #576: Versioned Documentation
It would be preferable to have a docs website that includes a historical list of release versions that the user could select in order to reference the documentation of past versions (not just whatever is latest on `main`).

This might take some research and discovery for how to best implement this with the least amount of technical debt.

> Maybe we should change that and the docs should be deployed from the latest release branch?

This would be the right thing to do. Ideally our docs site would let the user select the version they are using.  RAPIDS, for example, deploys 3 versions of docs: previous version, current stable release, and next release (from main).  See screenshot. I think two would be fine (stable and dev). But this is not crucial for this PR.

image

_Originally posted by @harrism in https://github.com/openvdb/fvdb-core/pull/563#discussion_r3011906993_

swahtz:
Replace the EC2-based docs.yml workflow (which compiled fvdb with CUDA to run sphinx-build) with a Read the Docs configuration that uses autodoc_mock_imports to mock the C++ extension. This eliminates the need for CUDA hardware during doc builds and gives us automatic versioned documentation from git tags.

Changes:
- Add .readthedocs.yaml and docs/requirements.txt for RTD builds
- Add autodoc_mock_imports for _fvdb_cpp in docs/conf.py so Sphinx can introspect the Python API without the compiled extension
- Make versions.json loading resilient to FileNotFoundError (old tags lack this file)
- Set Sphinx version/release from fvdb_core_stable_version
- Add html_context for "Edit on GitHub" links
- Change docs.yml to deploy a redirect from openvdb.github.io to fvdb-core.readthedocs.io (removes expensive module build and GPU run)
- Update docs/release-process.md to reflect RTD-based workflow

fixes Versioned Documentation
Fixes #576


harrism (approved):
This is so much simpler. Nice work! Can we see what the docs will look like yet?

swahtz:
> This is so much simpler. Nice work! Can we see what the docs will look like yet?

Soon!  Once I merge this in I'll have a config file for RTD and I can make RTD build it and I'll send them to you.

PR #611, opened 2026-04-16
CI: GH Actions Version Updates

swahtz:
This is just to update versions of the Actions we use because there have been various notices about upgrading to actions that are compatible with Node 24 for security reasons.  Also updated the Convolution notebook to have more aggressive training parameters to make it faster to run.


harrism (approved):
Thanks for doing this. I was putting it off.

PR #624, opened 2026-04-17
Reorder Gaussian2D to improve field alignment

matthewdcong:
Currently, the alignment of the fields in Gaussian2D is not optimal. The `xy` member is a `float2`, but it is 4-byte aligned so it cannot be loaded in a vectorized manner. This can be addressed by switching the order to `opacity` and `xy`.

### DIFF
--- a/src/fvdb/detail/utils/gsplat/Gaussian2D.cuh
+++ b/src/fvdb/detail/utils/gsplat/Gaussian2D.cuh
@@ -8,14 +8,18 @@
 
 namespace fvdb::detail::ops {
 
-template <typename ScalarType> struct alignas(32) Gaussian2D { // 28 bytes
+template <typename ScalarType> struct alignas(32) Gaussian2D { // 32 bytes
     using vec2t = nanovdb::math::Vec2<ScalarType>;
     using vec3t = nanovdb::math::Vec3<ScalarType>;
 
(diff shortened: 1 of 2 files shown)


src/fvdb/detail/utils/gsplat/Gaussian2D.cuh
```diff
@@ -8,14 +8,18 @@
...
-    ScalarType opacity; // 4 bytes
-    vec3t conic;        // 12 bytes
+    // First 16 bytes: id(4) + opacity(4) + xy(8, 8-byte aligned)
+    int32_t id;         // 4 bytes  (offset 0)
+    ScalarType opacity; // 4 bytes  (offset 4)
+    vec2t xy;           // 8 bytes  (offset 8)
+
+    // Second 16 bytes: conic(12) + pad(4)
+    vec3t conic;     // 12 bytes (offset 16)
+    ScalarType _pad; // 4 bytes  (offset 28)
```
harrism: Is the padding even necessary if alignas(32) is used?

matthewdcong: You're right, it's not necessary due to the `alignas(32)`. Removed the padding.

harrism (approved):
Nice catch.

PR #628, opened 2026-04-18
Add `sample_nearest` operator for GridBatch and Grid

### LINKED ISSUE #612: Add `sample_nearest` operator to `GridBatch` and `Grid`
Add a `sample_nearest` method to `GridBatch` (and `Grid`) that performs nearest-neighbor sampling of voxel data at arbitrary world-space query positions. This complements the existing `sample_trilinear` and `sample_bezier` interpolation methods with a piecewise-constant alternative.

### The problem

Users frequently need to look up voxel features at world-space query points **without interpolation**. Currently, this requires manually chaining together several primitives:

```python
voxel_coords = grid.world_to_voxel(points)
coord_ijks = fvdb.JaggedTensor(voxel_coords.jdata.round().int(), voxel_coords.joffsets)
indices = grid.ijk_to_index(coord_ijks, cumulative=True)
result = voxel_data.jdata[indices.jdata]
```

This is verbose, easy to get wrong (forgetting to handle `-1` for inactive voxels, constructing the intermediate `JaggedTensor` correctly), and creates unnecessary intermediate tensors that a fused kernel could avoid.

### Concrete use case: sparse UNet point feature unprojection

A common pattern in sparse 3D UNets (e.g., MinkUNet-style architectures built on fVDB) is:

1. **Voxelize** a point cloud into a `GridBatch`, aggregating per-point features into per-voxel features.
2. Run the **encoder-decoder** UNet, producing per-voxel output features on the finest grid.
3. **Unproject** the per-voxel features back to the original point coordinates -- each point receives the exact feature of the voxel it falls in, with no blending.

Step 3 is a nearest-neighbor lookup. In practice this looks like ([`fvdb_unet.py`](https://github.com/example/sauron/blob/main/sauron/models/backbones/fvdb_unet.py)):

```python
# During voxelization -- pre-compute the manual nearest-neighbor index map:
voxel_coords = grid.world_to_voxel(coordinates)
coord_ijks = fvdb.JaggedTensor(voxel_coords.jdata.round().int(), voxel_coords.joffsets)
inverse_indices = grid.ijk_to_index(coord_ijks, cumulative=True)

# ... run UNet encoder/decoder ...

# Unproject voxel features back to original points:
points = features.jagged_like(up_data.jdata[inverse_indices.jdata])
```

With `sample_nearest`, step 3 becomes:

```python
points = up_grid.sample_nearest(coordinates, up_data)
```

Trilinear interpolation is wrong here because the network outputs are discrete per-voxel features (not a continuous field), and multiple input points mapping to the same voxel must receive the identical feature vector.

### Other use cases

- **Discrete/categorical data** (labels, class IDs, material indices) where interpolation produces meaningless intermediate values.
- **Hash-grid style feature lookups** where you want the exact stored feature, not a blend.
- **Debugging and visualization**, where seeing raw unblended voxel values at query points is useful.

## Proposed API

On `GridBatch`:

```python
def sample_nearest(self, points: JaggedTensor, voxel_data: JaggedTensor) -> JaggedTensor:
    """
    Sample data associated with this GridBatch at world-space points using
    nearest-neighbor lookup (value of the closest active voxel center among
    the 8 surrounding corners).

    Args:
        points: World-space query positions.
            Shape: (batch_size, num_points_for_grid_b, 3).
        voxel_data: Per-voxel data.
            Shape: (batch_size, total_voxels, channels*).

    Returns:
        Sampled data at each point.
            Shape: (batch_size, num_points_for_grid_b, channels*).
            Returns zero for points where none of the 8 surrounding corners
            are active, matching sample_trilinear's boundary behavior.
    """
```

And the equivalent on `Grid`:

```python
def sample_nearest(self, points: torch.Tensor, voxel_data: torch.Tensor) -> torch.Tensor:
```

## Implementation

The implementation mirrors `sample_trilinear`'s structure. For each query point:

1. Transform the world-space point to voxel space and find the floor ijk (same as trilinear).
2. Check the same 8 surrounding corners using the NanoVDB accessor (`resolveTrilinearStencil`-style pattern from `TrilinearStencil.h`).
3. Among the **active** corners, select the one closest (in voxel-space Euclidean distance) to the query point.
4. Return that corner's voxel data value. If no corners are active (`activeMask == 0`), return zero.

This reuses the same 8 accessor lookups and `activeMask` logic, just with selection-by-distance instead of weighted blending. The boundary behavior is consistent with `sample_trilinear`: a point up to one voxel outside the active region can still find an active corner and return its value; points where all 8 corners are inactive return zero.

### Differentiability

- **w.r.t. `voxel_data`:** Supported. The gather from `voxel_data` by index is differentiable, so gradients flow back to the selected voxel entry.
- **w.r.t. `points`:** Gradient is zero almost everywhere (piecewise-constant function). This is expected and correct for nearest-neighbor. Users who need position gradients should use `sample_trilinear`.

### Files to modify (following `sample_trilinear` as template)

**C++ forward and backward kernels:**
- `src/fvdb/detail/ops/SampleNearest.cu` -- new, modeled on `SampleTrilinear.cu`
- `src/fvdb/detail/ops/SampleNearest.h` -- new
- `src/fvdb/detail/ops/SampleNearestBackward.cu` -- new, backward kernel (scatter grad to selected voxel); modeled on `SampleTrilinearWithGradBackward.cu`
- `src/CMakeLists.txt` -- add new source files

**C++ bindings:**
- `src/python/GridBatchOps.cpp` -- expose `sample_nearest` to Python

**Python autograd and API (following existing pattern in `fvdb/functional/`):**
- `fvdb/functional/_interpolation.py` -- add `_SampleNearestFn(torch.autograd.Function)` with forward/backward, plus public `sample_nearest_batch` and `sample_nearest_single` functions
- `fvdb/functional/__init__.py` -- export new functions
- `fvdb/grid_batch.py` -- add `GridBatch.sample_nearest` method delegating to `functional.sample_nearest_batch`
- `fvdb/grid.py` -- add `Grid.sample_nearest` method delegating to `functional.sample_nearest_single`
- `fvdb/_fvdb_cpp.pyi` -- type stub

**Tests:**
- `tests/unit/test_sample.py` -- unit tests

## Acceptance Criteria

- [ ] `GridBatch.sample_nearest` and `Grid.sample_nearest` implemented with CUDA (and CPU) support.
- [ ] Python autograd backward pass w.r.t. `voxel_data` (in `fvdb/functional/_interpolation.py`).
- [ ] Points where all 8 corners are inactive return zero (matching `sample_trilinear` boundary convention).
- [ ] Unit tests covering: correctness vs. brute-force reference, gradient checks, empty grids, batch consistency, and boundary behavior.
- [ ] Python bindings and docstrings following existing `sample_trilinear` conventions.

## Future extensions

- `max_search_radius` parameter to expand beyond the 8-corner stencil using tree-guided shell search with NanoVDB's `getDim` to skip empty tiles efficiently.

swahtz:
Implements the `sample_nearest` operator for `GridBatch` and `Grid`, providing nearest-neighbor sampling of voxel data at arbitrary world-space query positions. Fixes #612.

- **CUDA forward kernel** (`SampleNearest.cu/.h`): For each query point, transforms to voxel space, checks the 8 surrounding voxel centers via the NanoVDB accessor, and selects the closest active one by Euclidean distance. Returns zero when all 8 are inactive, consistent with `sample_trilinear` boundary behavior. Includes both scalar and vectorized `float4` dispatch paths, mirroring the existing trilinear kernel structure.
- **Python autograd** (`_SampleNearestFn` in `_interpolation.py`): The C++ kernel returns sampled values and the selected voxel index per query point. The backward pass uses `torch.index_add_` to scatter gradients back to the selected voxels, following the project's convention of keeping autograd in Python.
- **Public API**: `sample_nearest` on `GridBatch` and `Grid`, plus `sample_nearest_batch`/`sample_nearest_single` in `fvdb.functional`.
- **Tests**: Brute-force reference implementation plus 4 parameterized test cases covering correctness vs. reference (with gradients), boundary points, exact voxel-center queries, and fully-inactive regions.

## Test plan

- [x] `cd tests && pytest unit/test_sample.py -k "nearest" -v` -- all 4 new test methods pass across device/dtype/channel combinations
- [x] Existing `test_sample.py` tests still pass (no regressions to trilinear/bezier)
- [x] Build succeeds with `SampleNearest.cu` added to CMakeLists

### DIFF
--- a/fvdb/functional/_interpolation.py
+++ b/fvdb/functional/_interpolation.py
@@ -160,6 +160,8 @@ def sample_nearest_batch(
     """Sample voxel data at world-space points using nearest-neighbor lookup for a grid batch.
 
     For each query point the 8 nearest voxel centers are checked and the value of the closest active one is returned.
+    When two or more active corners are equidistant, the corner encountered first in the stencil's zig-zag
+    traversal order wins (the same ordering used by ``sample_trilinear``).
     Points where none of the 8 surrounding voxel centers are active return zero.
 
     Supports backpropagation w.r.t. ``voxel_data``.
@@ -186,6 +188,8 @@ def sample_nearest_single(
     """Sample voxel data at world-space points using nearest-neighbor lookup for a single grid.
 
     For each query point the 8 nearest voxel centers are checked and the value of the closest active one is returned.
+    When two or more active corners are equidistant, the corner encountered first in the stencil's zig-zag
+    traversal order wins (the same ordering used by ``sample_trilinear``).
     Points where none of the 8 surrounding voxel centers are active return zero.
 
     Supports backpropagation w.r.t. ``voxel_data``.
--- a/fvdb/functional/_interpolation.py
+++ b/fvdb/functional/_interpolation.py
@@ -4,7 +4,7 @@
 """Functional API for sparse grid interpolation: sampling and splatting."""
 from __future__ import annotations
 
-from typing import Any, TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, Any, cast
 
 import torch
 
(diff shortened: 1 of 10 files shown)


fvdb/functional/_interpolation.py
```diff
@@ -123,6 +123,88 @@ def backward(ctx: Any, *grad_outputs: torch.Tensor | None) -> tuple[torch.Tensor
...
+
+
+def sample_nearest_batch(
+    grid: GridBatch,
+    points: JaggedTensor,
+    voxel_data: JaggedTensor,
+) -> JaggedTensor:
+    """Sample voxel data at world-space points using nearest-neighbor lookup for a grid batch.
+
+    For each query point the 8 nearest voxel centers are checked and the value of the closest active one is returned.
```
harrism, line 162: **❓ question:** ‏How are distance ties decided? Seems like we should document that the order is implementation-defined so that users don't rely on this.

harrism (approved):
Looks great.

fvdb/functional/_interpolation.py
swahtz: The first corner encountered in the stencil's zig-zag traversal order is the one that wins.  Sure, good point, I'll document that.

PR #632, opened 2026-04-21
Promote GridBatchData to public header

swahtz:
Move GridBatchData, TorchDeviceBuffer, VoxelCoordTransform, and TypeTraits out of fvdb::detail into the fvdb namespace as public headers. This was because `fvdb::detail::GridBatchData.h` was used in the public FVDB.h header but GridBatchData.h was in the private detail headers (ones we don't ship when building libfvdb making it impossible to use downstream)

Key changes:
- Move GridBatchData.{h,cu} from detail/ to fvdb/
- Move TorchDeviceBuffer.{h,cpp} from detail/ to fvdb/
- Move VoxelCoordTransform.h from detail/ to fvdb/
- Add new TypeTraits.h public header
- Move non-inlined GridBatchData methods (hostAccessor, deviceAccessor, nanoGridHandle, device, isEmpty, checkDevice) to the .cu file to keep the header lightweight
- Update all include paths and namespace references across detail/ops, detail/io, detail/utils, python bindings, tests, and benchmarks
- Add missing Utils.h includes in IjkForMesh.cu, NearestIjkForPoints.cu, and Refine.cu (AT_DISPATCH_V2 was previously resolved transitively)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -50,7 +50,7 @@ set(FVDB_CPP_FILES
     fvdb/detail/utils/cuda/Prefetch.cpp
     fvdb/detail/io/SaveNanoVDB.cpp
     fvdb/detail/ops/jagged/JaggedReductions.cpp
-    fvdb/detail/TorchDeviceBuffer.cpp
+    fvdb/TorchDeviceBuffer.cpp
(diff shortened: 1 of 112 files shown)


harrism (approved):
Thanks for doing this!

PR #633, opened 2026-04-21
Add CMake installation support for public headers and configuration files

swahtz:
This enhances the CMake build system by adding installation rules for public headers, including those for downstream C++ consumers. Key changes include:
- Installation of public headers from the `src/fvdb` directory.
- Inclusion of headers from the `nanovdb` library and specific macros.
- Configuration files for `find_package(fvdb)` to facilitate usage in other projects.
- Refactoring of target include directories to support both build and install interfaces.

Additionally, the commit removes the obsolete `CONV_FIX_PLAN.md` documentation file.


blackencino (approved):
LGTM!

PR #638, opened 2026-04-23
Add `__launch_bounds__` to forEach CUDA kernels

swahtz:
## Summary

- Replace the runtime `numThreads` function argument in `ForEachCUDA.cuh` and `ForEachPrivateUse1.cuh` with a compile-time `NumThreads` template parameter (defaulting to `DEFAULT_BLOCK_DIM` = 256), enabling `__launch_bounds__` on all `__global__` kernels for better register allocation and occupancy tuning.
- For jagged/tensor forEach families, template params are ordered `<ScalarT, NDIMS, NumThreads=DEFAULT_BLOCK_DIM, ...>` so existing call sites like `<scalar_t, 2>` work unchanged and a non-default block size can be appended as `<scalar_t, 2, 512>`.
- Update ~40 `.cu` call sites to remove the former runtime `numThreads` argument; callers needing non-default block sizes now specify them as a trailing template arg.
- Remove erroneous `__launch_bounds__` from `__device__`/`__hostdev__` callback functions that were causing compiler errors.
- Convert `SimpleOpHelper.h` runtime `num_threads` parameter to a template parameter on both `execute()` overloads.
- Fix `GridDim.h` `GET_BLOCKS` from `static` to `inline` to avoid `-Wunused-function` errors in translation units that include the header but don't call it.

In my experience, this has given me a free 5% speed gain in kernels like `SampleRaysUniform`.

## Test plan

- [x] Full build completes without warnings or errors (`./build.sh`)
- [x] C++ test suite passes (`./build.sh ctest`)
- [x] Python test suite passes (`cd tests && pytest unit -v`)
- [x] Spot-check that kernels using non-default block sizes (e.g. 384, 512, 1024) still launch correctly

### DIFF
--- a/src/fvdb/detail/ops/ActiveVoxelsInBoundsMask.cu
+++ b/src/fvdb/detail/ops/ActiveVoxelsInBoundsMask.cu
@@ -64,7 +64,7 @@ GetActiveVoxelsInBoundsMask(const GridBatchData &gridBatch,
             activeGridVoxelInBoundsMaskCallback<TorchRAcc64>(
                 batchIdx, leafIdx, voxelIdx, gridAccessor, bboxAcc, outMaskAcc);
         };
-        forEachVoxelCUDA(1024, 1, gridBatch, cb);
+        forEachVoxelCUDA<1024>(1, gridBatch, cb);
     } else {
         auto cb = [=](int32_t batchIdx,
                       int32_t leafIdx,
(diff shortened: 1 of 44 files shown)


harrism (approved):
Nice to see all the hard coded numbers cleaned up. 640 is a strange block size...

swahtz:
> Nice to see all the hard coded numbers cleaned up. 640 is a strange block size...

Thanks.  Yes I'm not sure what that one's about…

PR #639, opened 2026-04-24
Add Vec2 and double fast paths to `SampleGridTrilinear`

swahtz:
## Add Vec2 and double fast paths to `SampleGridTrilinear`

Broaden the GPU trilinear-sampling fast paths so they cover more shapes:

- Introduce `AlignedVec2<T>` / `AlignedVec4<T>`: `alignas(2*sizeof(T))` / `alignas(4*sizeof(T))` subclasses of `nanovdb::math::Vec2<T>` / `nanovdb::math::Vec4<T>`. nanovdb's `Vec2`/`Vec4` only carry `alignof(T)` (4B for float, 8B for double), so reinterpreting a row as `Vec2<T>*`/`Vec4<T>*` would let nvcc split the access into multiple scalar loads. Bumping the alignment to match `float2`/`double2`/`float4` is what the runtime `aligned8`/`aligned16` checks rely on to guarantee a single `LDG.64` / `LDG.128` transaction. Subclassing keeps `+=`, `* scalar`, etc. usable without rewriting any arithmetic. No `AlignedVec4<double>` is used because CUDA has no single-instruction 16B vector load for doubles; the double path tops out at `Vec2`.
- Add `sampleTrilinearStencilCallbackVec2<ScalarType>`: one-thread-per-point callback for `numChannels == 2` that issues one 8-byte (float) or 16-byte (double) aligned vector load per stencil corner, halving per-sample memory transactions vs. the scalar path.
- Template the existing `sampleTrilinearStencilCallbackVec4` on `ScalarType` (instantiated for `float` only today; structurally generic).
- In `SampleGridTrilinear`'s GPU dispatcher, compute alignment once (`aligned8` / `aligned16`) and pick:
  - `float`  + `nc == 2`                + 8B  aligned -> Vec2 path
  - `float`  + `nc >= 4 && nc % 4 == 0` + 16B aligned -> Vec4 path
  - `double` + `nc == 2`                + 16B aligned -> Vec2 path
  - otherwise -> scalar callback

  The scalar callback is hoisted as a shared lambda so each branch just dispatches to the right specialization.

CPU paths and non-`{float,double}` scalar types continue to use the scalar callback unchanged; correctness is unaffected — this is a throughput optimization only.

## Tests

- Extended the existing `test_sample` matrix to include `num_channels == 2`, exercising the new Vec2 fast path for both float and double.
- Added `test_trilinear_sparse_misaligned_input`, which slices a flat buffer to produce a contiguous-but-misaligned `[N, C]` view (`data_ptr = storage_base + sizeof(dtype)`), defeating both the `aligned8` and `aligned16` checks. It sweeps the same `(device, dtype, num_channels)` matrix as the aligned tests, asserts parity with the naive reference, and guards against an accidental future re-alignment (e.g. an extra `.contiguous()`) silently neutering the fallback coverage.

## Performance

On a Blackwell RTX 6000 PRO, trilinear sampling of a 2-channel float feature in my application sees a ~21% speedup in trilinear-sampling time.

### DIFF
--- a/src/fvdb/detail/ops/SampleTrilinear.cu
+++ b/src/fvdb/detail/ops/SampleTrilinear.cu
@@ -64,23 +64,43 @@ sampleTrilinearStencilCallback(int32_t bidx,
     }
 }
 
-// Maps a scalar type to its matching 2-wide CUDA vector type. Only the types we
-// actually emit a Vec2 fast path for are specialized.
-template <typename T> struct Vec2Trait;
-template <> struct Vec2Trait<float> {
-    using type = float2;
-};
-template <> struct Vec2Trait<double> {
-    using type = double2;
+namespace {
+
+// Aligned wrapper around nanovdb::math::Vec2<T>. nanovdb's Vec2 stores a bare
+// T[2] without an alignas, so its natural alignment is alignof(T) (4B for
+// float, 8B for double). Reinterpreting a row as Vec2<T>* and dereferencing
+// would let nvcc split the access into two scalar loads. Bumping the alignment
+// to 2*sizeof(T) (8B for float, 16B for double) matches float2/double2 and is
+// the contract the runtime address checks (`aligned8`/`aligned16`) rely on to
+// guarantee a single LDG.64/LDG.128 transaction. Inheriting from Vec2 keeps
+// arithmetic operators (`+=`, `*scalar`, ...) usable without rewriting them.
+template <typename T> struct alignas(2 * sizeof(T)) AlignedVec2 : nanovdb::math::Vec2<T> {
+    using Base = nanovdb::math::Vec2<T>;
+    __hostdev__
+    AlignedVec2()
+        : Base(T(0), T(0)) {}
+    __hostdev__
+    AlignedVec2(const Base &b)
+        : Base(b) {}
 };
 
-// Maps a scalar type to its matching 4-wide CUDA vector type. Only float is
-// specialized: CUDA has no native 4-wide double vector with single-instruction
-// 16-byte load semantics (double2 is the widest aligned double vector type).
-template <typename T> struct Vec4Trait;
-template <> struct Vec4Trait<float> {
-    using type = float4;
+// Aligned wrapper around nanovdb::math::Vec4<T>. Same rationale as
+// AlignedVec2: nanovdb::math::Vec4 has alignof(T) (4B for float), but a single
+// LDG.128 needs the address (and the type) to be 16B aligned. Only
+// instantiated for float in this TU because CUDA has no native 4-wide double
+// vector with single-instruction 16-byte load semantics; doubles fall through
+// to the scalar path or the Vec2 fast path.
+template <typename T> struct alignas(4 * sizeof(T)) AlignedVec4 : nanovdb::math::Vec4<T> {
+    using Base = nanovdb::math::Vec4<T>;
+    __hostdev__
+    AlignedVec4()
+        : Base(T(0), T(0), T(0), T(0)) {}
+    __hostdev__
+    AlignedVec4(const Base &b)
+        : Base(b) {}
 };
+
+} // namespace
 
 // One-thread-per-point callback specialized for numChannels == 2. Resolves the
 // 8-corner stencil once, then issues a single 8-byte (float2) or 16-byte
@@ -99,7 +119,7 @@ sampleTrilinearStencilCallbackVec2(int32_t bidx,
                                    TensorAccessor<ScalarType, 2> gridData,
                                    BatchGridAccessor batchAccessor,
                                    TensorAccessor<ScalarType, 2> outFeatures) {
-    using Vec = typename Vec2Trait<ScalarType>::type;
+    using Vec = AlignedVec2<ScalarType>;
 
     const auto &pointsData               = points.data();
     const nanovdb::OnIndexGrid *grid     = batchAccessor.grid(bidx);
@@ -117,13 +137,11 @@ sampleTrilinearStencilCallbackVec2(int32_t bidx,
     if (activeMask == 0)
         return;
 
-    Vec accum{};
+    Vec accum;
 #pragma unroll
     for (int corner = 0; corner < 8; ++corner) {
-        const ScalarType wt = weights[corner];
-        const Vec val       = *reinterpret_cast<const Vec *>(&gridData[indices[corner]][0]);
-        accum.x += wt * val.x;
-        accum.y += wt * val.y;
+        const Vec val = *reinterpret_cast<const Vec *>(&gridData[indices[corner]][0]);
+        accum += val * weights[corner];
     }
     *reinterpret_cast<Vec *>(&outFeatures[eidx][0]) = accum;
 }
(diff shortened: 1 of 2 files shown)


src/fvdb/detail/ops/SampleTrilinear.cu
```diff
@@ -64,31 +64,104 @@ sampleTrilinearStencilCallback(int32_t bidx,
...
+          typename TensorAccessor>
+__device__ void
+sampleTrilinearStencilCallbackVec2(int32_t bidx,
+                                   int32_t eidx,
+                                   int32_t /*cidx*/,
+                                   JaggedAccessor<ScalarType, 2> points,
+                                   TensorAccessor<ScalarType, 2> gridData,
+                                   BatchGridAccessor batchAccessor,
+                                   TensorAccessor<ScalarType, 2> outFeatures) {
+    using Vec = typename Vec2Trait<ScalarType>::type;
```
harrism: **❓ question:** ‏Why can't this just be `nanovdb::Vec2<ScalarType>`? As long as nano's vec types are properly aligned (declare `alignas` on the struct), then you will get vectorized loads/stores. As a bonus, you get operators on those types, which `float2`/`double2` in the CUDA headers do not have.

src/fvdb/detail/ops/SampleTrilinear.cu
```diff
@@ -64,31 +64,104 @@ sampleTrilinearStencilCallback(int32_t bidx,
...
+    int64_t indices[8] = {};
+    ScalarType weights[8];
+    const uint8_t activeMask = resolveTrilinearStencil(xyz, gridAcc, baseOffset, indices, weights);
+
+    if (activeMask == 0)
+        return;
+
+    Vec accum;
+    accum.x = ScalarType(0);
+    accum.y = ScalarType(0);
```
harrism: **🤓 nitpick (non-blocking):** ‏This could be simplified to one line. This zero-initializes:
```suggestion
    Vec accum{};
```

src/fvdb/detail/ops/SampleTrilinear.cu
```diff
@@ -64,31 +64,104 @@ sampleTrilinearStencilCallback(int32_t bidx,
...
-template <template <typename T, int32_t D> typename JaggedAccessor,
+// Maps a scalar type to its matching 2-wide CUDA vector type. Only the types we
+// actually emit a Vec2 fast path for are specialized.
+template <typename T> struct Vec2Trait;
+template <> struct Vec2Trait<float> {
+    using type = float2;
+};
+template <> struct Vec2Trait<double> {
+    using type = double2;
+};
```
harrism: **💡 suggestion:** ‏Put these in an anonymous namespace. Could also define them in a type traits header if we have one. But I still wonder if you could just use nano's vec types.

tests/unit/test_sample.py
```diff
@@ -23,7 +23,7 @@
 ]
 
 all_device_dtype_channel_combos = [
-    (*combo, nc) for combo, nc in itertools.product(all_device_dtype_combos, [1, 3, 4, 5, 16])
+    (*combo, nc) for combo, nc in itertools.product(all_device_dtype_combos, [1, 2, 3, 4, 5, 16])
```
harrism, line 26: **💡 suggestion:** ‏Possible testing gap here -- you cover channels and dtypes but not misalignment fallback.  A simple test that takes a slice (`feats[:, 1:]` after padding `feats` with one extra leading channel) to misalign the start, then calls `sample_trilinear`, would exercise the misalignment fallback and catch future regressions.

harrism (changes requested):
A few small suggestions.

src/fvdb/detail/ops/SampleTrilinear.cu
swahtz: True!

src/fvdb/detail/ops/SampleTrilinear.cu
swahtz: Unfortunately nano's vec types lack alignment.  But I realized I could just create a wrapper of them with alignment which gets me the consistency with the rest of fvdb and the math operators which makes the code more concise.

src/fvdb/detail/ops/SampleTrilinear.cu
swahtz: Done

tests/unit/test_sample.py
swahtz: Good call.  I added a `test_trilinear_sparse_misaligned_input` which constructs a misaligned [N, C] view by slicing one element off a flat buffer (data_ptr = storage_base + sizeof(dtype)), triggering both the aligned8 and aligned16 checks

src/fvdb/detail/ops/SampleTrilinear.cu
```diff
@@ -64,31 +64,120 @@ sampleTrilinearStencilCallback(int32_t bidx,
...
+namespace {
+
+// Aligned wrapper around nanovdb::math::Vec2<T>. nanovdb's Vec2 stores a bare
+// T[2] without an alignas, so its natural alignment is alignof(T) (4B for
+// float, 8B for double). Reinterpreting a row as Vec2<T>* and dereferencing
+// would let nvcc split the access into two scalar loads. Bumping the alignment
+// to 2*sizeof(T) (8B for float, 16B for double) matches float2/double2 and is
+// the contract the runtime address checks (`aligned8`/`aligned16`) rely on to
+// guarantee a single LDG.64/LDG.128 transaction. Inheriting from Vec2 keeps
+// arithmetic operators (`+=`, `*scalar`, ...) usable without rewriting them.
```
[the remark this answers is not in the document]
swahtz: Fixed

src/fvdb/detail/ops/SampleTrilinear.cu
```diff
@@ -64,31 +64,120 @@ sampleTrilinearStencilCallback(int32_t bidx,
...
+}
+
+// One-thread-per-point callback specialized for numChannels % 4 == 0. Resolves
+// the 8-corner stencil once, then iterates channels in 4-wide groups with a
+// single 16-byte (float4) vector load per corner per group, quartering
+// per-sample transactions relative to the scalar path. GPU only, float only.
+// Caller is responsible for alignment checks.
+template <typename ScalarType,
+          template <typename T, int32_t D>
+          typename JaggedAccessor,
```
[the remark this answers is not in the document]
swahtz: Sure, would also be useful to explain the logic for this in code as well as the comments

src/fvdb/detail/ops/SampleTrilinear.cu
```diff
@@ -64,31 +64,125 @@ sampleTrilinearStencilCallback(int32_t bidx,
...
+// to the scalar path or the Vec2 fast path.
+template <typename T> struct alignas(4 * sizeof(T)) AlignedVec4 : nanovdb::math::Vec4<T> {
+    using Base = nanovdb::math::Vec4<T>;
+    __hostdev__
+    AlignedVec4()
+        : Base(T(0), T(0), T(0), T(0)) {}
+    __hostdev__
+    AlignedVec4(const Base &b)
+        : Base(b) {}
+};
```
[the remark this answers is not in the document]
swahtz: The wrappers are actually `is_standard_layout` and `is_trivially_copyable` (the inheritance does not break that since `nanovdb::math::Vec2/Vec4` have all-private data and we add no new members), and the project compiles with C++20 where standard-layout + trivially-copyable types are implicit-lifetime. I've added `static_assert`s for size/alignment/standard-layout/trivially-copyable on `AlignedVec2<float>`, `AlignedVec2<double>`, and `AlignedVec4<float>`, and switched the load/store sites to `*std::launder(reinterpret_cast<Vec*>(...))` so the object identity is explicit to the compiler. I tried the `std::memcpy` variant first but verified via `cuobjdump --dump-sass` that NVCC drops the LDG.128/STG.128 coalescing in that case (the user-defined ctors block its memcpy folder). The `std::launder` form preserves bit-for-bit identical SASS to the original `reinterpret_cast`.

harrism (approved):

PR #640, opened 2026-04-24
Optimizations for `volume_render` and move its autograd layer to Python

swahtz:
Rewrite the volume-render forward pass around per-ray register accumulation plus a dedicated inference fast path, and move the backward plumbing out of C++ autograd into a Python torch.autograd.Function wrapper to match our other functional ops.

- volumeRenderFwdCallback now accumulates `rgb`, `depth`, and `opacity` into per-thread registers (a stack-allocated `accRgb[MAX_VOLUME_RENDER_CHANNELS]` with `#pragma unroll` plus scalars for depth/opacity) and writes each per-ray output exactly once, eliminating the per-sample global-memory read-modify-write on `outRGB` / `outDepth` / `outOpacity`. Output tensors are now allocated with `torch::empty` (plus `torch::zeros` for `outWs` on the training path, since backward's `thrust::inclusive_scan` reads into the early-termination tail).

- Add a compile-time `NeedsBackward` template parameter on `volumeRenderFwdCallback`, `volumeRenderCPU`, and the `volumeRender` CUDA kernel. When false the kernel skips the per-sample `outWs` store and the per-ray `outDepth` / `outTotalSamples` stores, and the host returns size-0 placeholder tensors with matching dtype/device. The dispatch layer picks the specialization via a tiny `std::true_type` / `std::false_type` tag-dispatch lambda inside `AT_DISPATCH_V2`.

- Replace the public `fvdb::volumeRender` with `fvdb::volumeRenderForward` / `fvdb::volumeRenderBackward` in `FVDB.h` / `FVDB.cpp`, and bind them to Python as `_fvdb_cpp.volume_render_fwd` / `volume_render_bwd` in `src/python/Bindings.cpp` (with updated `_fvdb_cpp.pyi` stubs).

- Delete `src/fvdb/detail/autograd/VolumeRender.{cpp,h}` (and its `CMakeLists.txt` entry) and implement the autograd layer in Python (`fvdb/_volume_render.py`), following the same `torch.autograd.Function` pattern as `_gaussian_autograd.py`. The new `volume_render` wrapper routes through `_VolumeRenderFn.apply` only when grad is enabled and at least one differentiable input has `requires_grad=True`; otherwise it calls `_fvdb_cpp.volume_render_fwd(..., needs_backward=False)` directly and returns size-0 placeholders for `depth` / `ws` / `total_samples`. `fvdb/__init__.py` / `__init__.pyi` now re-export `volume_render` from this new module instead of `_fvdb_cpp`.

- Add full Doxygen docstrings for both `volumeRenderForward` / `volumeRenderBackward` in the public `FVDB.h` and their `detail::ops` counterparts in `detail/ops/VolumeRender.h`, documenting all parameters, shapes, the early-termination condition (`T <= transmittanceThresh`), the channel-count limit (`1 <= C <= MAX_VOLUME_RENDER_CHANNELS`), and the inference fast path's size-0 output behavior.

- Update the one `ws` / `total_samples`-asserting test (`test_volume_render_total_samples_counts_terminating_sample`) to call `sigmas.requires_grad_(True)` so it continues to exercise the backward-aware path now that the no-grad path returns size-0 placeholders for those tensors.

Performance (fire renderer, 256 spp):

| stage                  |     baseline | + register accum | + inference fast path |
|------------------------|-------------:|-----------------:|----------------------:|
| volume_render, total   |    8772.5 ms |        6171.4 ms |             2127.0 ms |
| volume_render, avg/spp |     34.27 ms |         24.11 ms |               8.31 ms |
| share of total runtime |        60.7% |            51.9% |                 27.1% |
| total render walltime  |   14461.7 ms |       11887.0 ms |             7862.8 ms |

Stage-on-stage:

- Register accumulation (Stage 1): -29.7% on volume_render and -17.8% on end-to-end walltime (2.57 s saved).
- Inference fast path (Stage 2): a further ~2.9x on volume_render (6171 ms -> 2127 ms) -- the per-sample outWs store and the accompanying host-side zero-init of the N-sized buffer were the single largest remaining source of global-memory traffic in the stage.
- Combined: ~4.1x on the volume_render stage and ~1.86x end-to-end (14.5 s -> 7.86 s) vs. the pre-optimization baseline. volume_render's share of total runtime dropped from 60.7% to 27.1%, handing the "top bottleneck" title to sample_trilinear (which is then addressed in #639).

Correctness is unchanged on the backward-aware path: the Python wrapper routes through the CUDA forward with needsBackward=true whenever any input requires grad, so saved tensors and the backward kernel are identical to before. The inference fast path is observationally equivalent for callers that consume only rgb / opacity.


fvdb/_volume_render.py
```diff
@@ -0,0 +1,217 @@
...
+        Consumes ``dL/drgb``, ``dL/ddepth``, ``dL/dopacity`` and
+        ``dL/dws`` from ``grad_outputs`` (the incoming gradient on
+        ``total_samples`` is unused — that output is an integer counter
+        and is intentionally non-differentiable). Returns a 6-tuple of
+        gradient slots matching :meth:`forward`'s positional arguments:
+        gradients flow only into ``sigmas`` and ``rgbs``; the slots for
+        ``delta_ts``, ``ts``, ``pack_info`` and ``transmittance_thresh``
+        are ``None``.
+        """
+        dL_drgb, dL_ddepth, dL_dopacity, dL_dws, _dL_dtotal = grad_outputs
```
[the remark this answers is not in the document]
swahtz: Good catch.  I've fixed that up and added tests for this

src/fvdb/detail/ops/VolumeRender.cu
```diff
@@ -42,6 +45,18 @@ volumeRenderFwdCallback(const TensorAccessor<scalar_t, 1> sigmas,
     const JOffsetsType sampleStartIdx = jOffsets[rayIdx];
     const JOffsetsType numRaySamples  = jOffsets[rayIdx + 1] - sampleStartIdx;
 
+    // Per-ray accumulators. Kept thread-local (and, for accRgb, ideally
+    // register-resident) so the inner loop does not perform a global-memory
+    // read-modify-write on outRGB / outDepth / outOpacity for every sample.
+    // Output tensors are zero-initialized on the host.
```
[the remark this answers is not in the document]
swahtz: fixed

tests/unit/test_ray_marching.py
```diff
@@ -594,7 +594,11 @@ def test_volume_render_total_samples_counts_terminating_sample(self):
         dt = torch.full((N,), 0.1, dtype=self.dtype, device=self.device)
         t = torch.arange(N, dtype=self.dtype, device=self.device) * 0.1
         # Single-channel rgbs keep the comparisons simple and are enough to
-        # exercise the accumulation path.
+        # exercise the accumulation path. Mark sigmas as requires_grad so the
+        # Python wrapper takes the backward-aware path and actually materializes
+        # outWs / outTotalSamples (the inference fast path returns size-0
+        # placeholders for those tensors).
+        sigmas.requires_grad_(True)
```
harrism, line 601: **💡 suggestion:** ‏I think there's a new fast-path that has requires_grad(False) that is not tested.

fvdb/_volume_render.py
```diff
@@ -0,0 +1,238 @@
...
+    Note:
+        Callers that need the full set of outputs under no-grad (for
+        example, a diagnostic that wants ``total_samples``) should either
+        run under :func:`torch.enable_grad` with at least one
+        differentiable input (typically ``sigmas.requires_grad_(True)``)
+        or call :func:`_fvdb_cpp.volume_render_fwd` directly with
+        ``needsBackward=True``.
+    """
+    needs_grad = torch.is_grad_enabled() and any(
+        isinstance(t, torch.Tensor) and t.requires_grad for t in (sigmas, rgbs, delta_ts, ts)
```
harrism: **💡 suggestion:** docs say gradients do not flow into ‏delta_ts and ts, so this can be tightened up.

```suggestion
    needs_grad = torch.is_grad_enabled() and any(
        isinstance(t, torch.Tensor) and t.requires_grad for t in (sigmas, rgbs)
```

src/fvdb/detail/ops/VolumeRender.cu
```diff
@@ -24,7 +24,10 @@ namespace {
 // allocation. Bump if a use case legitimately needs more channels.
 static constexpr int MAX_VOLUME_RENDER_CHANNELS = 16;
```
harrism: **🤓 nitpick (non-blocking):** ‏Inside an anon namespace, `static` is redundant.

```suggestion
constexpr int MAX_VOLUME_RENDER_CHANNELS = 16;
```

src/fvdb/FVDB.h
```diff
@@ -13,12 +13,96 @@ namespace fvdb {
...
+///                            ``ws`` and ``totalSamples`` (see note above).
+/// @return Tuple ``(rgb, depth, opacity, ws, totalSamples)``.
+std::tuple<torch::Tensor, torch::Tensor, torch::Tensor, torch::Tensor, torch::Tensor>
+volumeRenderForward(const torch::Tensor &sigmas,
+                    const torch::Tensor &rgbs,
+                    const torch::Tensor &deltaTs,
+                    const torch::Tensor &ts,
+                    const torch::Tensor &packInfo,
+                    double transmittanceThresh,
+                    bool needsBackward);
```
harrism, line 52: **💡 suggestion:** ‏This `needsBackward` has no default, but the detail op has `needsBackward = true`. Suggest either both have the same default, or neither has a default.

harrism (commented):
A few minor comments. Also Claude noticed: 

>  Tutorial doc is now silently inference-incompatible `docs/tutorials/volume_rendering.md` invokes `volume_render` and uses `depth` downstream. Worth either:
  - A one-line comment near line 403 noting that depth is only populated when at least one of ray_density/ray_color has requires_grad=True, or
  - Updating the tutorial's volume_render_func to assert/raise if it's invoked without grad and depth is needed.

fvdb/_volume_render.py
swahtz: True and I've cleaned up comments elsewhere to clarify the gradients of ts, delta_ts

src/fvdb/FVDB.h
swahtz: I think neither should have a default to make the choice explicit.  I'll remove the default in the detail op

tests/unit/test_ray_marching.py
swahtz: This is true, I've added a new `test_volume_render_inference_fast_path` function that exhaustively tests this path and the assumptions laid out in the docstrings.

fvdb/_volume_render.py
```diff
@@ -0,0 +1,242 @@
...
+
+    Note:
+        Callers that need the full set of outputs under no-grad (for
+        example, a diagnostic that wants ``total_samples``) should
+        either run under :func:`torch.enable_grad` with
+        ``sigmas.requires_grad_(True)`` or ``rgbs.requires_grad_(True)``,
+        or call :func:`_fvdb_cpp.volume_render_fwd` directly with
+        ``needsBackward=True``. Setting ``requires_grad`` on
+        ``delta_ts`` or ``ts`` does not select this path because no
+        gradient is propagated into those inputs.
```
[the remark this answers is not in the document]
swahtz: I think this comment might be about an older commit because this is correct

harrism (approved):

PR #641, opened 2026-04-24
NanoVDB loading: fix mixed grid type loading and add read_metadata API

swahtz:
## Problem

Two bugs surfaced when trying to read real-world `.nvdb` files in fVDB:

1. **Wrong-type dispatch for multi-grid handles.** `detail::io::loadOneGrid` calls `handle.gridType()` and `handle.gridMetaData()` without the `gridId` argument, so every grid in a multi-grid `GridHandle` is dispatched as if it were grid 0. Loading e.g. `{Float, Float, Vec3f}` therefore fails at grid 2 with `Grid has type Vec3f, but expected float`.

2. **Cryptic concat error for heterogeneous per-voxel shapes.** Even when dispatch is correct, `JaggedTensor` packs per-grid data along dim 0 via `torch::cat`, which requires matching dtype and matching trailing sizes. A file that contains a `Float` grid at `[N, 1]` alongside a `Vec3f` grid at `[M, 3]` bubbles up as `Sizes of tensors must match except in dimension 0. Expected size 1 but got size 3 for tensor number 2...` with no hint at which grids are actually incompatible or how to load them.

A third papercut: `.nvdb` files saved with ZIP compression fail to open because `NANOVDB_USE_ZIP` was never set at build time.

### C++ (`src/fvdb`)

- **`detail/io/LoadNanovdb.cpp` — `loadOneGrid`**: thread `gridId` through every `handle.gridType(...)` / `handle.gridMetaData(...)` call so per-grid dispatch is actually per-grid.
- **`detail/io/LoadNanovdb.cpp` — new `validateTrailingShapeAndDtype` helper**: runs in `fromNVDB` before `JaggedTensor` construction and raises a `TORCH_CHECK_VALUE` that names both the reference grid and the offending grid, prints their shapes/dtypes, and tells the caller to use `name=` / `index=` / `load_nanovdb_single` to load them one at a time. No perf impact on the happy path (`data.size() <= 1` early-out).
- **New `NanoVDBGridMetadata` struct** (`src/fvdb/NanoVDBGridMetadata.h`): POD with `name`, `type`, `gridClass`, `voxelCount`, `voxelSize`, `indexBBoxMin`, `indexBBoxMax`.
- **New `fvdb::read_metadata(path)`** (`src/fvdb/FVDB.{h,cpp}`): public C++ entry point backed by a new `detail::io::readNVDBMetaData` wrapping `nanovdb::io::readGridMetaData`. File-header-only; no voxel data is materialized.

### Python bindings (`src/python/Bindings.cpp`)

- Bound `NanoVDBGridMetadata` as a Python class with read-only accessors and a compact `__repr__`.
- Bound `fvdb.read_metadata` as a free function taking `path`.

### Python API (`fvdb/functional`)

- `_io.py`: added `read_nanovdb_metadata(path)` and a `grid_names_in_nanovdb(path)` convenience wrapper; both with docstrings and a short usage example.
- `__init__.py`: added both names to `__all__`. Reordered the existing `from ._xxx import ...` blocks into alphabetical order while I was in there.

### Build (`src/CMakeLists.txt`)

- `find_package(ZLIB REQUIRED)`, `-DNANOVDB_USE_ZIP`, and `ZLIB::ZLIB` in `target_link_libraries` so NanoVDB's ZIP-compressed `.nvdb` reader is actually compiled in.

### Tests (`tests/unit`)

- `__init__.py`: bump `set_testing_git_tag` to the fvdb-test-data commit that ships the new `io/mixed_types.nvdb` fixture (`46e7bd4...`).
- `test_io.py`: new `TestLoadMixedTypeNanovdb` class driven by a single `_MIXED_TYPES_GRIDS` truth table covering all four grids in the fixture (`density` float, `pressure` double, `label` int32, `velocity` Vec3f). 19 tests total covering:
  - `read_nanovdb_metadata` enumerates names / types / voxel counts.
  - `grid_names_in_nanovdb` returns the right names.
  - Loading the whole file without a selector raises `ValueError`, the message is lowercase-matched against `"mixed value type"`, references at least one grid by name, and contains an actionable hint (`name=` / `index=` / `load_nanovdb_single`).
  - By-name load for each of the 4 types on CPU and CUDA, with round-trip value verification.
  - By-index load for each of the 4 types on CPU and CUDA.


fvdb/functional/_io.py
```diff
@@ -132,6 +132,47 @@ def load_nanovdb(
...
+        ``name``, ``type`` (e.g. ``"float"``, ``"Vec3f"``, ``"OnIndex"``), ``grid_class``,
+        ``voxel_count``, ``voxel_size``, ``index_bbox_min`` and ``index_bbox_max``.
+
+    Example:
+        >>> for m in read_nanovdb_metadata("explosion.nvdb"):
+        ...     print(m.name, m.type, m.voxel_count)
+    """
+    from .._fvdb_cpp import read_metadata as _read_metadata
+
+    return _read_metadata(path)
```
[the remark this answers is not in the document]
swahtz: Fixed up

src/python/Bindings.cpp
```diff
@@ -372,6 +372,37 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
...
+                               })
+        .def_property_readonly("index_bbox_max",
+                               [](const fvdb::NanoVDBGridMetadata &m) {
+                                   return std::vector<int32_t>{
+                                       m.indexBBoxMax[0], m.indexBBoxMax[1], m.indexBBoxMax[2]};
+                               })
+        .def("__repr__", [](const fvdb::NanoVDBGridMetadata &m) {
+            return "NanoVDBGridMetadata(name='" + m.name + "', type='" + m.type +
+                   "', grid_class='" + m.gridClass +
+                   "', voxel_count=" + std::to_string(m.voxelCount) + ")";
```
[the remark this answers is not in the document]
swahtz: Fixed

fvdb/functional/_io.py
```diff
@@ -8,13 +8,13 @@
...
 import torch
 
-from .. import _fvdb_cpp
 from ..jagged_tensor import JaggedTensor
 from ..types import DeviceIdentifier, resolve_device
 
 if TYPE_CHECKING:
     from ..grid import Grid
     from ..grid_batch import GridBatch
+    from ..nanovdb_grid_metadata import NanoVDBGridMetadata
```
harrism: **Blocker — broken TYPE_CHECKING import.**

There is no `fvdb/nanovdb_grid_metadata.py` — the class is bound only inside `fvdb._fvdb_cpp`. mypy/pyright will fail to resolve the annotation `list[NanoVDBGridMetadata]` on `read_nanovdb_metadata` below. Runtime is fine because `from __future__ import annotations` makes the annotation lazy, but `inspect.get_type_hints()`, IDE hover, and stubgen all break.

Change to:
```python
from .._fvdb_cpp import NanoVDBGridMetadata
```
(and add the matching stub in `_fvdb_cpp.pyi` — see comment on `Bindings.cpp`).

src/python/Bindings.cpp
```diff
@@ -372,6 +372,39 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
...
+        .def("__repr__", [](const fvdb::NanoVDBGridMetadata &meta) {
+            return "NanoVDBGridMetadata(name='" + meta.name + "', type='" + meta.type +
+                   "', grid_class='" + meta.gridClass +
+                   "', voxel_count=" + std::to_string(meta.voxelCount) + ")";
+        });
+
+    m.def("read_metadata",
+          &fvdb::read_metadata,
+          py::arg("path"),
+          "Read per-grid metadata from a .nvdb file without loading voxel data.");
```
harrism, line 424: **Blocker — missing stubs in `fvdb/_fvdb_cpp.pyi`.**

`NanoVDBGridMetadata` (the class above) and `read_metadata` (this function) are bound here but not added to `_fvdb_cpp.pyi`. Type-stub consumers (mypy, pyright, IDE autocomplete) will see them as missing/untyped, and the `TYPE_CHECKING` import in `fvdb/functional/_io.py` has no resolvable target. Please add:

```python
class NanoVDBGridMetadata:
    @property
    def name(self) -> str: ...
    @property
    def type(self) -> str: ...
    @property
    def grid_class(self) -> str: ...
    @property
    def voxel_count(self) -> int: ...
    @property
    def voxel_size(self) -> tuple[float, float, float]: ...
    @property
    def index_bbox_min(self) -> tuple[int, int, int]: ...
    @property
    def index_bbox_max(self) -> tuple[int, int, int]: ...
    def __repr__(self) -> str: ...

def read_metadata(path: str) -> list[NanoVDBGridMetadata]: ...
```
(Tuple types match the suggested `def_readonly` change below; if you keep the lambda-and-vector approach instead, use `list[float]` / `list[int]`.)

src/fvdb/NanoVDBGridMetadata.h
```diff
@@ -0,0 +1,27 @@
...
+#include <string>
+
+namespace fvdb {
+
+/// @brief Lightweight per-grid metadata read from a .nvdb file header without loading any voxel
+///        data. Useful for enumerating grids (names / types) before deciding which ones to load.
+struct NanoVDBGridMetadata {
+    std::string name;      ///< Grid name as stored in the file
+    std::string type;      ///< nanovdb::GridType as a string (e.g. "float", "Vec3f", "OnIndex")
+    std::string gridClass; ///< nanovdb::GridClass as a string (e.g. "LS", "FOG", "TensorGrid")
```
harrism: **Doc bug that will mislead callers.**

The actual strings produced by `nanovdb::toStr(GridClass)` (NanoVDB.h) are `"?"`, `"SDF"`, `"FOG"`, `"MAC"`, `"PNTIDX"`, `"PNTDAT"`, `"TOPO"`, `"VOX"`, `"INDEX"`, `"TENSOR"`. Two of the three examples in this comment are wrong:

- `"LS"` → should be `"SDF"` (LevelSet)
- `"TensorGrid"` → should be `"TENSOR"`

Anyone writing `if meta.grid_class == "TensorGrid"` will silently never match. Suggest:

```cpp
std::string gridClass; ///< nanovdb::GridClass as a string (e.g. "SDF", "FOG", "TENSOR", "?")
```

Please mirror correct examples in the Python docstring on `read_nanovdb_metadata` so the user-facing API doc is accurate on both sides.

src/fvdb/detail/io/LoadNanovdb.cpp
```diff
@@ -528,6 +597,8 @@ fromNVDB(nanovdb::GridHandle<nanovdb::HostBuffer> &handle,
     nanovdb::GridHandle<TorchDeviceBuffer> resCpu = nanovdb::mergeGrids(grids);
     auto ret = makeGridBatchData(std::move(resCpu), voxSizes, voxOrigins);
 
+    validateTrailingShapeAndDtype(data, names);
```
harrism: **Hoist validation before `mergeGrids`.**

`mergeGrids` allocates and copies all grid bytes, and `makeGridBatchData` builds the GridBatch — on the heterogeneous-type error path we throw both away. The validator only inspects `data` and `names`, so it can run immediately after the per-grid load loop, before line 597. Same fix applies at L644-647 in the multi-handle overload. Happy path is unchanged; error path becomes much cheaper and the code reads more naturally (validate, then build).

src/python/Bindings.cpp
```diff
@@ -372,6 +372,39 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
...
+                                   return std::vector<int32_t>{meta.indexBBoxMin[0],
+                                                               meta.indexBBoxMin[1],
+                                                               meta.indexBBoxMin[2]};
+                               })
+        .def_property_readonly("index_bbox_max",
+                               [](const fvdb::NanoVDBGridMetadata &meta) {
+                                   return std::vector<int32_t>{meta.indexBBoxMax[0],
+                                                               meta.indexBBoxMax[1],
+                                                               meta.indexBBoxMax[2]};
+                               })
```
harrism, line 414: **Use `def_readonly` directly — `pybind11/stl.h` already auto-converts `std::array<T, N>`.**

`pybind11/stl.h` is included at the top of this file, so `std::array<double, 3>` and `std::array<int32_t, 3>` already convert to a Python `tuple` automatically. The lambdas-and-`std::vector<>` approach allocates a fresh list per access and gives callers a mutable `list` instead of a `tuple` — semantically odd for a fixed-size coordinate triple. Suggest:

```cpp
.def_readonly("voxel_size", &fvdb::NanoVDBGridMetadata::voxelSize)
.def_readonly("index_bbox_min", &fvdb::NanoVDBGridMetadata::indexBBoxMin)
.def_readonly("index_bbox_max", &fvdb::NanoVDBGridMetadata::indexBBoxMax)
```

Less code, zero per-access allocation, immutable on the Python side. Update the `.pyi` stubs to `tuple[float, float, float]` / `tuple[int, int, int]` to match.

tests/unit/test_io.py
```diff
@@ -483,5 +484,119 @@ def test_voxelsize_preserved_with_different_scales_per_grid(self, device):
...
+
+    def test_read_metadata_enumerates_all_grids(self):
+        meta = fvdb.functional.read_nanovdb_metadata(self._fixture_path())
+        self.assertEqual(len(meta), len(_MIXED_TYPES_GRIDS))
+
+        by_name = {m.name: m for m in meta}
+        for _, name, type_str, _, _, _ in _MIXED_TYPES_GRIDS:
+            self.assertIn(name, by_name)
+            self.assertEqual(by_name[name].type, type_str, f"unexpected type for grid {name!r}")
+            self.assertEqual(by_name[name].voxel_count, 1, f"unexpected voxel_count for grid {name!r}")
```
harrism: **Incomplete metadata assertions.**

This test checks `name`, `type`, `voxel_count`. The PR also exposes `grid_class`, `voxel_size`, `index_bbox_min`, `index_bbox_max` — none of which are tested. `grid_class` in particular is the field where the doc-comment bug above (LS / TensorGrid) could otherwise have shipped without notice. Worth at least:

```python
self.assertEqual(by_name["density"].grid_class, "FOG")
self.assertEqual(tuple(by_name["density"].voxel_size), (1.0, 1.0, 1.0))
```
(or whatever the fixture was actually authored with — also worth asserting `index_bbox_min`/`max` for at least one grid).

tests/unit/test_io.py
```diff
@@ -483,5 +484,119 @@ def test_voxelsize_preserved_with_different_scales_per_grid(self, device):
...
+    @parameterized.expand(_MIXED_TYPES_BY_INDEX_PARAMS)
+    def test_load_single_grid_by_index(self, device, grid_spec):
+        index, expected_name, expected_dtype, trailing = grid_spec
+        _, data, names = fvdb.GridBatch.from_nanovdb(self._fixture_path(), index=index, device=device)
+
+        self.assertEqual(names, [expected_name])
+        self.assertEqual(data.jdata.dtype, expected_dtype)
+        self.assertEqual(data.jdata.dim(), 2)
+        self.assertEqual(data.jdata.size(0), 1)
+        self.assertEqual(data.jdata.size(-1), trailing)
```
harrism, line 614: **CUDA assertion gap.**

Both parameterized tests run with `device="cuda"` but never assert that the loaded data ended up on CUDA. If `device=` were ever silently dropped on the load path, these tests would still pass on CUDA. One line per test:

```python
self.assertEqual(data.jdata.device.type, device.split(":")[0])
```

harrism, line 614: **By-index test should also verify the round-tripped value.**

The by-name variant above asserts the actual voxel value, which is what proves the `gridId` plumb-through actually works for non-`Float` types. The by-index variant only checks dtype/shape — so a regression where, say, grid 0's value got returned for `index=2` would slip past this test. Mirror the value check from `test_load_single_grid_by_name`:

```python
expected_value = next(value for idx, _, _, _, _, value in _MIXED_TYPES_GRIDS if idx == index)
row = data.jdata[0].cpu().tolist()
if trailing == 1:
    self.assertAlmostEqual(row[0], expected_value, places=5)
else:
    for got, want in zip(row, expected_value):
        self.assertAlmostEqual(got, want, places=5)
```

fvdb/functional/_io.py
```diff
@@ -132,6 +132,47 @@ def load_nanovdb(
 # ---------------------------------------------------------------------------
 
 
+def read_nanovdb_metadata(path: str) -> list[NanoVDBGridMetadata]:
```
harrism, line 135: **Section divider in the wrong place.**

This function and `grid_names_in_nanovdb` below it are read-only metadata operations, but they sit under the `# Save` divider at L130-132 — `save_nanovdb` is the next thing after them. Either move them above the `# Save` divider (with their own `# Metadata` divider) or insert them between `load_nanovdb` and `# Save`.

harrism (changes requested):
Re-reviewed in detail. The two bug fixes (per-grid type dispatch and the actionable mixed-type error) are clearly correct and well-tested for the load path; the metadata API is a clean addition.

A few things to address before merge — comments are inline. Summary by severity:

**Blockers**
- `fvdb/functional/_io.py:17` — `TYPE_CHECKING` import points to a non-existent module; static type-check / IDE tooling will fail to resolve `NanoVDBGridMetadata`.
- `fvdb/_fvdb_cpp.pyi` — new bindings (`NanoVDBGridMetadata`, `read_metadata`) are not added to the type stubs.

**Doc bugs that will mislead callers**
- `src/fvdb/NanoVDBGridMetadata.h:18` — example `grid_class` strings (`"LS"`, `"TensorGrid"`) don't match what `nanovdb::toStr(GridClass)` actually returns (`"SDF"`, `"TENSOR"`).

**Structural / readability**
- `src/fvdb/detail/io/LoadNanovdb.cpp` — `validateTrailingShapeAndDtype` runs after `mergeGrids` + `makeGridBatchData`; hoisting it before saves all that work on the error path.
- `src/python/Bindings.cpp` — `def_property_readonly` lambdas are unnecessary; `pybind11/stl.h` already auto-converts `std::array<T, N>` to a Python tuple via `def_readonly`.
- `fvdb/functional/_io.py` — new metadata functions sit under the `# Save` section divider.

**Test coverage gaps**
- `grid_class` / `voxel_size` / `index_bbox_*` are exposed but not asserted (and would have caught the `grid_class` doc bug above).
- CUDA-parameterized loads don't assert `data.jdata.device.type`, so a silent `device=` regression would pass.
- The by-index load test only checks shape/dtype, not the round-tripped value — the by-name test does, and the same proof matters for the index path.

**Adjacent (not blocking this PR)**
- Filed #642 for a pre-existing latent bug in `nanovdbTensorGridToFVDBGrid` (always reads `blindMetaData(0)` inside the loop). Not introduced here, but the per-grid dispatch fix in this PR makes the bug more reachable for multi-grid TensorGrid files.

(Review by Claude Opus 4.7 with guidance from @harrism).

fvdb/functional/_io.py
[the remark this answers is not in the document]
swahtz: Yes, late night mistake, thanks for catching.

tests/unit/test_io.py
swahtz: Done

swahtz: Addressed

src/python/Bindings.cpp
swahtz: Looking more into it, apparently this approach would still make these properties become Python lists, not tuples.  This is a function of the type caster in pybind and not how the binding is declared.  I'll update these properties to be tuples with a `std::make_tuple` so we get the correct cast to a Python tuple.  This also make me look closer at the types of the C++ members and I realize they should be nanovdb::Vec/Coord types rather than just std::array's to be consistent (and more useful for future functionality)

src/python/Bindings.cpp
swahtz: Done

src/fvdb/NanoVDBGridMetadata.h
swahtz: Fixed

tests/unit/test_io.py
swahtz: Added checks for all the values in the fixture.

fvdb/functional/_io.py
swahtz: Fixed this up, was a late evening change I made erroneously

fvdb/functional/_io.py
swahtz: Fixed

src/fvdb/detail/io/LoadNanovdb.cpp
swahtz: Very true, fixed up

fvdb/__init__.pyi
```diff
@@ -17,7 +14,7 @@ def _parse_device_string(device_string: str | torch.device) -> torch.device: ...
 # The following import needs to come after the GridBatch and JaggedTensor imports
 # immediately above in order to avoid a circular dependency error.
 from . import nn, utils, viz
-from ._fvdb_cpp import config, hilbert, morton, volume_render
+from ._fvdb_cpp import NanoVDBGridMetadata, config, hilbert, morton, volume_render
 from .attention import scaled_dot_product_attention
```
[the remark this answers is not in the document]
swahtz: Fixed

src/CMakeLists.txt
```diff
@@ -24,6 +24,9 @@ include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/get_cutlass.cmake)
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/get_nanovdb.cmake)
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/get_blosc.cmake)
 
+# zlib is required for NANOVDB_USE_ZIP (reading ZIP-compressed .nvdb files).
+find_package(ZLIB REQUIRED)
+
```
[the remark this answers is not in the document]
swahtz: I think we should just always have zip support for being able to read nanos written with zip

src/fvdb/detail/io/LoadNanovdb.cpp
```diff
@@ -495,11 +498,77 @@ loadOneGrid(const nanovdb::GridHandle<nanovdb::HostBuffer> &handle, uint32_t gri
...
+        }
+        const bool dtypeOk = t.dtype() == ref.dtype();
+        if (!shapeOk || !dtypeOk) {
+            TORCH_CHECK_VALUE(
+                false,
+                "Cannot load grids of mixed value type into a single GridBatch/JaggedTensor. ",
+                "Grid 0 (\"",
+                nameAt(0),
+                "\") has data ",
+                formatShapeAndDtype(ref),
```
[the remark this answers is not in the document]
swahtz: True.  I've updated the message

harrism (approved):

PR #645, opened 2026-04-27
ci(nightly): anchor nightly version to upcoming release in pyproject.toml

swahtz:
Previously the nightly publish workflow hard-coded the base version as 0.0.0, producing wheels like 0.0.0.dev20260427+pt280.cu128. This made nightlies sort below every real release and gave no hint of which release cycle they belonged to.

Read the current `version` from pyproject.toml instead, strip any trailing pre/dev/post segment and local label, and use the resulting base release as the anchor. With pyproject.toml at 0.5.0.dev0 today's nightly becomes 0.5.0.dev20260427+pt280.cu128, which under PEP 440 sorts as:

    0.5.0.dev0  <  0.5.0.devYYYYMMDD+ptX.cuY  <  0.5.0

so `pip install --pre fvdb-core` tracks nightlies until 0.5.0 ships, prefers the final release once tagged, and resumes tracking nightlies after the next dev-cycle bump. Fails fast if the version string can't be parsed.

### DIFF
--- a/.github/workflows/nightly-publish.yml
+++ b/.github/workflows/nightly-publish.yml
@@ -157,7 +157,16 @@ jobs:
         run: |
           TORCH_TAG="$(echo "${{ matrix.torch-version }}" | tr -d '.')"
           CUDA_TAG="$(echo "${{ matrix.cuda-version }}" | tr -d '.')"
-          NIGHTLY_VERSION="0.0.0.dev$(date -u '+%Y%m%d')+pt${TORCH_TAG}.cu${CUDA_TAG}"
+          # Anchor the nightly at the upcoming release recorded in pyproject.toml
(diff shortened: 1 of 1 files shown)


swahtz:
@phapalova I was thinking of making the change to make our nightly build versions make more sense.  This would make depending on dev/yet-to-be-released version numbers easier for downstream users… you wouldn't have the confusion of needing '0.0.0'(nightly number) _OR_ '0.5.0' (unreleased dev version).  Is that an improvement for your needs?

phapalova:
> @phapalova I was thinking of making the change to make our nightly build versions make more sense. This would make depending on dev/yet-to-be-released version numbers easier for downstream users… you wouldn't have the confusion of needing '0.0.0'(nightly number) _OR_ '0.5.0' (unreleased dev version). Is that an improvement for your needs?

Thanks that will simplify the way we pull the nightly

phapalova (approved):

PR #650, opened 2026-05-10
`saveNVDB` Optimizations

swahtz:
## Summary

Move grid construction in `saveNVDB` / `to_nanovdb` off the CPU `setValue` rebuild and onto a per-residency fast path. CUDA-resident `GridBatchData` is now built on the GPU via `nanovdb::tools::cuda::indexToGrid`. CPU-resident `GridBatchData` is built via a new host-only port that mirrors `indexToGrid` node-for-node and never touches the GPU. Both paths produce a bit-identical typed grid with the same `fvdb_jdata` blind metadata as before. Also unblocks BLOSC-compressed writes after moving the file to a `.cu` translation unit.

## Motivation

The previous `SaveNanoVDB.cpp` was the dominant CPU cost when calling `fvdb.save(...)` on typed grids: every batch rebuilt the NanoVDB grid on the host through the `nanovdb::tools::build::Grid` proxy, which inserts each active voxel via a hashtable-backed `setValue(ijk, value)` and then re-walks the proxy in `CreateNanoGrid` to lay out the final tree. Profiling on a representative `Vec3f` grid pinned that rebuild plus a few avoidable copies as the bulk of the time. Two sub-issues showed up while addressing it:

1. `std::vector<HostGridHandle> buffers(batchSize)` default-constructed all slots up front, so the existing `batchSize == 1` early return never fired.
2. Per-batch `data.index({bi})` calls re-derived the same shape metadata inside the inner loop.
3. `JaggedTensor` and `std::vector<std::string>` were passed by value across `SaveNanoVDB.{h,cpp}`, `FVDB.{h,cpp}`, and `python/Bindings.cpp`.
4. `getIndexGrid` issued `cudaMemcpyAsync` followed immediately by `cudaStreamSynchronize`, which is just a more expensive synchronous copy.

An earlier iteration of this PR routed *both* CPU- and CUDA-resident inputs through `nanovdb::tools::cuda::indexToGrid`. That regressed CPU-only workloads (every save paid CUDA context init, H2D upload of the grid + values, kernel launch, and D2H copy). The current PR keeps the GPU path for CUDA inputs and adds a host-only port for CPU inputs.

### Core: per-residency grid construction

- Rename `src/fvdb/detail/io/SaveNanoVDB.cpp` to `SaveNanoVDB.cu` and update `src/CMakeLists.txt` accordingly.
- New `indexToGridHost<DstBuildT>` host helper that mirrors `nanovdb::tools::cuda::indexToGrid` sequentially: computes the destination buffer layout from the source `NanoGrid<ValueOnIndex>`'s node counts, allocates a single `nanovdb::HostBuffer`, copies the Grid/Tree/Root headers (with `mGridType` patched), walks root tiles, upper internal nodes, lower internal nodes, and leaves in order, fixes child offsets when source/dest node sizes differ (matching the GPU kernel's `nodeSkip * sizeof()` arithmetic), and translates per-slot index lookups via `srcValues[srcSlot]`. Calls `nanovdb::tools::updateChecksum` at the end.
- New `fvdbToNanovdbGridWithValuesHost` per-batch wrapper that:
  - copies `data` to CPU + contiguous if needed,
  - reuses a single `std::vector<ValueT>` values buffer of size `numVoxels + 1` across batches,
  - leaves slot `[0]` zero (matches `indexToGrid`'s background convention) and `memcpy`s slots `[1..N]` from the user data slice,
  - calls `indexToGridHost`, then `patchGridWithBlindShape` to append the `fvdb_jdata` blind metadata record (rdim + tailSizes),
  - returns the size-1 single grid directly or `mergeGrids`-merges a multi-grid batch.
- Existing GPU path (`fvdbToNanovdbGridWithValues` CUDA branch) is rebuilt on top of `nanovdb::tools::cuda::indexToGrid`. CUDA-resident `GridBatchData` is fed straight to the kernel; per-batch `indexToGrid` calls and D2H copies are enqueued on a single CUDA stream with one final `cudaStreamSynchronize`. Per-batch values buffer is allocated as `nanovdb::cuda::DeviceBuffer` of size `numVoxels + 1` with `[0]` zero-initialised and `[1..N]` D2D-copied from `cudaData`.
- `fvdbToNanovdbGridWithValues` dispatches on `gridBatchData.device().is_cuda()`: false → host wrapper (zero GPU touch), true → existing GPU wrapper.
- `Fp16` grids no longer use the typed `indexToGrid` path on either side (`BuildToValueMap<Fp16>::type` is `float`, and the kernel rejects the quantized leaf layout). They fall through to `saveIndexGridWithBlindData` with an `fvdb_jdataHalf` tag so `c10::Half` data is preserved exactly via blind metadata.
- New `checkPerBatchVoxelCounts` validates that `data.joffsets` per-batch row counts match `GridBatchData::numVoxelsAt(bi)` before any device indexing, preventing OOB access when the global `totalVoxels` check happens to pass with mismatched per-batch offsets.

### Cleanups in the same area

- Replace `std::vector<HostGridHandle> buffers(batchSize)` with `reserve(batchSize)` so the existing `batchSize == 1` early return actually fires.
- Hoist `data.index({bi})` shape reads (`rdim`, `tailSizes`) out of the per-batch loops in `fvdbToNanovdbGridWithValues` and `saveIndexGridWithBlindData`.
- Pass `std::optional<JaggedTensor>` and `std::vector<std::string>` by `const&` in `SaveNanoVDB.{h,cu}`, `FVDB.{h,cpp}`, and `python/Bindings.cpp`.
- Replace `cudaMemcpyAsync` immediately followed by `cudaStreamSynchronize` in `getIndexGrid` with a single `cudaMemcpy` under a `CUDAGuard`.

### Build

- Add `-DNANOVDB_USE_BLOSC` and `-DNANOVDB_USE_ZIP` to the `COMPILE_LANGUAGE:CUDA` block in `src/CMakeLists.txt`. These were previously only set for CXX, so after moving `SaveNanoVDB` to a `.cu` file NanoVDB's `IO.h` selected its "BLOSC disabled" path and BLOSC-compressed writes failed at runtime.

## Files changed

- `src/CMakeLists.txt`
- `src/fvdb/FVDB.cpp`, `src/fvdb/FVDB.h`
- `src/fvdb/detail/io/SaveNanoVDB.cpp` (deleted)
- `src/fvdb/detail/io/SaveNanoVDB.cu` (new)
- `src/fvdb/detail/io/SaveNanoVDB.h`
- `src/python/Bindings.cpp`

## Test plan

```sh
./build.sh install
cd tests
pytest unit/test_io.py -v
pytest unit -q
```

- [x] `tests/unit/test_io.py`: 622 passed, 0 failed.
- [x] BLOSC-compressed save/load tests pass (no `BLOSC compression codec was disabled during build` errors).
- [x] CPU-resident `GridBatchData` save path is bit-identical to the pre-branch host rebuild via the existing roundtrip equality assertions.
- [x] CUDA-resident `GridBatchData` save path exercised on `float`, `Vec3f`, `double`, `Vec3d`, plus blind-data roundtrip for `Fp16`.

## Performance

Measured on this machine, `save_nanovdb(grid, "/dev/null", data)` with a clustered Vec3f batch (real-workload-like, not random ijk):

| Workload | Pre-branch | This PR | Speedup |
|---|---|---|---|
| CPU 100K voxels, batch=2 | 11.66 ms | 5.51 ms | 2.1x |
| CPU 40K voxels, batch=8 | 17.29 ms | 1.25 ms | 13.8x |
| CUDA 100K voxels, batch=2 | 7.54 ms | 1.38 ms | 5.5x |
| CUDA 40K voxels, batch=8 | 17.58 ms | 4.04 ms | 4.4x |

For the user-reported "CPU + few large grids + tight loop" workload (~100K voxels, batch=2), this PR is now ~2x faster than the pre-branch baseline. An interim version of this branch that routed CPU inputs through the GPU was ~50x slower than the pre-branch baseline; the host fast-path in this PR is what makes the CPU case correct and faster.

## Notes / risks

- Roundtripping a typed grid that was saved with either path produces the same coordinates and the same per-voxel values as before; verified by `test_save_and_load_gridbatch` and `test_io.py`'s exact-equality / set-equality assertions for CPU and CUDA loads respectively.
- `indexToGrid`'s 1-based active-voxel indexing is handled internally by the per-batch values buffer (slot 0 = background, slots 1..N = active values) on both paths; callers see the same data layout they had before.
- `indexToGridHost` is a manual port of `nanovdb::tools::cuda::indexToGrid`. If NanoVDB upgrades change the kernel's per-node offset arithmetic, the host port must be updated to match. The two paths share the same `BuildToValueMap` and stats predicates so any divergence should surface immediately in `test_io.py`.
- `Fp16` no longer uses NanoVDB's quantized leaf encoding on save. It is now stored as `c10::Half` blind data tagged `fvdb_jdataHalf`, which is bit-exact on roundtrip but is a format change for any external consumer that was relying on the previous Fp16 leaf layout. Roundtrip via `fvdb` itself is unchanged.
- `checkPerBatchVoxelCounts` is a stricter validation than the previous global-only check; it can now reject inputs that previously crashed with a CUDA OOB instead of failing fast at the API boundary.


src/fvdb/detail/io/SaveNanoVDB.cu
```diff
@@ -0,0 +1,705 @@
...
+        names.size() == 0 || names.size() == (size_t)gridBatchData.batchSize(),
+        "Invalid parameter for names, must be empty or a list of the same length as the batch size. Got " +
+            std::to_string(names.size()) + " names for batch size " +
+            std::to_string(gridBatchData.batchSize()));
+
+    using HostGridHandle   = nanovdb::GridHandle<nanovdb::HostBuffer>;
+    using DeviceGridHandle = nanovdb::GridHandle<nanovdb::cuda::DeviceBuffer>;
+    using ValueT           = typename nanovdb::BuildToValueMap<OutBuildT>::type;
+
+    // Hoist tensor shape info out of the per-batch loop (issue #4). The data tensor has shape
```
harrism: **☑️ todo:** ‏
Remove unhelpful reference to nonexistent issue.

```suggestion
    // Hoist tensor shape info out of the per-batch loop. The data tensor has shape
```

src/fvdb/detail/io/SaveNanoVDB.cu
```diff
@@ -0,0 +1,705 @@
...
+
+    // Wait for all kernels and D2H copies to land before patching host-side metadata.
+    cudaCheck(cudaStreamSynchronize(stream.stream()));
+    deviceHandles.clear();     // free GPU buffers as soon as possible
+    perBatchValueBufs.clear(); // ditto
+
+    // Second pass: patch host-side GridData (mGridSize, mMap, name) and append fvdb_jdata blind
+    // data, then wrap the buffer in a HostBuffer-backed GridHandle.
+    std::vector<HostGridHandle> buffers;
+    buffers.reserve(gridBatchData.batchSize()); // issue #2: reserve, don't size
```
harrism:
```suggestion
    buffers.reserve(gridBatchData.batchSize()); 
```

src/fvdb/detail/io/SaveNanoVDB.cu
```diff
@@ -0,0 +1,705 @@
...
+
+    // Get pointer to grid read from (possibly on the device)
+    const bool isCuda = nanoGridHdl.buffer().device().is_cuda();
+    void *readHead    = isCuda ? nanoGridHdl.buffer().deviceData() : nanoGridHdl.buffer().data();
+    const size_t sourceGridByteSize = nanoGridHdl.buffer().size();
+
+    // Issue #6: collapse the async-then-immediate-sync pattern. The previous code launched a
+    // single cudaMemcpyAsync and then immediately synchronized, which is equivalent to a plain
+    // cudaMemcpy with extra noise. Use a synchronous copy under the device guard for clarity.
+    if (isCuda) {
```
harrism: **💡 suggestion:** Remove ‏excessive Claude-commenting.

```suggestion
    if (isCuda) {
```

src/fvdb/detail/io/SaveNanoVDB.cu
```diff
@@ -0,0 +1,705 @@
...
+        const size_t sourceGridByteSize = nanoGridHdl.gridSize(bi);
+        if (isCuda) {
+            c10::cuda::CUDAGuard deviceGuard(gridBatchData.device());
+            at::cuda::CUDAStream defaultStream =
+                at::cuda::getCurrentCUDAStream(gridBatchData.device().index());
+            cudaMemcpyAsync((void *)writeHead,
+                            (void *)readHead,
+                            sourceGridByteSize,
+                            cudaMemcpyDeviceToHost,
+                            defaultStream.stream());
```
harrism: **🤓 nitpick (non-blocking):** ‏I wouldn't call this `defaultStream` since it's not necessarily the CUDA default stream, which has a specific meaning.

```suggestion
            at::cuda::CUDAStream stream =
                at::cuda::getCurrentCUDAStream(gridBatchData.device().index());
            cudaMemcpyAsync((void *)writeHead,
                            (void *)readHead,
                            sourceGridByteSize,
                            cudaMemcpyDeviceToHost,
                            stream.stream());
```

src/fvdb/detail/io/SaveNanoVDB.cu
```diff
@@ -0,0 +1,705 @@
...
+        memcpy((void *)writeHead, (const void *)srcBytes, jdataSize);
+        writeHead += jdataSize;
+        writeHead += blindDataPadding[bi]; // Add padding to make sure we're 32 byte aligned
+    }
+
+    // Synchronize cuda stream if we just did a bunch of GPU -> CPU transfers
+    if (isCuda) {
+        at::cuda::CUDAStream defaultStream =
+            at::cuda::getCurrentCUDAStream(gridBatchData.device().index());
+        cudaStreamSynchronize(defaultStream.stream());
```
harrism:
```suggestion
        at::cuda::CUDAStream stream =
            at::cuda::getCurrentCUDAStream(gridBatchData.device().index());
        cudaStreamSynchronize(stream.stream());
```

src/fvdb/detail/io/SaveNanoVDB.cu
```diff
@@ -0,0 +1,705 @@
...
+                           const std::vector<std::string> &names,
+                           nanovdb::io::Codec codec,
+                           bool verbose) {
+    const nanovdb::GridHandle<TorchDeviceBuffer> &nanoGridHdl = gridBatchData.nanoGridHandle();
+
+    // Make a (possible) cpu copy of the data jagged tensor
+    JaggedTensor cpuData = data.cpu().contiguous();
+    TORCH_CHECK(cpuData.is_contiguous(), "Jagged tensor must be contiguous");
+
+    // Hoist tensor shape info (issue #4): rdim and tail (everything past dim 0) are constant
```
harrism:
```suggestion
    // Hoist tensor shape info: rdim and tail (everything past dim 0) are constant
```

harrism (approved):
Looks good. Mostly just comment / naming nits.

PR #651, opened 2026-05-10
Avoid repeated delta computation

matthewdcong:
Delta is already computed in the line above so we can use the equivalent `sigma(...)` overload to save a few FLOPs.


swahtz (approved):
Good catch, cheers!

PR #653, opened 2026-05-16
docs: fix marching_cubes return type (unique vertex indices, not normals)

### LINKED ISSUE #422: Incorrect marching cube documentation
The doc for [Grid.marching_cubes()](https://openvdb.github.io/fvdb-core/api/sparse_grids.html#fvdb.Grid.marching_cubes) mentions that the third-argument is the vertex normals. But the tensor is of type `int64`. Looking at it looks like it's returning a list of vertex ids or something?

https://github.com/openvdb/fvdb-core/blob/2a595e1a7a29c6bed935997a256c3d6c40a9c280/src/fvdb/detail/ops/MarchingCubes.cu#L372-L382

mvanhorn:
## Summary

`marching_cubes` returns three values: vertices, faces, and an int64 tensor of unique vertex indices (the output of `torch.unique_dim` at `MarchingCubes.cu:357-385`). The docstrings in `fvdb/grid.py`, `fvdb/grid_batch.py`, and `fvdb/functional/_meshing.py` described the third return as "vertex normals", which the implementation does not produce.

## Why this matters

The docs led users to expect a normals tensor and write code like `vertices, faces, normals = ...`. The third value is actually `unqVertIdx` from the unique-vertex pass, which has a completely different shape (int64) and meaning.

## Changes

- `fvdb/grid.py`, `fvdb/grid_batch.py`, `fvdb/functional/_meshing.py` - update the docstring of `marching_cubes` at all four call sites to describe the third return as "unique vertex indices (int64)".

Fixes #422

### DIFF
--- a/fvdb/functional/_meshing.py
+++ b/fvdb/functional/_meshing.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 """Functional API for meshing and TSDF integration on sparse grids."""
+
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
@@ -36,7 +37,8 @@ def marching_cubes_batch(
     Returns:
         vertices (JaggedTensor): Mesh vertex positions, shape ``(B, -1, 3)``.
         faces (JaggedTensor): Triangle face indices.
-        normals (JaggedTensor): Per-vertex normals.
+        unique_vertex_ids (JaggedTensor): Unique vertex IDs with int64 values,
+            shape ``(B, -1, 3)``.
 
     .. seealso:: :func:`marching_cubes_single`
     """
@@ -60,7 +62,8 @@ def marching_cubes_single(
     Returns:
         vertices (torch.Tensor): Mesh vertex positions, shape ``(N, 3)``.
         faces (torch.Tensor): Triangle face indices.
-        normals (torch.Tensor): Per-vertex normals.
+        unique_vertex_ids (torch.Tensor): Unique vertex IDs with dtype ``torch.int64``,
+            shape ``(N, 3)``.
 
     .. seealso:: :func:`marching_cubes_batch`
     """
--- a/fvdb/functional/_meshing.py
+++ b/fvdb/functional/_meshing.py
@@ -37,8 +37,13 @@ def marching_cubes_batch(
     Returns:
         vertices (JaggedTensor): Mesh vertex positions, shape ``(B, -1, 3)``.
         faces (JaggedTensor): Triangle face indices.
-        unique_vertex_ids (JaggedTensor): Unique vertex IDs with int64 values,
-            shape ``(B, -1, 3)``.
+        vertex_edge_keys (JaggedTensor): Edge keys used to deduplicate mesh
+            vertices, dtype ``torch.int64``, shape ``(B, -1, 3)``. Each row
+            ``[batch_idx, voxel_a, voxel_b]`` identifies the grid edge on which
+            the vertex was interpolated, where ``voxel_a`` and ``voxel_b`` are
+            flat voxel indices of the two endpoint voxels (``voxel_a >=
+            voxel_b``). This can be used to map mesh vertices back to the grid
+            edges and voxels they originated from.
 
     .. seealso:: :func:`marching_cubes_single`
     """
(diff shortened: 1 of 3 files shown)


mvanhorn:
The two red CUDA 13.0 jobs look unrelated to this docs-only change. Both fail at import time with `ImportError: /...fvdb/libfvdb.so: undefined symbol: _ZNK2at10TensorBase14const_data_ptrIaEEPKT_v` (demangled: `at::TensorBase::const_data_ptr<signed char>()`), which is a PyTorch ATen ABI mismatch between the pip CUDA 13.0 build's libtorch and the one installed at test time. The same tests pass on Conda and pip CUDA 12.8.

No code change should be needed here. Happy to leave it open for a re-run once the CUDA 13.0 build pipeline is reconciled with the test environment.

swahtz (changes requested):
Thanks for your contribution @mvanhorn and fixing up this issue.  The tests should pass now if you want to just merge `main` into your PR branch.

While the docstring for `unique_vertex_ids` is technically accurate, it just defines the output using the variable name and is not really helpful for the user to understand what the meanings are of those vertex IDs.

As I look at what `unique_vertex_ids` actually is, I think we should expand the scope of this Issue to improve what is being returned here.  The shape of this output for the `single` function is `(N, 3)` and for the `batch` function it is `(B,-1,3)` and the meaning of the rows per vertex is something like`[batch_idx, voxel_a, voxel_b]` which identifies the grid edge on which the vertex was interpolated, where `voxel_a` and `voxel_b` are flat voxel indices of the two endpoint voxels.  `batch_idx` is entirely repetitive information because the batch is encoded by the JaggedTensor implicitly in the `batch` function and has no meaning (or is 0) in the `single` function.  We could just change this output to be shape `(N,2)`. 

So if you want to keep this PR scoped as-is, I think we should change the name of this return value to something like `vertex_edge_keys` (the uniqueness was just an internal implementation detail) and change our documentation something like (update shape for the `batch` version as appropriate):

"Edge keys used to deduplicate mesh vertices, dtype `torch.int64`, shape `(N, 3)`.  Each row `[batch_idx, voxel_a, voxel_b]` identifies the grid edge on which the vertex was interpolated, where `voxel_a` and `voxel_b` are flat voxel indices of the two endpoint voxels (`voxel_a >= voxel_b`).  This can be used to map mesh vertices back to the grid edges and voxels they originated from."

If you want to just open an Issue for the return shape difference and just land this docstring change, that's fine, or happy if we want to do both in this one PR since they're both small changes.

mvanhorn:
The failing CUDA `Doc Tests` and `Unit Tests` look unrelated to this PR - the change is a 9-line docstring fix for `marching_cubes` return-type wording, no source or test files touched. Static analysis (Python parsing, whitespace lint) passed; couldn't run sphinx-build locally without the CUDA env. Could a maintainer re-run CI when the queue allows? Happy to dig in if a re-run still fails.

swahtz:
> The failing CUDA `Doc Tests` and `Unit Tests` look unrelated to this PR - the change is a 9-line docstring fix for `marching_cubes` return-type wording, no source or test files touched. Static analysis (Python parsing, whitespace lint) passed; couldn't run sphinx-build locally without the CUDA env. Could a maintainer re-run CI when the queue allows? Happy to dig in if a re-run still fails.

The issue with CI is related to your branch not having merged the latest from `main`, if you do that then the tests should pass.  The review notes also need to be addressed.

swahtz:
Pushed the requested changes, keeping this PR scoped to docs:

- Merged `main` in — should clear up the CI failures from the stale base.
- Renamed the third return value's docstring label from `unique_vertex_ids` to `vertex_edge_keys` in `fvdb/grid.py`, `fvdb/grid_batch.py`, and `fvdb/functional/_meshing.py`, and documented the `[batch_idx, voxel_a, voxel_b]` encoding as discussed in review. Shape stays `(N, 3)` / `(B, -1, 3)` for now — no behavior change, this is still docs-only.

Opened #678 to track dropping the redundant `batch_idx` column (shrinking the output to `(N, 2)` / `(B, -1, 2)`) as separate follow-up work, so it doesn't block this fix.

Should be good to merge from my end once CI is green.

swahtz (approved):

PR #657, opened 2026-05-20
More optimal prefetching for mGPU Gaussian splatting

matthewdcong:
This PR improves and optimizes prefetching for mGPU Gaussian splatting.

1. Prefetching is batched via `cudaMemPrefetchBatchAsync` when possible. This amortizes the syscall overhead across multiple prefetches.
2. Instead of calling `zeros`, we call `empty` instead followed by a (batched) prefetch and memset.

In addition, this PR includes two cleanup changes:

1. Remove `warp_group_g in` backwards projection. `g` is guaranteed to be unique for each thread in the warp, so the warp sum is unnecessary.
2. Allocate CUB temporaries on the mGPU path using `cudaMallocAsync/cudaFreeAsync` as opposed to the built-in PyTorch CUDA block allocator. This prevents memory contention between the PyTorch block allocator and built-in CUDA mempool.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ requires-python = ">=3.10"
 readme = "README.md"
 dependencies = [
     # Require torch; users will choose the correct CUDA build from PyTorch's index
-    "torch>=2.8,<2.12",
+    "torch>=2.8,<2.13",
     "numpy",
 ]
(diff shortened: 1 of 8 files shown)


blackencino (approved):
LGTM.

My kneejerk response in seeing the code is that the std::vector allocations seem like potential speedbumps, but that's not based on anything other than an outdated aversion to allocations in code that's speed-focused. I don't really think there's any problem, it's just a conditioned response.

PR #658, opened 2026-05-20
Add warp level early exit for forward rasterization

matthewdcong:
In addition to the block-level early exit, we can further improve the efficiency of the rasterization operator via a nested warp-level early exit. If each thread in the warp has completed, we can skip the per-Gaussian inner loop.


blackencino (approved):
SHIP IT!

PR #663, opened 2026-05-28
`ray_implicit_intersection` improvements

swahtz:
Optimise the per-ray SDF zero-crossing kernel (`ray_implicit_intersection`) and the shared HDDA traversal layer it sits on, fixing a few correctness issues along the way and carrying two adjacent ray-marching ops (`uniform_ray_samples`, `voxels_along_rays`) along for the ride.

The structural heart of the PR is a reorganization of the HDDA iterators: the formerly named `HDDAVoxelIterator` becomes `HDDAActiveValueIterator`, and a new leaf-only `HDDALeafVoxelIterator` lets `ray_implicit_intersection` skip active coarse tiles entirely (à la `nanovdb::ZeroCrossing`). Both are aliases over `HDDAValueIteratorImpl`, so they share a lot of functionality in one class.  The rest of this PR is supporting infrastructure (a shared `__ldg`/`__stwt` cache-hint helper), compile-time specialisations, warp-divergence cleanups, and broadened tests.

This started after being inspired by Efty's 'branchless getValue' work in NanoVDB and wanting to apply the same principles to the ray-marching kernels; as I pulled the thread, more fruit fell out of the tree (to branch metaphors).

End-to-end on my RTX PRO 6000 Blackwell, 1 M rays per workload: **1.67×–1.95× speedup for `ray_implicit_intersection`** (−48 to −51% kernel instructions, −2 registers/thread) and **1.05×–1.10× speedup for `uniform_ray_samples`** (−6 registers/thread on the count pass). `c10::Half` inputs are now dispatched for `ray_implicit_intersection` (previously `NotImplementedError`). All fp32 outputs are bit-identical to `main`.

---

## 1. HDDA iterator reorganization (`HDDAIterators.h`) — the foundational change

**The slight misnomer.** `HDDAVoxelIterator` did not yield voxels — it yielded *active values at any node level*: active coarse tiles (`getDim > 1`) as well as active leaf voxels (`getDim == 1`). Callers were expected to branch on `getDim()` to tell them apart. Anything indexing a strictly per-leaf-voxel buffer (like `ray_implicit_intersection`, which reads `gridScalars[getValue(ijk) - 1]`) was therefore walking tiles it never needs to see.

**The fix — one impl, two named aliases:**

- Rename `HDDAVoxelIterator` → **`HDDAActiveValueIterator`** (honest name: walks every active value node, tiles included).
- Add **`HDDALeafVoxelIterator`**: yields only active leaf voxels; active tiles are skipped in a single coarse HDDA step.
- Both are aliases over one `HDDAValueIteratorImpl<AccT, ScalarT, bool LeafOnly>`. `LeafOnly` is a **compile-time** gate (like `EpsZero`/`ConeZero` below) that reuses the `dim` already computed by the realign loop, so the active-value path costs nothing extra and the leaf path gets free tile-skipping.

**Why it matters:** the leaf alias is exactly the narrow-band structure `nanovdb::ZeroCrossing` walks, and adopting it in `ray_implicit_intersection` is the single biggest performance lever in this PR (see §3). It also removes a latent mis-index on tile-bearing grids.

**Control-flow / warp-divergence cleanups in the same header** (benefit every consumer):

- `HDDASegmentIterator::nextSegment`: predicated `TimeSpan` setters instead of an `if`/`else` ladder; only the "active region ended" `break` stays a real branch, so rays in a warp with differing `active` state no longer diverge at the setter level.
- `nextVoxel`: a bounded `for (pass < 3)` level-realign loop instead of three unrolled `getDim`/level-update passes; collapses to a single descent in the common level-aligned case.

**Blast radius.** The iterators are public API reached by `ray_implicit_intersection`, `uniform_ray_samples`, `voxels_along_rays`, and `segments_along_rays`. `VoxelsAlongRays.cu` moves to the renamed `HDDAActiveValueIterator` with **no behaviour change** (it intentionally walks all active values).

## 2. Shared CUDA cache-hint helpers (`Caching.cuh`, new) — the other shared building block

A small header surfacing reusable, host-safe cache hints, consumed by both ops in §3 and §4:

- `_loadReadOnly` — `__ldg` (`.NC`) for read-only SDF data, keeping it off the active-mask working set.
- `_storeStreaming` / `_storeStreamingPair` — `__stwt` (`.CS`) for write-once outputs so they aren't promoted into L1 (where they'd evict active-mask leaf data).

Both fall back to a plain load/store on host and for types without the intrinsic (e.g. `c10::Half`).

## 3. `ray_implicit_intersection` (`RayImplicitIntersection.cu`) — the headline op

With the building blocks in place, the op rewrite splits cleanly into correctness and performance.

### Correctness / semantics

- **Convention-agnostic seed.** The first valid (non-NaN) voxel along each ray seeds the sign reference; the first subsequent opposite-sign voxel is the hit. Previously the kernel baked in a "positive outside, negative inside" sentinel. Now both SDF sign conventions work, and rays that start *inside* the surface (my primary concern) report the **exit** crossing instead of a spurious result. Matches `nanovdb::ZeroCrossing`.
- **Band-continuity gating** via `lastT1`. Sub-voxel-interpolate the hit only when the previous voxel is contiguous along the ray (`t0 == lastT1`); otherwise fall back to bracket-entry time. This stops the kernel from interpolating a "hit" across empty space — a gap between two disjoint SDF regions, or a run of NaN tile values.
- **fp16 dispatch.** Add `c10::kHalf` to `AT_DISPATCH_V2` so half-precision rays are actually dispatched (previously `NotImplementedError` once the fp16 test skip was removed).
- **Sharper validation message** on `gridScalars.rsize(0) == totalVoxels()`, naming the iterator contract (leaf iterator needs one scalar per voxel; per-active-value data wants `HDDAActiveValueIterator`).

### Performance

- **Iterate leaf voxels only** (`HDDALeafVoxelIterator` from §1) — **the dominant speedup.** The old active-value traversal ran each coarse tile through the full per-voxel body even though a tile can never be the leaf-level zero crossing; the leaf iterator skips them in one coarse HDDA step.
- **`bool EpsZero` specialisation.** The launcher dispatches on `eps == 0.0f` (the common case), dropping the per-voxel `deltaT < eps` branch and a register.
- **Cache hints** (from §2): `gridScalars` via `_loadReadOnly`, `outTimes` via `_storeStreaming`.
- **fp32 math on half inputs**: interpolation/time arithmetic in `at::opmath_type<ScalarT>`, cast back only at the store.
- **Branchless sign-flip detection** (`scalarSign != voxelSign`).

## 4. `uniform_ray_samples` (`SampleRaysUniform.cu`) — adjacent op tightening

Same playbook applied to the neighbouring sampler:

- Adopt the shared `Caching.cuh` helpers (drop the local copies).
- Factor `_emitSample` into a free function template — extended `__device__` lambdas can't nest in the launchers' generic lambdas; separately-deduced pair types avoid per-call casts of `at::opmath_type<Half>` back to `c10::Half`.
- **Compile-time specialise** `countSamplesPerRayCallback` / `generateRaySamplesCallback` on `ConeZero` / `IncludeEndpoints` / `ReturnMidpoint`. The `ConeZero` case hoists `stepSize = minStepSize` out of the inner loop, dropping a `Clamp+mul` and registers (count pass goes from 80 → 74 registers).

## 5. Docs & tests

**Docstrings** (`fvdb/grid.py`, `fvdb/grid_batch.py`, `fvdb/functional/_ray.py`): document the convention-agnostic semantics — first valid voxel seeds the sign reference, first opposite-sign voxel is the hit; both conventions work, inside-surface rays report the exit crossing.

**Python** (`tests/unit/test_basic_ops.py`):

- Two regressions for the correctness fixes: `..._starts_inside_surface` (inside-origin ray reports the exit crossing) and `..._two_disjoint_regions` (reports the first surface, not the empty gap).
- 7 new `all_device_dtype_combos` tests (34 instances) adapted from OpenVDB's `TestLevelSetRayIntersector.cc` / `TestNanoVDB.cc`: sign-of-zero equivalence, axis-aligned & diagonal analytic roots, four miss configs, non-trivial transform, a 64×64 sweep, and single-voxel bracket-entry interpolation.
- `..._wrong_scalar_count_errors`: a mismatched `gridScalars` count raises an error naming the iterator contract.

**C++** (`src/tests/HDDAIteratorsTest.cpp`, new gtest): builds a `NanoGrid<float>` and drives both aliases on the host. A pure-leaf region yields identical voxels; on a ray crossing an active 128³ tile, `HDDAActiveValueIterator` surfaces it (`getDim > 1`) while `HDDALeafVoxelIterator` skips it. (Only way to exercise the active-tile path — fvdb's Python builders produce leaf-only grids.)

## 6. Performance

Measured on RTX PRO 6000 Blackwell. Timing uses `torch.cuda.Event` over a 1024² = 1,048,576-ray camera-pinhole bundle aimed at each grid's active bbox; median of 20 iterations after trimming the top and bottom two outliers. ncu metrics use `--profile-from-start off` with a `cudaProfilerStart/Stop` window around a single 512² = 262,144-ray launch, fp32, on the optimization-engaged setting (`eps = 0` / `cone_angle = 0`).

### Wall-clock timing (median ms per call, 1 M rays)

| op | dataset | dtype | setting | main | wip | speedup |
|---|---|---|---|---:|---:|---:|
| `ray_implicit_intersection` | dragon (267 MB SDF) | fp32 | eps=0 | 1.263 | 0.646 | **1.95×** |
| `ray_implicit_intersection` | dragon | fp32 | eps=1e-3 | 1.267 | 0.658 | **1.93×** |
| `ray_implicit_intersection` | dragon | fp16 | eps=0 | n/a (unsupported) | 0.687 | _new on wip_ |
| `ray_implicit_intersection` | dragon | fp16 | eps=1e-3 | n/a (unsupported) | 0.692 | _new on wip_ |
| `ray_implicit_intersection` | emu (1.0 GB SDF) | fp32 | eps=0 | 1.832 | 1.059 | **1.73×** |
| `ray_implicit_intersection` | emu | fp32 | eps=1e-3 | 1.830 | 1.093 | **1.67×** |
| `ray_implicit_intersection` | emu | fp16 | eps=0 | n/a (unsupported) | 1.058 | _new on wip_ |
| `ray_implicit_intersection` | emu | fp16 | eps=1e-3 | n/a (unsupported) | 1.075 | _new on wip_ |
| `ray_implicit_intersection` | crawler (1.6 GB SDF) | fp32 | eps=0 | 3.586 | 2.143 | **1.67×** |
| `ray_implicit_intersection` | crawler | fp32 | eps=1e-3 | 3.649 | 2.145 | **1.70×** |
| `ray_implicit_intersection` | crawler | fp16 | eps=0 | n/a (unsupported) | 2.541 | _new on wip_ |
| `ray_implicit_intersection` | crawler | fp16 | eps=1e-3 | n/a (unsupported) | 2.539 | _new on wip_ |
| `uniform_ray_samples` | dragon | fp32 | cone=0 | 2.953 | 2.790 | 1.06× |
| `uniform_ray_samples` | dragon | fp32 | cone=1e-3 | 2.943 | 2.776 | 1.06× |
| `uniform_ray_samples` | dragon | fp16 | cone=0 | 2.973 | 2.750 | 1.08× |
| `uniform_ray_samples` | dragon | fp16 | cone=1e-3 | 2.986 | 2.802 | 1.07× |
| `uniform_ray_samples` | wdas_cloud (2.6 GB fog) | fp32 | cone=0 | 17.216 | 16.391 | 1.05× |
| `uniform_ray_samples` | wdas_cloud | fp32 | cone=1e-3 | 17.154 | 16.346 | 1.05× |
| `uniform_ray_samples` | wdas_cloud | fp16 | cone=0 | 18.144 | 16.493 | 1.10× |
| `uniform_ray_samples` | wdas_cloud | fp16 | cone=1e-3 | 18.235 | 17.186 | 1.06× |

All fp32 outputs are bit-equivalent main vs this PR branch (output checksum match within `atol=1e-3 * max(|main|, |wip|)`), including the `ray_implicit_intersection` rows where the leaf iterator now skips tiles.

### ncu microarchitecture (single launch, fp32, 262k rays)

Columns are **main → wip**. Lower-is-better for registers, instructions, cycles, and DRAM bytes; higher-is-better for occupancy, SIMD efficiency (`smsp__thread_inst_executed_per_inst_executed.ratio`, out of 32), and L1 hit rate.

| op | dataset | pass | regs | occ % | SIMD eff /32 | insts | inst Δ | cycles | L1 hit % | DRAM MB |
|---|---|---|---|---|---|---|---:|---|---|---|
| `ray_implicit_intersection` | dragon | implicit | 80 → 78 | 14.0 → 14.6 | 10.62 → 8.46 | 1.29e+08 → 6.35e+07 | **−50.6%** | 2.08e+06 → 1.02e+06 | 94.5 → 87.8 | 17.8 → 17.7 |
| `ray_implicit_intersection` | emu | implicit | 80 → 78 | 12.3 → 12.8 | 9.67 → 7.75 | 1.89e+08 → 9.89e+07 | **−47.6%** | 3.11e+06 → 1.59e+06 | 94.0 → 87.2 | 30.9 → 30.7 |
| `ray_implicit_intersection` | crawler | implicit | 80 → 78 | 8.4 → 9.2 | 7.32 → 6.92 | 1.57e+08 → 7.73e+07 | **−50.8%** | 9.19e+06 → 5.87e+06 | 93.4 → 87.0 | 29.9 → 29.6 |
| `uniform_ray_samples` | dragon | count | **80 → 74** | 14.1 → 14.4 | 7.36 → 7.74 | 2.23e+08 → 2.41e+08 | +8.2% | 3.19e+06 → 2.97e+06 | 93.5 → 93.5 | 22.0 → 21.6 |
| `uniform_ray_samples` | dragon | emit | 78 → 80 | 14.0 → 14.2 | 7.35 → 7.66 | 2.26e+08 → 2.20e+08 | −2.6% | 3.24e+06 → 3.04e+06 | 92.9 → 92.9 | 23.0 → 23.0 |
| `uniform_ray_samples` | wdas_cloud | count | **80 → 74** | 16.0 → 16.1 | 4.68 → 4.86 | 1.55e+09 → 1.62e+09 | +4.4% | 1.88e+07 → 1.77e+07 | 88.4 → 88.4 | 211.6 → 211.9 |
| `uniform_ray_samples` | wdas_cloud | emit | 78 → 80 | 16.1 → 16.1 | 4.68 → 4.82 | 1.56e+09 → 1.56e+09 | −0.2% | 1.89e+07 → 1.77e+07 | 88.2 → 88.1 | 230.6 → 222.8 |

Reading the numbers:

- **The ≈ −50% instruction drop on `ray_implicit_intersection` is overwhelmingly the leaf-iterator switch (§1).** Skipped tile steps no longer run the full per-voxel body; output is bit-identical, confirming that work never affected the result. `EpsZero`, the `__ldg`/`__stwt` hints, and the predicated `HDDAIterators` setters contribute the remainder.
- **Registers/thread:** 80 → 78 on the implicit kernel, **80 → 74 on the `uniform_ray_samples` count pass** (the `ConeZero` step-size hoist).
- **Occupancy gains are small (+0.3 to +0.8 pts)** — the kernel is register-bound on Blackwell and −2 registers doesn't cross a boundary; the win is instruction- and cycle-driven (cycles roughly halve on the implicit kernel).
- **SIMD efficiency on `ray_implicit_intersection` falls (10.62 → 8.46 on dragon)** — expected and benign: the skipped tile steps were perfectly convergent, so removing them lowers the per-instruction average while removing far more total instructions. `uniform_ray_samples` SIMD efficiency *improves* from the `HDDAIterators` predication.
- **L1 hit rate dips on `ray_implicit_intersection`** (94.5 → 87.8 on dragon): the streaming-store path bypasses L1 deliberately and tile-skipping removes trivially cache-friendly reads, so narrow-band accesses dominate the average. Net wall-clock is ~2× faster.
- **Instruction count rises on the `uniform_ray_samples` count pass** (+8.2% dragon) even as cycles fall — the specialised count callback does more per-thread work but with a better register profile, keeping more eligible warps in flight.

## 7. Notes / risks / migration

- **API semantics change for `ray_implicit_intersection`.** Now SDF-sign-agnostic; reports the EXIT crossing for rays that start inside the surface (previously `-1` or a spurious bracket-entry). Matches `nanovdb::ZeroCrossing`. Pinned by `..._starts_inside_surface` and `..._two_disjoint_regions`. Callers relying on the old "positive-outside-only" assumption will see different outputs on inside-surface rays — but those were almost certainly being treated as no-hit anyway.
- **`c10::Half` is now dispatched.** Code that caught `NotImplementedError` and fell back to fp32 conversion at the Python boundary should drop that branch.
- **Leaf-iterator switch is correct-by-construction for active tiles.** fvdb's own Python builders only produce leaf voxels, but a grid loaded via `from_nanovdb` may contain active coarse tiles authored elsewhere; the old iterator would have walked those into the per-voxel body (mis-indexing or reading `gridScalars[-1]`). The leaf iterator skips them, and `HDDAIteratorsTest` exercises this directly.


src/fvdb/detail/utils/nanovdb/HDDAIterators.h
```diff
@@ -195,29 +208,26 @@ template <typename AccT, typename ScalarT> struct HDDAVoxelIterator {
...
-            if (mHdda.dim() != dim) {
-                mRay.setMinTime(mHdda.time());
-                mHdda.update(mRay, dim);
-            }
-            dim = mAcc.getDim(mHdda.voxel(), mRay);
-            if (mHdda.dim() != dim) {
+            for (int pass = 0; pass < 3 && mHdda.dim() != dim; ++pass) {
                 mRay.setMinTime(mHdda.time());
                 mHdda.update(mRay, dim);
+                dim = mAcc.getDim(mHdda.voxel(), mRay);
```
areidmeyer, line 227: Looking at this again, I am wondering how much of the speedup is actually avoiding pointless calls to getDim(). As in the compiler could be doing loop unrolling itself, but not making that mistake.

areidmeyer:
```            
            int dim = mAcc.getDim(mHdda.voxel(), mRay);
            if (mHdda.dim() != dim) {
                mRay.setMinTime(mHdda.time());
                mHdda.update(mRay, dim);
                dim = mAcc.getDim(mHdda.voxel(), mRay);
            }     
            if (mHdda.dim() != dim) {
                mRay.setMinTime(mHdda.time());
                mHdda.update(mRay, dim);
                dim = mAcc.getDim(mHdda.voxel(), mRay);
            }
            if (mHdda.dim() != dim) {
                mRay.setMinTime(mHdda.time());
                mHdda.update(mRay, dim);
                // we can skip getDim() here, nothing uses it
            }
```
Wondering if you can even take it one step further and skip the last getDim().

areidmeyer (approved):
Approved, but one experiment I think worth trying.

src/fvdb/detail/utils/nanovdb/HDDAIterators.h
swahtz: Yes!  So actually (and I should have mentioned this in the PR description) but I've been working on [this](https://github.com/AcademySoftwareFoundation/openvdb/pull/2220) too over in NanoVDB which is a fused call to getDim and isActive which amortizes the tree walk that both do into a 1-time cost (which is some of the biggest part of the overall runtime).

Before I split the iterator into the active and active+leaf behaviours, the NanoVDB getDim/isActive fusion helped a lot but then I realized that for the primary target of this PR (`ray_implicit_intersection` performance), we didn't need the full active node behaviour (just the leaf-level iterator).  So the NanoVDB's 2220 PR will still speed up the ActiveValue iterator (implicit intersection doesn't need it), but I'll follow up with that when it merges.

This optimization you point out on the last getDim is something I'll roll into this PR

swahtz: So in theory `getDim`, in the future, comes along 'for free', but it's worth putting this in in the meantime.

swahtz: Actually that last `getDim`'s result is used in the next block to drive `isLeaf`.  So we do need it in the last of the loops.  However, when looking at this with Claude, we did find that nvcc wasn't auto-unrolling this loop so I've inserted a `#pragma unroll` which did produce a minor gain.

PR #664, opened 2026-05-30
Improve mGPU Gaussian tile intersection

matthewdcong:
Previous iterations of mGPU Gaussian tile intersection include:

1.  Distribute every step of the single mGPU tile sort. Compute the tile intersections for all Gaussians across all tiles into a single tensor, followed by a parallel mGPU radix sort. Requires significantly more communication and synchronization during the parallel radix sort, as well as a temp output key and value array for cross-device merging. (previously in `main`)
2. Observe that the radix sort is independent for each camera and compute the radix sort for each camera entirely on a single device. This works well when batch size = num GPUs, but performs poorly when batch size = 1 and num GPUs > 1 because only one GPU is used and all data must be gathered to this GPU. (currently in `main`)

This PR introduces a more performant strategy. First, we assign each GPU a subset of the tiles/the tile range that will be rendered by that GPU. Then, on each GPU, we compute the intersections of all Gaussians with only that subset of the tiles/the tile range. Since the tile keys are monotonically increasing, this means that the subsequent sorting process is decoupled, i.e. we can sort the per-GPU Gaussian tile intersection lists independently and the resulting flattened array is guaranteed to be sorted. This significantly reduces the amount of communication and data transfer required during the sorting process. Moreover, a switch from radix sort to merge sort enables us to remove the temp output buffers, further reducing stalls due to prefetching as well as decreasing peak memory utilization.

This is a performance improvement across the board, but becomes more significant as the number of GPUs increases. On 8x A100s, this improves end-to-end reconstruction performance about 15% with a batch size of 1 (on a relatively small problem).

### DIFF
--- a/src/fvdb/detail/ops/gsplat/IntersectGaussianTiles.cu
+++ b/src/fvdb/detail/ops/gsplat/IntersectGaussianTiles.cu
@@ -830,7 +830,8 @@ intersectGaussianTilesPrivateUse1Impl(
     }
     for (const auto deviceId: c10::irange(deviceCount)) {
         C10_CUDA_CHECK(cudaSetDevice(deviceId));
-        C10_CUDA_CHECK(cudaDeviceSynchronize());
+        auto stream = c10::cuda::getCurrentCUDAStream(deviceId);
+        C10_CUDA_CHECK(cudaStreamSynchronize(stream));
         deviceIntersectionOffset[deviceId] = totalIntersections;
         deviceIntersectionCount[deviceId]  = deviceTotals[deviceId];
         totalIntersections += deviceTotals[deviceId];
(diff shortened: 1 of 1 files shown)


src/fvdb/detail/ops/gsplat/IntersectGaussianTiles.cu
```diff
@@ -683,45 +747,103 @@ intersectGaussianTilesPrivateUse1Impl(
...
-    // Allocate tensor to store the number of tiles each gaussian intersects
-    torch::Tensor tilesPerGaussianCumsum =
-        torch::empty({totalGaussians}, means2d.options().dtype(torch::kInt32));
-
     const auto tileMaskPtr =
         tileMask.has_value() ? tileMask.value().const_data_ptr<bool>() : nullptr;
 
-    for (const auto deviceId: c10::irange(c10::cuda::device_count())) {
+    const int deviceCount        = static_cast<int>(c10::cuda::device_count());
+    const uint32_t totalTileKeys = numCameras * totalTiles;
```
swahtz:
```suggestion
    const int64_t totalTileKeys = static_cast<int64_t>(numCameras) * totalTiles;
```

The single GPU path computes this same variable as an int64_t.  `deviceChunk`, which  `totalTileKeys` is argument to, will widen it to `size_t` anyway, so might as well change it.

swahtz (approved):
One super minor comment not worth blocking for but other than that, it looks great to me!  Thanks

PR #665, opened 2026-06-03
Improve prefetch granularity for rasterization kernels

matthewdcong:
Prior to this PR, we prefetched image-space tensors with camera granularity. That is, if an image in a camera was rendered across multiple GPUs, each GPU would prefetch the entire image. This is overly conservative and increases the amount of bandwidth required.

This PR refines that prefetching to have better granularity by prefetching on the tiles rendered by each GPU. This reduces the amount of cross-device traffic and leads to an approximately 5% speedup on 2x RTX 3090s for training.

### DIFF
--- a/src/fvdb/detail/utils/cuda/Prefetch.cpp
+++ b/src/fvdb/detail/utils/cuda/Prefetch.cpp
@@ -9,8 +9,115 @@
 #include <c10/core/ScalarType.h>
 #include <c10/cuda/CUDAException.h>
 
+#include <algorithm>
+#include <cstdint>
+#include <vector>
+
 namespace fvdb {
 namespace detail {
+
+namespace {
+
+void
+appendPerTileTensorRanges(std::vector<void *> &prefetchPointers,
+                          std::vector<size_t> &prefetchSizes,
+                          const torch::Tensor &tensor,
+                          const TilePrefetchRange &range) {
+    const size_t scalarSize    = c10::elementSize(tensor.scalar_type());
+    const size_t tileSizeBytes = tensor.stride(2) * scalarSize;
+    const size_t prefetchSize  = range.count * tileSizeBytes;
+    auto *const prefetchPointer =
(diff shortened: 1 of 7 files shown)


src/fvdb/detail/utils/cuda/Prefetch.cpp
```diff
@@ -9,9 +9,116 @@
...
+    auto *const prefetchPointer =
+        static_cast<uint8_t *>(tensor.data_ptr()) + range.offset * tileSizeBytes;
+    if (prefetchSize > 0) {
+        prefetchPointers.emplace_back(prefetchPointer);
+        prefetchSizes.emplace_back(prefetchSize);
+    }
+}
+
+void
+appendPerTileImageRanges(std::vector<void *> &prefetchPointers,
```
swahtz, line 38: **🧹 chore:** ‏We should have some tests for this function.  `appendPerTileImageRanges` has some complex logic in it and correctness errors would mean a silent performance regression (instead of something breaking), so I think it makes sense we should have some tests for `appendPerTileImageRanges` (as opposed to the other Prefetch functions which seem pretty straightforward).

swahtz (changes requested):
Overall it looks good, I think we should just add a few gtests for the logic in Prefetch's `appendPerTileImageRanges` to make sure we don't have any silent regressions in logic before we merge this.

src/fvdb/detail/utils/cuda/Prefetch.cpp
matthewdcong: Added a test with several configurations (1 device, 1 camera, multiple devices w/ 1 camera, multiple cameras w/ 1 device, and multiple cameras and devices). Since `appendPerTileImageRanges` is not exposed outside of the translation unit, I've indirectly tested it via `appendPerTilePrefetchRanges`.

swahtz (approved):
Looks good to me, thanks for adding those tests!

PR #685, opened 2026-07-08
Migrate Gaussian splatting wrappers from fvdb-core to fvdb-reality-capture

matthewdcong:
This MR removes the high-level Gaussian splatting Python API, enums, tests, and docs from fvdb-core in preparation for their migration to fvdb-reality-capture. Low-level kernels and viewer support in remaining in fvdb-core. In order to decouple the visualization from the migrated high-level wrapper, we introduce a more agnostic `GaussianSplatViewData` and `Scene.add_gaussian_splat_tensors`


fvdb/viz/_gaussian_splat_3d_view.py
```diff
@@ -46,7 +46,7 @@ def __init__(
         eps_2d: float = 0.3,
         antialias: bool = False,
         sh_degree_to_use: int = -1,
-        sh_ordering_mode: ShOrderingMode = ShOrderingMode.RGB_RGB_RGB,
+        sh_ordering_mode: _ShOrdering = _RGB_RGB_RGB_SH_ORDERING,
```
swahtz: **❓ question:** ‏Why remove the `fvdb.ShOrderingMode` enum from fvdb-core if we need it here to define this for the view?  Couldn't `fvdb_reality_capture` import this enum from core for its usage or would it create a strange dependency between gsplat and fvdb that I'm not seeing.

matthewdcong: Good catch, fixed.

fvdb/viz/_gaussian_splat_3d_view.py
```diff
@@ -60,20 +55,23 @@ def __init__(
...
             tile_size (int): The tile size to use for rendering. Default is 16.
             min_radius_2d (float): The minimum radius in pixels to use when rendering splats. Default is 0.0.
             eps_2d (float): The epsilon value to use when rendering splats. Default is 0.3.
             antialias (bool): Whether to use antialiasing when rendering splats. Default is False.
             sh_degree_to_use (int): The degree of spherical harmonics to use when rendering colors.
                 If -1, the maximum degree supported by the Gaussian splat 3D scene is used. Default is -1.
-            sh_ordering_mode (ShOrderingMode): The spherical harmonics ordering mode to use when rendering colors.
-                Default is :attr:`ShOrderingMode.RGB_RGB_RGB`.
-            _private (Any): A private object to prevent direct construction. Must be :attr:`GaussianSplat3dView.__PRIVATE__`.
+            sh_ordering_mode (str): The spherical harmonics tensor layout to use when rendering colors. Must be
```
swahtz: **☑️ todo:** ‏This should revert back to being an `ShOrderingMode` enum

fvdb/viz/_gaussian_splat_3d_view.py
```diff
@@ -91,20 +89,17 @@ def __init__(
...
         view.eps_2d = eps_2d
         view.antialias = antialias
         view.sh_degree_to_use = sh_degree_to_use
-        if sh_ordering_mode == ShOrderingMode.RRR_GGG_BBB:
-            view.rgb_rgb_rgb_sh = False
-        elif sh_ordering_mode == ShOrderingMode.RGB_RGB_RGB:
-            view.rgb_rgb_rgb_sh = True
-        else:
-            raise ValueError(f"Invalid ShOrderingMode: {sh_ordering_mode}")
+        view.rgb_rgb_rgb_sh = sh_ordering_mode == ShOrderingMode.RGB_RGB_RGB
```
swahtz: **☑️ todo:** ‏We should just store the enum

docs/api/viz.rst
```diff
@@ -4,3 +4,14 @@ Visualization
 .. automodule:: fvdb.viz
     :members:
 
+Gaussian splats
+---------------
+
+The viewer accepts Gaussian splats through the core-owned
```
swahtz, line 14: **☑️ todo:** ‏We should `autoclass` the `ShOrderingMode` here

fvdb/enums.py
```diff
@@ -2,35 +2,7 @@
...
-
-    RRR_GGG_BBB = "rrr_ggg_bbb"
-    """
-    The feature channels of spherical harmonics are stored in separate blocks for each coefficient. *i.e.* The spherical harmonics
-    tensor corresponds to a (row-major) contiguous tensor of shape ``[num_coefficients, channels, num_sh_bases]``, where channels=3 for RGB.
-    """
+from enum import IntEnum
 
 
 class SmoothingMode(IntEnum):
```
swahtz, line 8: **🎵 note (non-blocking):** ‏Now that all the other Enums are removed or moved, we should probably just move `SmoothingMode` to a module more local to where it's being used.  An issue for another PR

fvdb/viz/_gaussian_splat_view_data.py
```diff
@@ -0,0 +1,147 @@
...
+        device = self.means.device
+        for name, tensor in tensors.items():
+            if tensor.dtype != dtype:
+                raise TypeError(f"All Gaussian splat tensors must have dtype {dtype}; {name} has dtype {tensor.dtype}")
+            if tensor.device != device:
+                raise ValueError(
+                    f"All Gaussian splat tensors must be on device {device}; {name} is on device {tensor.device}"
+                )
+
+        if self.means.ndim != 2 or self.means.shape[1] != 3:
```
swahtz, line 100: **🚩 issue:** ‏We should add a check for a zero-sized `num_gaussian` count.  That situation would crash the renderer if not checked here.

fvdb/viz/_gaussian_splat_view_data.py
```diff
@@ -0,0 +1,147 @@
...
+            "quats": self.quats,
+            "log_scales": self.log_scales,
+            "logit_opacities": self.logit_opacities,
+            "sh0": self.sh0,
+            "shN": self.shN,
+        }
+        for name, tensor in tensors.items():
+            if not isinstance(tensor, torch.Tensor):
+                raise TypeError(f"{name} must be a torch.Tensor, got {type(tensor).__name__}")
+            if not tensor.dtype.is_floating_point:
```
swahtz: **🚩 issue:** ‏We should make this check stronger to check these as `torch.float32` types.  Underlying code processes the buffers as `float` without checking beyond this.

fvdb/viz/_gaussian_splat_view_data.py
```diff
@@ -0,0 +1,147 @@
...
+        if self.sh0.shape[0] != num_gaussians:
+            raise ValueError(
+                f"sh0 must contain {num_gaussians} Gaussians in its leading dimension, got {self.sh0.shape[0]}"
+            )
+        if self.shN.shape[0] != num_gaussians:
+            raise ValueError(
+                f"shN must contain {num_gaussians} Gaussians in its leading dimension, got {self.shN.shape[0]}"
+            )
+
+        if self.sh_ordering == ShOrderingMode.RGB_RGB_RGB:
```
swahtz, line 128: **🚩 issue:** ‏The checks in this if/else block pass for any positive SH channel count.  The renderer hardcodes 3 (RGB) channels, we should explicitly check for 3 channels here.

src/tests/scripts/generate_top_gaussian_contributor_dataset.py
swahtz, line 1: **🚩 issue:** ‏This script generates the data that gtests use for validation of the basic ops, we shouldn't move these but perhaps we need to refactor them or change their imports to use reality-capture.  These are not used by any python unit tests so these should stay here.

src/tests/scripts/write_rasterize_forward_test_data.py
swahtz, line 1: **🚩 issue:** ‏This script generates the data that gtests use for validation of the basic ops, we shouldn't move these but perhaps we need to refactor them or change their imports to use reality-capture.  These are not used by any python unit tests so these should stay here.

tests/unit/test_gaussian_splat_3d.py
swahtz, line 1: **🚩 issue:** ‏By removing these unit tests (and `test_rasterize_from_world.py`), we actually lose all test coverage for a few operators in `fvdb-core`.  I did an analysis and came up with 6 C++ ops we'd lose coverage for:

`rasterizeWorldSpaceGaussiansFwd` — covered by all 11 active tests in` test_rasterize_from_world.py`.
`rasterizeWorldSpaceGaussiansBwd` — covered by the 8 tests that call `backward()`.
`projectGaussiansAnalyticJaggedFwd` — exercised by 5 `gaussian_render_jagged` tests.
`projectGaussiansAnalyticJaggedBwd` — covered by `test_jagged_projection_grads_multicamera`.
`saveGaussianPly` — 9 deleted tests.
`loadGaussianPly` — 6 deleted tests

So in order to remove these python unit tests, we need gtests (or python unit tests of the same python-bound op functionality) to make sure `fvdb-core` has coverage for the same functionality as the unit tests.  Something like:

| New target | Minimum cases to preserve |
|---|---|
| `GaussianRasterizeWorldSpaceTest` | Forward and finite-difference backward; SH0/SHN; masks/backgrounds; no intersections; pinhole/orthographic/OpenCV; depth/RGBD consistency; structural projected-path parity; ignored-distortion behavior. |
| `GaussianProjectionJaggedTest` | Variable `g_sizes`/`c_sizes` with `[2, 1]` cameras; forward parity with dense per-scene projection; backward multi-camera gradient additivity. |
| `GaussianPlyIOTest` | Standard and empty-SHN round-trips; NaN filtering; metadata-type round-trips; invalid keys; non-contiguous tensors. |

swahtz (changes requested):
There's just a few places we could cleanup with the enum usage and a few added checks but I think the biggest thing these changes should resolve is having test coverage for the Ops in `fvdb-core` that removing the Gaussian splat unit tests would take with it.

tests/unit/test_gaussian_splat_3d.py
matthewdcong: Added C++ tests to restore coverage

src/tests/scripts/generate_top_gaussian_contributor_dataset.py
harrism: So this introduces a depedency of core on frc -- is that what we want to do?

harrism (commented):
My only question is: should we deprecate the old functionality first before breaking it?

swahtz (approved):
Great, thanks for addressing those notes.  Looks good to me now

PR #687, opened 2026-07-11
Avoid explicit materialization of view directions in SH evaluation

matthewdcong:
Currently, for SH computations we first calculate a tensor consisting of a per-camera per-Gaussian view direction from the per-camera view matrix. This view direction is then passed into the SH operator to evaluate view dependent contributions for higher-order terms.

Instead, we pass the view matrices directory the kernel and compute the view directions on-the-fly within the kernel from the viewmats. This avoids materializing an O(numCameras * numGaussians * 3) tensor of view directions and sidesteps the costs associated with evaluating and accessing that tensor. Effectively, this is a fusion of view direction evaluation and SH evaluation.

This results in an 1.08x speed up in end-to-end training for the bicycle scene.


src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -285,14 +300,18 @@ computeShBackward(
     const int64_t K,
     const int64_t D,
     const int64_t shDegreeToUse,
-    const torch::PackedTensorAccessor64<T, 3, torch::RestrictPtrTraits> viewDirs,     // [C, N, 3]
-    const torch::PackedTensorAccessor64<T, 3, torch::RestrictPtrTraits> shNCoeffs,    // [K-1, N, D]
-    const int *__restrict__ radii,                                                    // [C, N, 2]
+    const T *__restrict__ means,                                                   // [G, 3]
```
swahtz: **⌨️ typo:** ‏I believe `N` is our usual abbreviation for the number of Gaussians, not `G`

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -285,14 +300,18 @@ computeShBackward(
...
+    const T *__restrict__ worldToCamMatrices,                                      // [V, 4, 4]
```
swahtz: **⌨️ typo:** ‏I believe this should be `[C, 4, 4]`, I'm not sure what `V` is?


```suggestion
    const T *__restrict__ worldToCamMatrices,                                      // [C, 4, 4]
```

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -285,14 +300,18 @@ computeShBackward(
...
+    const int32_t *__restrict__ gaussianIds,                                       // [N] optional
+    const torch::PackedTensorAccessor64<T, 3, torch::RestrictPtrTraits> shNCoeffs, // [K-1, N, D]
+    const int *__restrict__ radii,                                                 // [C, N, 2]
     const torch::PackedTensorAccessor64<T, 3, torch::RestrictPtrTraits>
-        dLossDRenderQuantities,                                                       // [C, N, D]
+        dLossDRenderQuantities,                                                    // [C, N, D]
     torch::PackedTensorAccessor64<T, 3, torch::RestrictPtrTraits> outDLossDSh0Coeffs, // [N, 1, D]
     torch::PackedTensorAccessor64<T, 3, torch::RestrictPtrTraits> outDLossDShNCoeffs, // [N, K-1, D]
-    T *__restrict__ outDLossDViewDirs // [C, N, 3] optiondl
+    T *__restrict__ outDLossDMeans,             // [G, 3] optional
```
swahtz: **⌨️ typo:** ‏Same as above, I believe `G`=`N` and `V`=`C`

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -371,36 +449,46 @@ computeShDiffuseOnlyBackward(
...
+std::tuple<torch::Tensor, torch::Tensor, torch::Tensor, torch::Tensor>
 dispatchEvaluateSphericalHarmonicsBwd(const int64_t shDegreeToUse,
                                       const int64_t numCameras,
                                       const int64_t numGaussians,
-                                      const torch::Tensor &viewDirs,  // [C, N, 3] or empty
-                                      const torch::Tensor &shNCoeffs, // [N, K-1, D]
-                                      const torch::Tensor &dLossDColors,
-                                      const torch::Tensor &radii,     // [C, N, 2]
-                                      const bool computeDLossDViewDirs);
+                                      const torch::Tensor &means,              // [G, 3]
```
swahtz:
```suggestion
                                      const torch::Tensor &means,              // [N, 3]
```

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -371,36 +449,46 @@ computeShDiffuseOnlyBackward(
...
+                                      const torch::Tensor &worldToCamMatrices, // [V, 4, 4]
```
swahtz:
```suggestion
                                      const torch::Tensor &worldToCamMatrices, // [C, 4, 4]
```

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -371,36 +449,46 @@ computeShDiffuseOnlyBackward(
...
-    const torch::Tensor &viewDirs,               // [C, N, 3]
-    const torch::Tensor &shNCoeffs,              // [N, K-1, D]
-    const torch::Tensor &dLossDRenderQuantities, // [C, N, D]
-    const torch::Tensor &radii,                  // [C, N, 2]
-    const bool computeDLossDViewDirs) {
+std::tuple<torch::Tensor, torch::Tensor, torch::Tensor, torch::Tensor>
+dispatchEvaluateSphericalHarmonicsBwd<torch::kCUDA>(const int64_t shDegreeToUse,
+                                                    const int64_t numCameras,
+                                                    const int64_t numGaussians,
+                                                    const torch::Tensor &means,
```
swahtz: **🤓 nitpick (non-blocking):** ‏Could we preserve the tensor shape comments here?  Just makes things easier to track when reading through in the future

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -428,15 +516,52 @@ dispatchEvaluateSphericalHarmonicsBwd<torch::kCUDA>(
...
+                         (hasMeans && hasViewMats && cameraIds.defined() && gaussianIds.defined() &&
+                          (means.size(0) != N || worldToCamMatrices.size(0) != C));
+
+    // View directions are computed in the kernel from the means and world-to-camera matrices.
     if (hasShNCoeffs && K > 1 && shDegreeToUse > 0) {
-        TORCH_CHECK_VALUE(viewDirs.dim() == 3, "viewDirs must have shape [C, N, 3]");
+        TORCH_CHECK_VALUE(means.is_cuda() && means.scalar_type() == torch::kFloat32,
+                          "means must be a float32 CUDA tensor");
+        TORCH_CHECK_VALUE(means.dim() == 2 && means.size(1) == 3 && means.is_contiguous(),
+                          "means must be contiguous with shape [G, 3]");
```
swahtz:
```suggestion
                          "means must be contiguous with shape [C, 3]");
```

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -428,15 +516,52 @@ dispatchEvaluateSphericalHarmonicsBwd<torch::kCUDA>(
...
+        TORCH_CHECK_VALUE(worldToCamMatrices.is_cuda() &&
+                              worldToCamMatrices.scalar_type() == torch::kFloat32,
+                          "worldToCamMatrices must be a float32 CUDA tensor");
+        TORCH_CHECK_VALUE(worldToCamMatrices.dim() == 3 && worldToCamMatrices.size(1) == 4 &&
+                              worldToCamMatrices.size(2) == 4 && worldToCamMatrices.is_contiguous(),
+                          "worldToCamMatrices must be contiguous with shape [V, 4, 4]");
```
swahtz:
```suggestion
                          "worldToCamMatrices must be contiguous with shape [C, 4, 4]");
```

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.h
```diff
@@ -20,30 +20,40 @@ namespace ops {
 /// @param[in] shDegreeToUse Degree of spherical harmonics used in the forward pass
 /// @param[in] numCameras Number of cameras used in the forward pass
 /// @param[in] numGaussians Number of Gaussians used in the forward pass
-/// @param[in] viewDirs Direction vectors [N, 3] (packed) or [C, N, 3] (unpacked) used in the
-/// forward pass
+/// @param[in] means Gaussian means in world space [G, 3], where G is N in dense mode
```
swahtz: **⌨️ typo:** ‏Here we have the G and V nomenclature used again which I think should be N and C.  The comments mention these are the values in dense mode, I'm not sure why their values change in sparse render mode?

fvdb/__init__.py
```diff
@@ -278,23 +277,25 @@ def gaussian_render_jagged(
 def evaluate_spherical_harmonics(
     sh_degree: int,
     num_cameras: int,
+    means: torch.Tensor,
+    world_to_camera_matrices: torch.Tensor,
```
swahtz: **☑️ todo:** ‏When moving from `view_directions` to `world_to_camera_matrices`, we removed the default that `world_to_camera_matrices` can be `None` (when they are not needed to be supplied when `shN` is `None` and we're just doing sh 0 evaluation that doesn't require view direction).  Could we restore that functionality?

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsForward.cu
```diff
@@ -363,15 +434,24 @@ dispatchEvaluateSphericalHarmonicsFwd<torch::kPrivateUse1>(
...
     if (hasShNCoeffs && K > 0 && shDegreeToUse > 0) {
-        TORCH_CHECK_VALUE(viewDirs.dim() == 3, "viewDirs must have shape [C, N, 3]");
-        TORCH_CHECK_VALUE(
-            shNCoeffs.size(0) == viewDirs.size(1),
-            "shNCoeffs must have shape [N, K, D] and viewDirs must have shape [C, N, 3]");
-        TORCH_CHECK_VALUE(viewDirs.is_privateuseone(), "dirs must be a PrivateUse1 tensor");
-        TORCH_CHECK_VALUE(viewDirs.size(-1) == 3, "dirs must have last dimension 3");
+        const bool hasCamIds   = cameraIds.defined() && cameraIds.numel() > 0;
+        const bool hasGaussIds = gaussianIds.defined() && gaussianIds.numel() > 0;
+        TORCH_CHECK_VALUE(!hasCamIds && !hasGaussIds,
```
swahtz: **❓ question:** ‏ Here we only pass if camera Ids and gaussian Ids are not supplied but in `gaussian_render_jagged` we always supply them:

```
        render_quantities = _EvaluateGaussianSHFn.apply(
            actual_sh_degree,
            1,
            means.jdata,
            viewmats.jdata,
            camera_ids,
            gaussian_ids,
...
```

So that will mean we'll always throw an error here for jagged rendering whereas previously we didn't.  Is that intentional?

swahtz (changes requested):
Thanks for taking care of this long-standing optimization TODO.  Just a few comments about annotations but also an issue we should fix about the default None argument for the camera matrices when we don't need them and how the SH evaluation with jagged rendering is set up.

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.h
matthewdcong: Fixed the nomenclature and removed the incorrect dense mode qualifier. This is the same for sparse and dense.

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
matthewdcong: restored

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsForward.cu
matthewdcong: Yes, it's intentional. The mGPU codepath does not support jagged rendering before or after this change.  The single GPU codepath earlier on in the file still supports jagged rendering and accepts camera and gaussian ids.

swahtz: OK, great, thanks for clearing that up.

fvdb/__init__.py
matthewdcong: Restored as well. `world_to_camera_matrices` can now be none.

swahtz (approved):
Looks great, thanks for doing this.

PR #688, opened 2026-07-14
Fix reference cycle in _InJectFn leading to a memory leak

### LINKED ISSUE #684: Memory leak in GridBatch.inject_from under autograd
## Summary

GridBatch.inject_from(src_grid, features) leaks GPU memory on every backward call. The memory is never released: it survives gc.collect() (0 objects freed, so it's not a Python reference cycle) and torch.cuda.empty_cache(). The equivalent GridBatch.inject_from_ijk(src_ijk, features) and a plain-torch index gather, with identical inputs and outputs, leak nothing.

## Environment
- fvdb 0.5.0
- torch 2.12.1+cu130, CUDA 13
- NVIDIA RTX 3090, driver 610.43.02

## Repro
```python
import gc
import torch
import fvdb
from fvdb import GridBatch, JaggedTensor

DEV = "cuda"
torch.manual_seed(0)

def gather_inject_from(src_ijk, tgt_ijk, feats):
    src = GridBatch.from_ijk(JaggedTensor([src_ijk]))
    tgt = GridBatch.from_ijk(JaggedTensor([tgt_ijk]))
    return tgt.inject_from(src, JaggedTensor([feats])).jdata

def gather_inject_from_ijk(src_ijk, tgt_ijk, feats):
    src = GridBatch.from_ijk(JaggedTensor([src_ijk]))
    tgt = GridBatch.from_ijk(JaggedTensor([tgt_ijk]))
    return tgt.inject_from_ijk(src.ijk, JaggedTensor([feats])).jdata

def gather_index(src_ijk, tgt_ijk, feats):
    src = GridBatch.from_ijk(JaggedTensor([src_ijk]))
    tgt = GridBatch.from_ijk(JaggedTensor([tgt_ijk]))
    idx = src.ijk_to_index(JaggedTensor([tgt.ijk.jdata]), cumulative=True).jdata
    return feats[idx]

IMPLS = [
    ("inject_from (grid->grid)", gather_inject_from),
    ("inject_from_ijk (ijk->grid)", gather_inject_from_ijk),
    ("index gather (plain torch)", gather_index),
]

def allocated_mb():
    torch.cuda.synchronize()
    return torch.cuda.memory_allocated() / 1024**2

def measure_leak(name, fn, s, t, feats, ref, n_iters=200, warmup=50):
    for _ in range(warmup):
        feats.grad = None
        fn(s, t, feats).sum().backward()
    gc.collect(); torch.cuda.empty_cache()
    start = allocated_mb()
    for _ in range(n_iters):
        feats.grad = None
        out = fn(s, t, feats)
        assert torch.equal(out.detach(), ref), f"{name}: output changed"
        out.sum().backward()
    n_garbage = gc.collect(); torch.cuda.empty_cache()
    growth = allocated_mb() - start
    print(f"  {name:30s} {growth:+8.1f} MB / {n_iters} iters "
          f"({growth/n_iters:+.3f} MB/iter, gc freed {n_garbage})")

print(f"fvdb {fvdb.__version__} | torch {torch.__version__} | {torch.cuda.get_device_name()}")
src_ijk = torch.unique(torch.randint(0, 64, (15000, 3), dtype=torch.int32, device=DEV), dim=0)
tgt_ijk = src_ijk[: len(src_ijk) // 2].clone()
feats = torch.randn(len(src_ijk), 16, device=DEV, requires_grad=True)

ref = gather_index(src_ijk, tgt_ijk, feats).detach()
for name, fn in IMPLS:
    assert torch.equal(fn(src_ijk, tgt_ijk, feats).detach(), ref)

# inject_from measured last so it can't fragment the others' baselines
for name, fn in reversed(IMPLS):
    measure_leak(name, fn, src_ijk, tgt_ijk, feats, ref)

```

## Output
```bash
fvdb 0.5.0 | torch 2.12.1+cu130 | NVIDIA GeForce RTX 3090
  index gather (plain torch)         +0.4 MB / 200 iters (+0.002 MB/iter, gc freed 0)
  inject_from_ijk (ijk->grid)        +0.4 MB / 200 iters (+0.002 MB/iter, gc freed 0)
  inject_from (grid->grid)         +226.8 MB / 200 iters (+1.134 MB/iter, gc freed 0)

```

The ~0.4 MB on the two leak-free variants is a constant one-time allocation (identical at 100 and 200 iters). inject_from scales linearly with iteration count (113 MB at 100 iters → 227 MB at 200 iters), i.e. a genuine leak.

matthewdcong:
Python's GC has the ability to detect and correctly handle reference cycles where multiple objects refer to each other. For example, `a.foo = b` followed by `b.bar = a` leads to a refcount of 1 for both `a` and `b` and requires cycle handling in order to decrement their refcounts to zero for garbage collection.

It is possible for Python reference cycles to introduced via two C++ objects that refer to each other, where both have Python bindings. In that case, the wrapping Python objects both have their refcounts incremented as one would expect. However, without an implementation of `tp_traverse` and `tp_clear`, which provide the Python GC with the necessary introspection, Python cannot detect nor resolve these reference cycles and therefore both objects remain allocated.

This occurs when using `JaggedTensor` in an autograd context. In this case, it's possible to obtain a reference cycle involving the autograd context, the jagged tensor, and the underlying `mData` PyTorch tensor. To address this for `JaggedTensor`, we could implement `tp_traverse` and `tp_clear` via `pybind11::custom_type_setup` to enable the GC to detect and resolve that edge in the cycle. However, it is extremely subtle especially since `TensorImpl`'s bindings are non-standard. Moreover, it is insufficient to resolve #684 completely. This is because there is another edge from the `Tensor` object `dst_out` to the autograd `Node`/`PyNode`. Resolving this would require extending `tp_traverse`/`tp_clear` for the `torch.Tensor` type in the PyTorch codebase itself.

Thus, we work around the cycle by detaching `dst_jdata` and `src_jdata` and calling `jagged_like` with the detached copies in order to recreate the `impl` instead of modifying it in-place. This breaks the reference cycle and enables the correct GC and fixes the leak.

### DIFF
--- a/fvdb/functional/_dense.py
+++ b/fvdb/functional/_dense.py
@@ -100,8 +100,13 @@ class _InjectFn(torch.autograd.Function):
     def forward(ctx, src_jdata, dst_jdata, dst_grid_data, src_grid_data, dst_jt_impl, src_jt_impl):
         ctx.dst_grid_data = dst_grid_data
         ctx.src_grid_data = src_grid_data
-        ctx.dst_jt_impl = dst_jt_impl
-        ctx.src_jt_impl = src_jt_impl
+        # Keep independent structure carriers for backward.  The caller replaces
+        # dst_jt_impl.jdata with dst_out below, so retaining that same implementation
+        # here would form a reference cycle through dst_out's autograd context.
+        # Detached data preserves the device and leading dimension needed by
+        # jagged_like without retaining either input's autograd history.
+        ctx.dst_jt_impl = dst_jt_impl.jagged_like(dst_jdata.detach())
+        ctx.src_jt_impl = src_jt_impl.jagged_like(src_jdata.detach())
         # Clone dst so the op is out-of-place.  mark_dirty on non-contiguous
         # views disconnects _InjectFnBackward from the autograd graph (PyTorch
         # inserts AsStridedBackward → CopySlices that bypasses the custom
--- a/src/fvdb/JaggedTensor.h
+++ b/src/fvdb/JaggedTensor.h
@@ -376,18 +376,6 @@ class JaggedTensor : public torch::CustomClassHolder {
     /// @param data A data tensor with the same number of elements as the original data
     void set_jdata(const torch::Tensor &data);
 
-    /// @brief Reset all tensor members to undefined tensors.
-    ///        Intended for use by the Python cyclic GC tp_clear callback; leaves
-    ///        the JaggedTensor in an invalid state and must only be called when
-    ///        the object is being collected.
-    void
-    invalidate() {
-        mData     = torch::Tensor{};
-        mBatchIdx = torch::Tensor{};
-        mOffsets  = torch::Tensor{};
-        mListIdx  = torch::Tensor{};
-    }
-
     /// @brief  Get the raw data indexed by this JaggedTensor
     /// @return The raw data tensor
     const torch::Tensor &
--- a/src/fvdb/JaggedTensor.h
+++ b/src/fvdb/JaggedTensor.h
@@ -376,6 +376,18 @@ class JaggedTensor : public torch::CustomClassHolder {
     /// @param data A data tensor with the same number of elements as the original data
     void set_jdata(const torch::Tensor &data);
 
+    /// @brief Reset all tensor members to undefined tensors.
+    ///        Intended for use by the Python cyclic GC tp_clear callback; leaves
+    ///        the JaggedTensor in an invalid state and must only be called when
+    ///        the object is being collected.
+    void
+    invalidate() {
+        mData     = torch::Tensor{};
+        mBatchIdx = torch::Tensor{};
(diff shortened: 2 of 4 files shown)


matthewdcong:
The script in #684 no longer leaks after this PR.

swahtz (changes requested):
Looks great and explanation is great.  I think we should just add a test to make sure we don't regress.  Something like:

```
import gc, weakref
def test_inject_fn_no_leak():
    # set up a src/dst grid pair
    # run inject in an autograd context
    # hold only a weakref to the result, del it, collect, assert the weakref is dead
    ...
    gc.collect()
    assert result_ref() is None, "reference cycle leak detected"
```

matthewdcong:
Added a test that runs the autograd and checks that the GC works as expected

swahtz (approved):
lgtm!

PR #693, opened 2026-07-16
Prevent including private headers of nanovdb-editor

phapalova:
CPM downloads the entire nanovdb-editor repository, but only the headers in the top-level nanovdb_editor folder (PNanoVDBExt.h and putil/) are the public, versioned interface that fVDB is allowed to include as <nanovdb_editor/...>. Everything else (editor/, compute/, raster/, vulkan/, imgui/, ...) is internal implementation detail.

In https://github.com/openvdb/nanovdb-editor/pull/218 editor has added a build config to only copy the public headers. This PR adds this config and points the include directory to the build folder instead of to the CPM cache with the cloned repository.


harrism (commented):
Could `nanovdb-editor` export an `INTERFACE` target for the public headers under `PUBLIC_HEADERS_ONLY`, so fVDB links `nanovdb_editor::public_headers` instead of hard-coding `${nanovdb_editor_BINARY_DIR}/include`?

I think this would be the more cmake-idiomatic approach. This PR is reasonable, but I think it would be nicer as an interface library.

harrism (approved):
Thanks for this, and thanks for incorporating my target_link_libraries suggestion.

PR #694, opened 2026-07-16
Update editor version and adjust viewer

phapalova:
Fixes TODO added in https://github.com/openvdb/fvdb-core/pull/690

### DIFF
--- a/src/cmake/get_nanovdb_editor.cmake
+++ b/src/cmake/get_nanovdb_editor.cmake
@@ -10,11 +10,9 @@ set(NANOVDB_EDITOR_BUILD_TYPE "Release" CACHE STRING "Build type for nanovdb_edi
 
 # fVDB pins NanoVDB Editor headers to an exact source commit via CPM
 # Note: only change this when the interfaces have changed
-# 7efb611: adds add_nanovdb_3 API and pnanovdb_pipeline_type_nanovdb_surface = 12
-#          required for add_level_set / add_fog_volume viewer methods.
 # 39fb452: adds get_pipeline_type() used to resolve pnanovdb_pipeline_type_t
 set(NANOVDB_EDITOR_TAG 39fb4523db9e4dac50f2e2ac54c3898152101c14)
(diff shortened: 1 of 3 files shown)


src/cmake/get_nanovdb_editor.cmake
```diff
@@ -10,11 +10,9 @@ set(NANOVDB_EDITOR_BUILD_TYPE "Release" CACHE STRING "Build type for nanovdb_edi
 
 # fVDB pins NanoVDB Editor headers to an exact source commit via CPM
 # Note: only change this when the interfaces have changed
-# 7efb611: adds add_nanovdb_3 API and pnanovdb_pipeline_type_nanovdb_surface = 12
-#          required for add_level_set / add_fog_volume viewer methods.
 # 39fb452: adds get_pipeline_type() used to resolve pnanovdb_pipeline_type_t
 set(NANOVDB_EDITOR_TAG 39fb4523db9e4dac50f2e2ac54c3898152101c14)
-set(NANOVDB_EDITOR_VERSION 0.1.5)   # version at this commit
+set(NANOVDB_EDITOR_VERSION 0.1.6)   # version at this commit
```
swahtz, line 15: **❓ question:** ‏Isn't `39fb4523db9e4dac50f2e2ac54c3898152101c14` the 0.1.5 release?

swahtz (approved):
Built this and it worked well for me.  Only question is whether we bump the `NANOVDB_EDITOR_TAG` to the 0.1.6 version or change the version back down to 0.1.5

src/cmake/get_nanovdb_editor.cmake
phapalova: The tag needs to be updated only when the interface changes to prevent the whole fvdb recompile, it gets pulledd from the repository and not from the python wheel.

phapalova: Now I see why this is confusing, there is the comment: `# version at this commit`, that is stale from times when the nanovdb-editor wheel wasn't used for the fvdb build. I will adjust that, it should be >= version on that tag.

PR #697, opened 2026-07-20
Reduce atomic contention for SH backwards pass

matthewdcong:
#687 avoided materializing the view directions by passing the viewmats to the spherical harmonics operator directly. However, this can result in significant atomics contention proportional O(N * C) to as a large number of Gaussians can atomically contribute to many camera matrices. The fix is to materialize the view direction gradient inside the backwards pass of the operator only followed by a block reduce and finally a atomic update the the viewmat gradient.

`frgs reconstruct data/360_v2/bicycle/ --device cuda --io.no-save-checkpoints --cfg.batch-size 8` for the first 20% of epochs becomes 6% faster after this change.

### DIFF
--- a/src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
+++ b/src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
@@ -399,17 +399,23 @@ template <typename T> struct AddViewDirectionGradients {
     }
 };
 
-template <bool UseSystemAtomics, typename T>
-__device__ inline void
-storeWorldToCamMatrixGradient(T *__restrict__ output, const int64_t index, const T value) {
-    if constexpr (UseSystemAtomics) {
-        atomicAdd_system(output + index, value);
-    } else {
-        output[index] = value;
+struct AssignOutput {
+    template <typename T>
+    __device__ static void
+    update(T *__restrict__ output, const T value) {
+        *output = value;
     }
-}
+};
 
-template <bool UseSystemAtomics, typename T>
+struct SystemAtomicAddOutput {
+    template <typename T>
+    __device__ static void
+    update(T *__restrict__ output, const T value) {
+        atomicAdd_system(output, value);
+    }
+};
+
+template <typename OutputUpdatePolicy, typename T>
 __device__ inline void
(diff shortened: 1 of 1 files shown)


src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -377,46 +378,154 @@ computeShBackward(
...
+    storeWorldToCamMatrixGradient<UseSystemAtomics>(output,
+                                                    11,
+                                                    worldToCamMatrix[8] * gradient.x +
+                                                        worldToCamMatrix[9] * gradient.y +
+                                                        worldToCamMatrix[10] * gradient.z);
+}
+
+template <typename T>
+__global__ __launch_bounds__(DEFAULT_BLOCK_DIM) void
+reduceUnpackedViewDirectionGradientsToWorldToCamMatrices(
```
swahtz, line 448: **🎵 note (non-blocking):** ‏It's probably not a big difference in speed but perhaps we should template this on the accumulation scope and only use the system atomics for `PrivateUse1` and direct stores for single GPU which should be sufficient.  It's only 12 per-camera so probably not a big win.

swahtz (approved):
Looks great to me, thanks for picking up on the speed difference!

Maybe the only suggestion is the usage of direct assignment in the non-`PrivateUse1` case in `reduceUnpackedViewDirectionGradientsToWorldToCamMatrices` but don't want to block on that.

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
matthewdcong: Thanks for the suggestion! It's not a noticeable difference in speed (and in fact atomics can actually be faster that direct stores) so I'd probably prefer to just keep it as is but I can take a look.

matthewdcong: Fixed!

swahtz: Really?  I always thought atomics are going to be slower than a direct store, maybe I need to go learn more about what they do under the hood.

PR #700, opened 2026-07-23
Drop support for PyTorch < 2.6

matthewdcong:
PyTorch 2.6 was released in January 2025 and introduced maps for strings <-> dtypes and a Python typecaster for dtypes. This PR removes the compatibility wrappers necessary for PyTorch versions prior to that thus dropping support for PyTorch < 2.6.

This also changes our pyproject.toml to require torch >= 2.7 instead of 2.8 based on the newly discussed policy.


swahtz (approved):
Looks great, I'm always happy to remove code and dropping support for versions older than 2.6 seems reasonable to me.  Thanks!

harrism:
Do any docs need to be updated? And fvdb-reality-capture?

swahtz:
I've added some documentation in #731 to try and clarify what versions we mention in our docs.  I think we should also bump the `pyproject.toml` pytorch minimum to 2.7 based on our conversations.

PR #701, opened 2026-07-23
Unit Test GPU RAM Reductions

swahtz:
Minor changes to unit tests so that tests run comfortably on a fractional L4 GPU with 6GB of VRAM.

Some tests changed hyper parameter values because the numbers chosen were too large (i.e. the unit test needs 'some significant number' but the test doesn't depend on the number being a specific large value) and a few tests (namely in `test_accessors.py`) are specifically to test that functionality works if the parameters are past some limit.  For the latter, I added documentation so that's clear and made the tests more robust (rather than simply testing they do not crash).

Per @matthewdcong's suggestion, the `conftest.py` periodic `empty_cache()` fixture is replaced by setting `PYTORCH_NO_CUDA_MEMORY_CACHING=1` on the CI unit test steps (`tests.yml`, `cu128.yml`, `cu130.yml`, `publish.yml`). The whole parametrized matrix shares one process, so the caching allocator's reserved pool only ever grows and its high-water mark ends up far above any single test's live footprint. Setting it in the workflows rather than in `conftest.py` means CI pays the cost (a `cudaMalloc`/`cudaFree` per allocation) where VRAM is the constraint, while local test runs keep the caching allocator and its speed.

### DIFF
--- a/.github/workflows/cu128.yml
+++ b/.github/workflows/cu128.yml
@@ -445,6 +445,14 @@ jobs:
           pip install ./dist/*.whl
 
       - name: Run tests
+        env:
+          # The whole parametrized matrix shares one process, so the caching allocator's
+          # reserved pool only ever grows and its high-water mark ends up far above any
+          # single test's live footprint. Freeing straight back to the driver keeps the
+          # suite within the runner's VRAM, at the cost of a cudaMalloc/cudaFree per
+          # allocation. See
+          # https://docs.pytorch.org/docs/stable/notes/cuda.html#optimizing-memory-usage-with-pytorch-alloc-conf
+          PYTORCH_NO_CUDA_MEMORY_CACHING: 1
         run: |
           source .venv/bin/activate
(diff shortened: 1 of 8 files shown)


harrism (approved):

tests/conftest.py
```diff
@@ -1,6 +1,36 @@
 # Copyright Contributors to the OpenVDB Project
 # SPDX-License-Identifier: Apache-2.0
 
+import gc
```
matthewdcong: Would setting `PYTORCH_NO_CUDA_MEMORY_CACHING=1` be a better solution here? See https://docs.pytorch.org/docs/2.13/notes/cuda.html#optimizing-memory-usage-with-pytorch-alloc-conf

swahtz: Hmmm yes, that's probably the more elegant and simplest solution to this.

swahtz: I have removed these changes in `conftest.py` and now we just set `PYTORCH_NO_CUDA_MEMORY_CACHING` at the Github Action job definitions

PR #703, opened 2026-07-23
Reduce UVM cross-device UVM traffic in intersection op

matthewdcong:
#664 changed the intersection keys to be computed independently for each device. Now, instead of concatenating the per-device intersection keys into a single tensor in managed memory, we keep them local to each device. Moreover, we can modify `computeTileOffsets` to operate on these device-local key ranges (including empty ranges) without reading the preceeding keys from neighboring devices since the number of intersections on each device is known a priori.

As a result, UVM traffic decreases by 12% end-to-end and about 30% within the intersection op, leading to about a 3% speedup on 2 GPUs. Single GPU behavior is unchanged.

### DIFF
--- a/src/fvdb/detail/ops/gsplat/IntersectGaussianTiles.cu
+++ b/src/fvdb/detail/ops/gsplat/IntersectGaussianTiles.cu
@@ -866,19 +866,14 @@ intersectGaussianTilesPrivateUse1Impl(
         torch::Tensor tileJOffsets = torch::empty({numCameras, numTilesH, numTilesW},
                                                   means2d.options().dtype(torch::kInt32));
 
-        std::vector<cudaEvent_t> events(deviceCount);
         for (const auto deviceId: c10::irange(deviceCount)) {
             C10_CUDA_CHECK(cudaSetDevice(deviceId));
-            auto stream                     = c10::cuda::getCurrentCUDAStream(deviceId);
-            const int64_t intersectionCount = deviceIntersectionCount[deviceId];
-            if (intersectionCount > 0) {
-                C10_CUDA_CHECK(
-                    cudaMallocAsync(&deviceIntersectionKeys[deviceId],
-                                    static_cast<size_t>(intersectionCount) * sizeof(int64_t),
-                                    stream));
+            auto stream = c10::cuda::getCurrentCUDAStream(deviceId);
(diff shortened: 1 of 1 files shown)


src/fvdb/detail/ops/gsplat/IntersectGaussianTiles.cu
```diff
@@ -847,72 +856,43 @@ intersectGaussianTilesPrivateUse1Impl(
...
                                                   means2d.options().dtype(torch::kInt32));
 
-        std::vector<cudaEvent_t> events(deviceCount);
         for (const auto deviceId: c10::irange(deviceCount)) {
             C10_CUDA_CHECK(cudaSetDevice(deviceId));
             auto stream = c10::cuda::getCurrentCUDAStream(deviceId);
-            C10_CUDA_CHECK(cudaEventCreateWithFlags(&events[deviceId], cudaEventDisableTiming));
-            C10_CUDA_CHECK(cudaEventRecord(events[deviceId], stream));
+            if (deviceIntersectionCount[deviceId] > 0) {
+                C10_CUDA_CHECK(cudaMallocAsync(&deviceIntersectionKeys[deviceId],
```
harrism, line 882: **🚩 issue:** ‏I don't think we have a lot of raw allocations in fvdb-core like this -- most allocations use Torch tensors. I would prefer to avoid raw allocation. Ultimately, we should move to a memory resource model like we have in NanoVDB. I won't block this PR for this, but please add a TODO comment.

harrism (approved):
Please add a TODO to remove the raw allocation. Otherwise looks good.

src/fvdb/detail/ops/gsplat/IntersectGaussianTiles.cu
matthewdcong: This raw allocation is in the mGPU codepath only and intentional. We can't use a `torch::Tensor` with an explicit `kCUDA` device here because:

1. PyTorch doesn't support mixing the use of multiple accelerators/accelerator backends, i.e. `kCUDA` and `kPrivateUse1`. It may work in some limited regard, but it often breaks between versions in subtle ways.
2. Even if it was supported, you can run into the case where the `CUDACachingAllocator` holds onto device memory blocks and ends up crowding out the unified memory pool allocations which leads to OOM errors.

I'll add a comment that explains this limitation as well as a TODO.

PR #704, opened 2026-07-25
Enable a user override for NVCC_THREADS

matthewdcong:
Currently, the `NVCC_THREADS` variable is entirely determined by the `TORCH_CUDA_ARCH_LIST`. However, when compiling for multiple archs (in the case of reusable Dockerfiles), `NVCC_THREADS` can become large which then leads to heavy quantization when calculating `CMAKE_BUILD_PARALLEL_LEVEL`. For example, when compiling for six different architectures on a machine with nproc = 10, this results in a non-parallel build (`MAX_JOBS = 1`) which degrades build performance overall since host code compilation and the initial nvcc calls is no longer parallelized.

The solution is to enable the user to set `NVCC_THREADS` in their environment like `CMAKE_BUILD_PARALLEL_LEVEL` and use that in `./build.sh` if set.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -53,20 +53,27 @@ setup_parallel_build_jobs() {
   if [ "$NUM_ARCH" -lt 1 ]; then
     NUM_ARCH=1
   fi
-  NVCC_THREADS=$NUM_ARCH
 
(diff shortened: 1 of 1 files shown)


swahtz (approved):
lgtm

PR #705, opened 2026-07-27
CI:  Use all 3 us-east-2 availability zones for starting runners

swahtz:
Change all AWS EC2 start actions to use fallback strategy with all 3 us-east-2 availability zones for runners.  Conformed GPU runner type across jobs.


matthewdcong (approved):
Amazing, this fixes the issue I've been running into. Thanks!

PR #706, opened 2026-07-28
Fix race condition in mGPU backwards rasterization

matthewdcong:
When a camera is rendered on multiple GPUs, every GPU clears/memsets the _entire_ output tensor as opposed to just the tiles that it owns. This results in a race condition with the subsequent per-GPU gradient calculation for the tiles that it owns. That is, another GPU rendering from the same camera could clear the output tensor while another GPU accumulates the gradients.

### DIFF
--- a/src/fvdb/detail/ops/gsplat/RasterizeScreenSpaceGaussiansBackward.cu
+++ b/src/fvdb/detail/ops/gsplat/RasterizeScreenSpaceGaussiansBackward.cu
@@ -1172,13 +1172,15 @@ callRasterizeBackwardPrivateUse1(
     const std::optional<torch::Tensor> &pixelMap        = std::nullopt) {
     TORCH_CHECK(tileSize > 0, "Tile size must be greater than 0");
 
-    torch::Tensor outDLossDMeans2d   = torch::empty_like(means2d);
-    torch::Tensor outDLossDConics    = torch::empty_like(conics);
-    torch::Tensor outDLossDFeatures  = torch::empty_like(features);
-    torch::Tensor outDLossDOpacities = torch::empty_like(opacities);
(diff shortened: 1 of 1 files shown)


src/fvdb/detail/ops/gsplat/RasterizeScreenSpaceGaussiansBackward.cu
```diff
@@ -1172,13 +1172,15 @@ callRasterizeBackwardPrivateUse1(
...
-    torch::Tensor outDLossDMeans2d   = torch::empty_like(means2d);
-    torch::Tensor outDLossDConics    = torch::empty_like(conics);
-    torch::Tensor outDLossDFeatures  = torch::empty_like(features);
-    torch::Tensor outDLossDOpacities = torch::empty_like(opacities);
+    // These are shared accumulation buffers on PrivateUse1. Initialize them before partitioning
+    // tile work so each buffer is cleared exactly once before any device starts accumulating.
+    torch::Tensor outDLossDMeans2d   = torch::zeros_like(means2d);
+    torch::Tensor outDLossDConics    = torch::zeros_like(conics);
+    torch::Tensor outDLossDFeatures  = torch::zeros_like(features);
+    torch::Tensor outDLossDOpacities = torch::zeros_like(opacities);
```
swahtz, line 1180: **❓ question:** ‏Does `torch::zeros_like` provide a cross-device synchronization guarantee?  Do we need to put a synch/barrier after the zeroing?

matthewdcong: Yes, it provides a synchronization guarantee. No need for an explicit barrier/sync.

swahtz (approved):
lgtm, thanks for answering my question

Issue #73, opened 2025-09-17
[BUILD] CUDA Unsupported Operation on OVX VM


swahtz:
## 📋 Originally created by @malam-mitre

> **Original author:** @malam-mitre
> **Date:** 2025-07-09T21:20:26Z

---

### Environment
**Operating System:** VM Ubuntu 22.04 / Container Ubuntu 24.04
**CMake Version:** 4.03
**Compiler:** gcc 11.4.0
**Container Image:** condaforge/miniforge3:24.11.3-2

### Describe the problem
When calling the nanovdb PointsToGrid function on a VMWare managed virtual machine on an OVX machine, we are receiving the following error:

> CUDA error 801: operation not supported (/<path to openvdb>/openvdb/fvdb/src/../../nanovdb/nanovdb/tools/cuda/PointsToGrid.cuh:401)

Please note that we have seen something similar with CudaMallocManaged on this system and it may be due to the usage of a vGPU. 

### To Reproduce
1. Make a new directory. 
2. Download the attached zip file and unzip the contents here.
3. Clone openvdb repository to this folder at the feature/fvdb branch
4. Docker build - `docker compose run --rm fvdb-dev`
5. Change to the fvdb directory `cd openvdb/fvdb`
6. Build fvdb `./build.sh wheel` 
7. Add fvdb to pip `pip install .`
8. Test using any operation that calls the PointsToGrid function. While I am unable to share our specific example, I will try to provide an example later outside of our use-case that utilizes PointsToGrid to ease testing. 

### Additional context
The attached zip file contains a dev_environment.yml, a docker-compose.yml, and a Dockerfile to help replicate the container we are using. This mostly builds on the fVDB docker by including an isaac sim installation. 

[fvdb_ovx_build_issue.zip](https://github.com/user-attachments/files/21150267/fvdb_ovx_build_issue.zip)

PR #709, opened 2026-07-29
CI: g6f instances

swahtz:
## Summary

Moves all three pull-request test workflows onto fractional-GPU (vGPU) EC2 instances and
adds the build option those instances require.  Note:  A successful run on the g6f instance type run off of this branch can be found here: https://github.com/openvdb/fvdb-core/actions/runs/30422551926

`cudaMallocAsync` depends on GPU unified memory, which is not available on a vGPU slice —
it fails at runtime with `CUDA error 801: operation not supported`. NanoVDB selects between
`cudaMallocAsync` and plain `cudaMalloc` at compile time via `NANOVDB_USE_SYNC_CUDA_MALLOC`,
so testing on fractional GPUs requires compiling with that macro defined.

### New build option

- `FVDB_USE_SYNC_CUDA_MALLOC` CMake option (default `OFF`, so the default build is unchanged)
  in `src/cmake/get_nanovdb.cmake`, which sets `NANOVDB_USE_SYNC_CUDA_MALLOC` through
  `add_compile_definitions`. The macro is consumed in a header by an inline function, so it
  has to be visible to every C++/CUDA translation unit — a per-target definition would give
  the library, bindings, tests, and benchmarks inconsistent inline bodies.
- `./build.sh sync_cuda_malloc` build modifier that forwards the CMake define, documented in
  the build-modifier list in `README.md`.

### CI runner configuration

- `fgpu_ami` added to `.github/versions.json`, exposed by `load-versions.yml` as the new
  `aws-fgpu-ami` and `aws-fgpu-az-config` outputs, following the existing CPU/GPU pattern of
  fanning the AMI across every subnet for availability.
- `tests.yml`, `cu128.yml`, and `cu130.yml` now start `g6f.2xlarge` (8 vCPU, 1/4 L4) test
  runners instead of `g6.xlarge` (4 vCPU, full L4), and build with `sync_cuda_malloc`.

### Restoring async-allocator coverage

Because every PR now tests the synchronous allocator, `nightly-publish.yml` gains a GPU test
stage on a full `g6.xlarge` that installs the published nightly wheel — built with the default
asynchronous allocator — and runs the Python unit tests once a night. This is purely additive:
the wheel build matrix is untouched, so no build times or artifacts change. A new
`cuda-default-patch` output pins the test container to the same CUDA patch version as the
build container rather than hardcoding it.

The nightly stage runs only the Python unit tests, not the gtests, since the gtests already
run on every PR and exercise the same kernels.

Relates to #73, which reports this exact `CUDA error 801` failure on an OVX/vGPU VM. The new
`sync_cuda_malloc` modifier gives affected users a supported way to build for those systems.

### DIFF
--- a/.github/versions.json
+++ b/.github/versions.json
@@ -35,6 +35,7 @@
     "region": "us-east-2",
     "cpu_ami": "ami-0e14a711dad782a70",
     "gpu_ami": "ami-0f5c0b65fde08ae43",
+    "fgpu_ami": "ami-0b5dca767eafff9d8",
     "subnet": "subnet-03f2320d6e6e0005b",
     "security_group": "sg-0cd08bd89d6212223",
     "cloudfront_distribution_domain": "d36m13axqqhiit.cloudfront.net"
--- a/.github/workflows/cu128.yml
+++ b/.github/workflows/cu128.yml
@@ -166,7 +166,9 @@ jobs:
           source .venv/bin/activate
           NANOVDB_EDITOR_SPEC=$(python .github/scripts/get_viewer_dependency.py)
           pip install "$NANOVDB_EDITOR_SPEC"
-          ./build.sh wheel verbose gtests benchmarks --cuda-arch-list "${NEEDS_VERSIONS_OUTPUTS_CUDA_ARCH_PR}"
+          # sync_cuda_malloc: the test runners are fractional-GPU (vGPU) instances
+          # where cudaMallocAsync is unavailable, so NanoVDB must use cudaMalloc.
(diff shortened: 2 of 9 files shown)


harrism (approved):
Looks good, and useful.

swahtz:
After discussion with @harrism, we've decided that regular PR CI testing that doesn't use the default malloc path (async) is an unnecessary risk and unless contention or cost become a big issue, we will leave the current runner configuration as-is.  Though it is good to know how to implement this for the future.

PR #713, opened 2026-08-02
Use intermediate device local gradient buffers for backwards rasterization

matthewdcong:
Currently, backwards rasterization for mGPU suffers degraded performance due to atomics and page contention. Fix this by switching to device local buffers for gradient accumulation followed by a NCCL `ReduceScatter` operation to compute the final accumulated output across GPUs in managed memory.

Even on the small MipNeRF 360 example training for 10 epochs, this reduces the runtime by 35%.

### DIFF
--- a/src/fvdb/detail/ops/gsplat/RasterizeScreenSpaceGaussiansBackward.cu
+++ b/src/fvdb/detail/ops/gsplat/RasterizeScreenSpaceGaussiansBackward.cu
@@ -12,8 +12,10 @@
 #include <fvdb/detail/utils/gsplat/GaussianRasterize.cuh>
 
 #include <ATen/cuda/Atomic.cuh>
+#include <ATen/ops/from_blob.h>
 #include <c10/core/DeviceType.h>
 #include <c10/cuda/CUDAGuard.h>
+#include <torch/csrc/cuda/nccl.h>
 
 #include <cooperative_groups.h>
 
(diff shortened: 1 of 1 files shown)


src/fvdb/detail/ops/gsplat/RasterizeScreenSpaceGaussiansBackward.cu
```diff
@@ -1270,6 +1334,51 @@ callRasterizeBackwardPrivateUse1(
...
+        const size_t conicsBytes    = conics.numel() * conics.element_size();
+        const size_t featuresBytes  = features.numel() * features.element_size();
+        const size_t opacitiesBytes = opacities.numel() * opacities.element_size();
+
+        ScalarType *&outDLossDMeans2DLocalPtr = outDLossDMeans2DLocalPtrs[deviceId];
+        C10_CUDA_CHECK(cudaMallocAsync(&outDLossDMeans2DLocalPtr, means2dBytes, stream));
+        C10_CUDA_CHECK(cudaMemsetAsync(outDLossDMeans2DLocalPtr, 0, means2dBytes, stream));
+        torch::Tensor &outDLossDMeans2DLocal = outDLossDMeans2DLocals[deviceId];
+        outDLossDMeans2DLocal =
+            at::from_blob(outDLossDMeans2DLocalPtr, means2d.sizes(), localTensorOptions);
```
swahtz, line 1350: **❓ question:** ‏I'm wondering, why do we need to allocate the tensors like this, can't we do something like:

```
c10::cuda::CUDAGuard deviceGuard(deviceId);
c10::cuda::CUDAStreamGuard streamGuard(stream);
auto outDLossDMeans2DLocal = at::zeros(means2d.sizes(), localTensorOptions);
// more tensors
```

And avoid the state where a potential error might leave us in where we don't hit the `cudaFreeAsync` block below?

swahtz (approved):
lgtm, just one question but don't want to block

src/fvdb/detail/ops/gsplat/RasterizeScreenSpaceGaussiansBackward.cu
matthewdcong: See https://github.com/openvdb/fvdb-core/pull/703#discussion_r3653757064

swahtz: Ah I see, ok, too bad.

PR #714, opened 2026-08-02
Improve compile times for forward rasterization

matthewdcong:
An unintentional side effect of #554 was that the compile times for the forward rasterization TU increased dramatically. This is because the output features array was accidentally specified using `NUM_CHANNELS` instead of the (much) smaller `NUM_SHARED_CHANNELS`, the latter of which mirrors the chunking in the backwards pass. As a result, for each specialization, NVCC needed to unroll the full output values array which resulted in additional duplicated state and unrolling. By switching `pixOut` to `NUM_SHARED_CHANNELS` instead of `NUM_CHANNELS` and writing directly to the slice, we significantly improve compile times and reduce the amount of stack and SASS:

  - Stack: 2,056 → 64 bytes
  - Total rasterizer SASS: 13.7 → 8.3 MB
  - Compile: 616 → 181 seconds

### DIFF
--- a/src/fvdb/detail/ops/gsplat/RasterizeScreenSpaceGaussiansForward.cu
+++ b/src/fvdb/detail/ops/gsplat/RasterizeScreenSpaceGaussiansForward.cu
@@ -99,15 +99,23 @@ struct RasterizeForwardArgs {
         mOutLastIds.data()[pixelIndex] = lastId;
     }
 
-    /// @brief Write the features for a pixel
+    /// @brief Write a chunk of features for a pixel
     /// @param pixelIndex The index of the pixel
+    /// @param channelStart The first output channel in the chunk
+    /// @param numChannels The number of output channels in the chunk
     /// @param f The function to write the features
     template <typename F>
     __device__ void
-    writeFeatures(uint64_t pixelIndex, F &&f) {
-#pragma unroll
-        for (uint32_t k = 0; k < NUM_CHANNELS; ++k) {
-            mOutFeatures.data()[pixelIndex][k] = f(k);
+    writeFeatureChunk(uint64_t pixelIndex, size_t channelStart, size_t numChannels, F &&f) {
+        if constexpr (IS_CHUNKED) {
+            for (uint32_t k = 0; k < numChannels; ++k) {
+                mOutFeatures.data()[pixelIndex][channelStart + k] = f(k);
+            }
+        } else {
+#pragma unroll NUM_CHANNELS
+            for (uint32_t k = 0; k < NUM_CHANNELS; ++k) {
+                mOutFeatures.data()[pixelIndex][k] = f(k);
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/fvdb/detail/ops/gsplat/RasterizeScreenSpaceGaussiansForward.cu
```diff
@@ -281,23 +286,20 @@ struct RasterizeForwardArgs {
                 }
             }
 
-            if (chunk == 0) {
-                accumTransmittanceFinal     = accumTransmittance;
-                lastIntersectionOffsetFinal = lastIntersectionOffset;
+            if (pixelIsActive) {
+                const auto pixIdx = commonArgs.pixelIndex(cameraId, row, col, activePixelIndex);
+                if (chunk == 0) {
```
swahtz, line 296: **💡 suggestion:** ‏Could we put the comment that was at line 184 down here (the one that reads "The alpha sequence is feature-independent (depends only on opacity + conic + position), so the per-chunk accumTransmittance and last intersection offset are identical across chunks.") so we preserve the explanation here?

src/fvdb/detail/ops/gsplat/RasterizeScreenSpaceGaussiansForward.cu
```diff
@@ -99,15 +99,23 @@ struct RasterizeForwardArgs {
...
-    writeFeatures(uint64_t pixelIndex, F &&f) {
-#pragma unroll
-        for (uint32_t k = 0; k < NUM_CHANNELS; ++k) {
-            mOutFeatures.data()[pixelIndex][k] = f(k);
+    writeFeatureChunk(uint64_t pixelIndex, size_t channelStart, size_t numChannels, F &&f) {
+        if constexpr (IS_CHUNKED) {
+            for (uint32_t k = 0; k < numChannels; ++k) {
+                mOutFeatures.data()[pixelIndex][channelStart + k] = f(k);
+            }
+        } else {
```
swahtz, line 114: **🚩 issue:** ‏When `IS_CHUNKED` is false, we ignore `channelStart` and `numChannels`.  Looking around, every time we currently call this in this scenario, we pass `channelStart=0` and `numChannels==NUM_CHANNELS` so it's all good but I think we should add a check to ensure no incorrect use in the future.

```suggestion
        } else {
               assert(channelStart == 0 && numChannels == NUM_CHANNELS);
```

src/fvdb/detail/ops/gsplat/RasterizeScreenSpaceGaussiansForward.cu
```diff
@@ -281,23 +286,20 @@ struct RasterizeForwardArgs {
                 }
             }
 
-            if (chunk == 0) {
-                accumTransmittanceFinal     = accumTransmittance;
-                lastIntersectionOffsetFinal = lastIntersectionOffset;
+            if (pixelIsActive) {
+                const auto pixIdx = commonArgs.pixelIndex(cameraId, row, col, activePixelIndex);
```
swahtz: **💡 suggestion:**  We now calculate `pixIdx` inside the chunk loop (whereas before it was done afterwards).  We could do this before the chunked loop to avoid redundant work since none of `pixIdx` depends on values in the loop.‏

src/tests/GaussianRasterizeForwardTest.cpp
```diff
@@ -594,11 +594,36 @@ TEST_F(GaussianRasterizeForwardTestFixture, TestConcatenatedChannels) {
                                                             static_cast<uint32_t>(imageOriginH),
                                                             tileSize,
                                                             tileOffsets,
-                                                            tileGaussianIds);
+                                                            tileGaussianIds,
+                                                            16); // force chunked rendering
```
swahtz: **🧹 chore:** ‏Forcing this to 16 forces chunked rendering but then we lose coverage for non-chunked.  In these two tests, could we parameterize the test over 16, 32, 64 channels or just have another test with the non-chunked?

swahtz (approved):
This is great, thanks for catching that.  Just a few minor comments, but no big changes requested.

PR #715, opened 2026-08-03
Fix CPU prune crash on batches with a single-voxel grid

swahtz:
`dispatchPruneGrid<kCPU>` read each grid's mask with `mask.index(bidx).jdata().squeeze()`. For a grid with exactly one voxel the mask list is shape [1], and `squeeze()` collapses it to a 0-D scalar, so the following `.accessor<bool, 1>()` threw `"TensorAccessor expected 1 dims but tensor has 0"`. Any `pruned_grid(mask)` on a CPU batch containing a single-voxel grid hit this.

Use `reshape({-1})` instead: it flattens a [N, 1] mask to [N] as intended while keeping a [1] mask 1-D. CUDA is unaffected (it passes the raw mask data pointer, no squeeze).


blackencino (approved):
Good catch.

PR #726, opened 2026-08-04
Unify sparse convolution and transpose semantics

### LINKED ISSUE #668: Conv Transposed Issue?
Hi everyone, thank you for the awesome project!

I am trying to understand the behavior of fvnn.SparseConvTranspose3d and ran into an unexpected result.
I defined a simple down-up block that downsamples a grid (stride=4) and then upsamples it back to the original topology:

```
class SimpleNet(nn.Module):
    def __init__(self, in_channels=3, dims=3):
        super().__init__()
        self.downconv = fvnn.SparseConv3d(in_channels, dims, kernel_size=4, stride=4, bias=False)
        self.upconv = fvnn.SparseConvTranspose3d(dims, in_channels, kernel_size=4, stride=4, bias=False)
     
    def forward(self, x, grid):
        coarse_grid = grid.conv_grid(4, 4)
        down_plan = fvdb.ConvolutionPlan.from_grid_batch(kernel_size=4, stride=4, source_grid=grid, target_grid=coarse_grid)
        up_plan = fvdb.ConvolutionPlan.from_grid_batch_transposed(kernel_size=4, stride=4, source_grid=coarse_grid, target_grid=grid
        )
        x = self.downconv(x, down_plan)
        x = self.upconv(x, up_plan)
        return x
```

My expectation was that the up-convolution would populate features for all the voxels in the original grid. However, it seems that some voxels on the output grid are completely unaffected by the transposed convolution, resulting in 0.0 values.

Image

Is this sparse feature drop-off expected behavior for this configuration, or could this be a bug in how the voxel mapping is handled?

Thanks in advance for your help!

blackencino:
Fixes #668

## Summary

- Define one Torch-phase sparse convolution relation for generated topology, execution, transpose, autograd, grid registration, and backend admission.
- Generate complete forward and transposed structural topology from that relation, while treating explicit targets as restrictions of the same relation.
- Reverse stored finite rulebooks exactly with `ConvolutionPlan.from_plan_transposed` instead of rebuilding topology.
- Preserve the proved direct `K=S=2` path and use a phase-aware direct `O(N)` projection for dilation-one componentwise `K=S`; use count-then-fill `O(N+M)` staging elsewhere.
- Add independent scalar and dense Torch oracles, even/mixed-kernel value and gradient coverage, multi-grid CUDA coverage, resource gates, migration notes, and backend boundary checks.
- Make coverage diagnostics lazy and shared by exact transposes.
- Expose typed topology and phase policies, coverage diagnostics, a named coverage warning, and typed builder resource statistics.

For the reported solid `16^3`, `K=S=4` case, generated forward topology is now `5^3 = 125` coarse voxels, and transposition back onto the original grid leaves zero unreachable rows rather than 721.

## Design policy

The canonical componentwise relation is:

```text
fine_ijk = stride * coarse_ijk + tap_ijk - padding_before
padding_before = floor((kernel_size - 1) / 2)
```

Here `fine_ijk` is a coordinate on the fine lattice, `coarse_ijk` is a coordinate on the strided lattice, and `tap_ijk` is a zero-based kernel tap satisfying `0 <= tap_ijk[axis] < kernel_size[axis]`.

Generated topology is the positive structural topology of this relation. A supplied target grid restricts its output domain; it does not define another kernel phase or guarantee inverse behavior. The exact transpose of a concrete plan reverses that plan's stored edge set. A tied weighted adjoint additionally transposes the channel dimensions of the weights.

Generated strided grids preserve the canonical origin and scale voxel size by stride. `coarsened_grid` retains its separate block-centroid semantics.

## CUDA allocation behavior

The generated-transpose coordinate fallback validates the exact overflow-safe `16*N*kernel_volume` staging size before allocation, then lets PyTorch's allocator decide whether the request fits. Leaf-mask and unshifted-subdivision paths do not advertise coordinate-staging context because they do not allocate that staging buffer. If coordinate allocation raises `OutOfMemoryError`, fVDB preserves that exception type and adds the exact staging requirement and remediation context.

This avoids false rejection when reusable memory resides in inactive split cache. The regression test recreates that allocator state in an isolated process and verifies that a public `conv_transpose_grid` request succeeds when the allocator can serve it.

## Compatibility

This intentionally changes generated topology for affected even kernels, `K=S>=3`, and `K=1,S>1`, and corrects generated convolution-grid transforms. Weight tap ordering is unchanged. Serialized or in-memory plan caches must be invalidated when upgrading.

No legacy geometry mode is included. The previous implementation used truncating integer division for negative coordinates and duplicated the even-kernel phase derivation across several builders, so retaining it would preserve incorrect topology rather than a valid alternative convention.

Generative transpose still has inherent `O(N*K_volume)` output staging. Genuinely large constructions can still fail with an allocator OOM; the failure is allocator-authoritative and diagnostic rather than predictively rejected.

## Verification

- `./build.sh install gtests`: passed
- Focused convolution semantics suites: 203/203 passed on CPU and CUDA
- Full Python unit suite: 2,980 passed, 10 skipped
- C++ test suite: 41/41 passed
- `./build.sh docstest`: 112/112 passed
- Black 120-column checks, clang-format 18, and `git diff --check`: passed
- Warning-strict Sphinx rendered the changed pages; the global command reports one unrelated duplicate-target warning in unchanged `docs/installation.rst`


harrism:
> The canonical componentwise relation is:

```
p = stride * q + u - floor((kernel_size - 1) / 2)
```

What are p, q and u?

docs/TEACHME/fvdb_core_cheatsheet.md
```diff
@@ -233,6 +233,25 @@ feat_fine = up(feat_coarse, plan_up)         # JaggedTensor on fine_grid
 head = fvnn.SparseConv3d(in_channels, out_channels, kernel_size=1).cuda()
 ```
 
+### Convolution semantics and migration
```
harrism: **❓ question:** ‏What do you mean by "migration" here? Is it relevant to the TEACHME?

blackencino: It's a misnomer, held over from the build plan which I removed. Changing to just "semantics".

blackencino: FIxed in new commit

blackencino:
@harrism  - agreed that the notation (which was just copying from a torch docs reference, not useful) is too underspecified. `p`, `q`, and `u` are ijk integer voxel coordinates (or ijk offsets). `p` means `fine-grid ijk`, `q` means `coarse_grid ijk`, `u` is the zero-based 3-D kernel tap, with `0 <= u[axis] < kernel_size[axis]`. 

In clearer names: 
```
padding_before = floor((kernel_size - 1)/2) 
fine_ijk = stride_ijk * coarse_ijk + tap_ijk - padding_before
```

Everything is componentwise. Untransposed convolution follows these edges fine-to-coarse; transposed convolution follows them coarse-to-fine. I've updated the PR wording and docs. 

In the expression, for a given coarse_ijk, and iterating tap_ijk over the full box of the kernel_size, we visit the same number of fine_ijk values as are in the kernel - these are multiplied by the corresponding kernel weight, then added for the reduction.

blackencino:
Here's a before and after image showing the reproduced bug from the OP, and our fix afterwards. 
issue_668_staircase_before_after

blackencino:
I tried several different shapes, just to see. 
issue_668_procedural_candidates

blackencino:
Here's a similar analysis using the dragon mesh, but the visualization method isn't as good. 
dragon_issue_668_legacy_vs_fixed

src/fvdb/detail/ops/BuildGridForConvTranspose.cu
```diff
@@ -277,9 +277,30 @@ dispatchBuildGridForConvTranspose<torch::kCUDA>(const GridBatchData &baseGridHdl
...
     }
 
-    // Fallback: general coordinate-list path.
-    JaggedTensor coords = convTransposeIJKForGrid(baseGridHdl, kernelSize, stride);
-    return ops::_createNanoGridFromIJK(coords);
+    // Coordinate fallback: preserve exact phase and let the CUDA allocator decide whether the
+    // checked request is serviceable. Only this path advertises coordinate-staging context.
+    const uint64_t stagingBytes = checkTransposeInputAndKernel(baseGridHdl.totalVoxels(), geometry);
+    try {
+        if (isUnshiftedSubdivision(geometry)) {
```
harrism: The comment above says "Only this path advertises coordinate-staging context," but the `isUnshiftedSubdivision` early-return inside the `try` doesn't stage coordinates — `fineGridHandleFromCoarseCUDA` builds on leaf masks. If it OOMs, the handler reports `stagingBytes` for an allocation that never happened.

This is reachable: `supportsLeafMaskSubdivision` (checked earlier) additionally requires `isUniformKernel`, so a non-uniform `K=S` with every axis in `{1,2}` — e.g. `K=S=(1,2,2)` — falls through to here with `paddingBefore == 0` and takes this branch. The stride-1 and stride-2 fast paths above are also uniform-kernel-gated, so nothing catches it first.

Suggest hoisting the subdivision branch above the staging check and `try`, so the staging figure only guards the path that actually stages:

```cpp
if (isUnshiftedSubdivision(geometry)) {
    return fineGridHandleFromCoarseCUDA(baseGridHdl, geometry.stride(), std::nullopt);
}
const uint64_t stagingBytes = checkTransposeInputAndKernel(baseGridHdl.totalVoxels(), geometry);
try {
    return ops::_createNanoGridFromIJK(convTransposeIJKForGrid(baseGridHdl, geometry));
} catch (const c10::OutOfMemoryError &error) {
    ...
}
```

Diagnostic-only, so low severity — the returned topology is correct either way.

fvdb/convolution_plan.py
```diff
@@ -25,6 +27,323 @@
 
 _ANY_CHANNEL_PAIRS: tuple[tuple[int, int], ...] = ()
 
+_TRANSFORM_COMPATIBILITY_ATOL = 1.0e-6
+_TRANSFORM_COMPATIBILITY_RTOL = 1.0e-6
+_WARNED_INCOMPLETE_COVERAGE_GEOMETRIES: set[tuple[tuple[int, int, int], tuple[int, int, int]]] = set()
```
harrism, line 33: This warn-once set accumulates for the lifetime of the process and is never cleared, which makes the warning behavior order-dependent across a test session.

`test_incomplete_residue_warning_is_proactive_and_acknowledgeable` asserts `pytest.warns(...)` for `kernel_size=(1,1,1), stride=(6,1,1)`. That passes today only because no other test constructs a plan with that exact geometry first — if one ever does (or if test order changes under `-p no:randomly` / `pytest-xdist` reordering), the warning is suppressed and the test fails with no obvious cause.

Worth adding a reset hook so the state is explicit, e.g. an autouse fixture:

```python
@pytest.fixture(autouse=True)
def _reset_coverage_warning_state():
    convolution_plan._WARNED_INCOMPLETE_COVERAGE_GEOMETRIES.clear()
```

For what it's worth, the test itself is carefully written — the acknowledgment half deliberately uses a *fresh* geometry `(7,1,1)` so that dedup can't be mistaken for the acknowledgment flag working. This comment is just about making that robustness structural rather than incidental.

Separately: because the key is `(kernel_size, stride)` and never expires, a long-running session or notebook only ever sees this warning once per geometry. That's presumably intended, but worth confirming it's the desired UX for an interactive user who hits the same geometry in a new model.

src/fvdb/detail/ops/BuildGridForConv.cu
```diff
@@ -7,27 +7,157 @@
...
+#include <algorithm>
+#include <limits>
+
 namespace fvdb {
 namespace detail {
 namespace ops {
 
+namespace {
+
+thread_local BuildGridForConvResourceStats gLastBuildGridForConvResourceStats;
```
harrism, line 36: `thread_local` "last call wins" global state. This is safe today — Python calls into the builder on the GIL-holding thread and `lastBuildGridForConvResourceStats()` is queried immediately afterward — but the correctness constraint is invisible at the call site.

The failure mode if that assumption ever breaks (grid building moved to a worker thread, an async/stream-based build, or a future C++ caller that batches builds before querying) is a silent wrong answer rather than an error, and the diagnostics it feeds are exactly the ones tests assert on.

A short comment stating the contract — "set by the last `dispatchBuildGridForConv` call on this thread; must be read on the same thread before any subsequent build" — would be enough. Alternatively, returning the stats alongside the handle from `buildGridForConv` would make the coupling explicit, though I realize that's a wider signature change than this PR needs.

Minor note: every forward path sets this (leaf-mask, unshifted direct, morphology, and both fallbacks), so there's no stale-read path today; the transpose builder correctly doesn't use it at all.

harrism (commented):
Strong PR — this is a genuine root-cause fix rather than a patch over symptoms: the old builders used C `%`/`/` on voxel coordinates, which truncate toward zero and were therefore wrong for any grid with negative coordinates, and the even-kernel phase was re-derived ad hoc in four places. Consolidating onto one canonical relation is the right call.

I verified the pieces most likely to hide a silent bug: the aggregate initializer in `reverseGatherScatterDefaultTopology` correctly cross-assigns `gather↔scatter` and `featureTotalVoxels↔outputTotalVoxels` against the struct declaration order; the direct `K=S` projection correctly reduces to `floorDiv(fine + paddingBefore, stride)`; and the stride-1 morphology pad passes are the correct mirror image between forward (`[-paddingAfter, +paddingBefore]`) and transpose (`[-paddingBefore, +paddingAfter]`). The executors' `direction` checks mean a reversed topology can't silently run through the forward kernel.

Test quality is good — the oracle is genuinely independent, and the float64 adjoint dot-product identity test at 1e-12 is the right way to prove the exact-transpose claim. I also confirmed no existing test assertions were weakened; the legacy suites only gained clarifying docstrings.

Three minor comments inline. Two questions:

1. The PR description still carries the underspecified `p = stride * q + u - floor(...)` notation. Your thread reply and `CHANGES.md` both use the clear names — could you fold that wording into the description, since that's what lands in the merge commit?
2. `fvdb/nn/simple_unet.py:431` chains `coarsened_grid(2).conv_grid(...)`, which `CHANGES.md` now warns against as a convolution target. As far as I can tell it's fine, because every plan there uses `stride=1` with matching transforms and the resolution changes go through `max_pool`/`refine` rather than strided convolution — so neither the voxel-size scaling nor the new transform validation bites. Can you confirm that's deliberate rather than incidental?

The main non-code judgment call is shipping a breaking topology + world-transform change with no legacy mode. Given the old behavior was actually wrong, I think that's defensible — but flagging it as a product decision worth making explicitly.

docs/api/convolution_plan.rst
swahtz, line 1: **🎵 note (non-blocking):** ‏ We don't usually put comments on classes in these rst files, could this just be rolled into the information in the class docstring?

fvdb/convolution_plan.py
```diff
@@ -25,6 +27,323 @@
...
+            f"Convolution currently supports only a=0. {migration_hint}"
+        )
+    if not compatibility.compatible:
+        raise RuntimeError("Convolution transform compatibility validation reached an inconsistent state")
+
+
+def _resolve_topology_policy(target_grid: GridBatch | None, topology_policy: str | None) -> str:
+    if topology_policy is None:
+        return "full_support" if target_grid is None else "restricted"
+    if topology_policy not in ("full_support", "restricted"):
```
swahtz: **💡 suggestion:** ‏I think using the term 'full support' is going to be confusing and I'd advocate we think about changing this term.  I know it means 'full structural support' but without the 'structural' qualifier, I think it could get confused for how we use 'support' in other parts of the codebase.  Imagine phrases like 'supported kernels sizes 3, 5, 7' or 'K=4 has full support' could collide in multiple meanings.

I propose the word 'complete' as a replacement.  'Complete' is already used in the vocabulary in this PR (e.g. 'complete structural support') and I think it pairs well with 'restricted'.

fvdb/convolution_plan.py
```diff
@@ -213,6 +621,9 @@ def from_grid_batch(
         *,
         expert_config: dict[str, Any] = _DEFAULT_CONFIG,
         channel_pairs: tuple[tuple[int, int], ...] = _ANY_CHANNEL_PAIRS,
+        topology_policy: str | None = None,
```
swahtz: **❓ question:** ‏Could we declare enums for `topology_policy` to make it clear what the values of these policies should be and reduce user error (similar to `SmoothingMode` elsewhere in fvdb-core)?  `topology_provenance` could use this too and `phase_policy` might also be helpful in case this gains more than 1 policy.

tests/unit/test_conv_semantics.py
```diff
@@ -0,0 +1,238 @@
...
+    dense_transpose_oracle,
+    floor_div,
+    forward_degrees,
+    forward_support,
+    relation_edges,
+    transpose_support,
+)
+
+
+@pytest.mark.parametrize("kernel", range(1, 7))
```
swahtz: **🧹 chore:** ‏The four existing `test_conv*` files contain `unittest.TestCase` classes whereas the two new files here contain bare pytest functions.  Let's wrap these into `unittest.TestCase` classes like the others for consistency

CHANGES.md
swahtz, line 1: **👏 praise:** ‏Thanks for updating the CHANGES!

fvdb/convolution_plan.py
```diff
@@ -25,6 +27,323 @@
...
+_TRANSFORM_COMPATIBILITY_RTOL = 1.0e-6
+_WARNED_INCOMPLETE_COVERAGE_GEOMETRIES: set[tuple[tuple[int, int, int], tuple[int, int, int]]] = set()
+_DENSE_BACKEND_DISABLED_MESSAGE = (
+    "The dense convolution backend is disabled because it does not yet implement the canonical sparse geometry "
+    "and public transposed-weight layout; use backend='default' or backend='gather_scatter'."
+)
+
+
+@dataclass(frozen=True)
+class ConvolutionCoverageReport:
```
swahtz, line 50: **☑️ todo:** ‏ConvolutionCoverageReport and ConvolutionTransformCompatibility below are returned by public properties but aren't exported from `fvdb/__init__.py`, that should be fixed and they should be autoclass-documented in `convolution_plan.rst`

src/python/GridBatchOps.cpp
```diff
@@ -759,6 +759,20 @@ bind_grid_batch_ops(py::module &m) {
         py::arg("kernel_size"),
         py::arg("stride"));
 
+    m.def("last_conv_grid_resource_stats", []() {
+        auto const stats = ops::lastBuildGridForConvResourceStats();
+        py::dict result;
```
swahtz: **💡 suggestion:** ‏We don't have any other functions that return a `py::dict` in the bindings, could we create a small bound struct to contain the stats to match the rest of the codebase.  It would also give the `.pyi` a real type instead of `dict[str, int | bool].`

fvdb/convolution_plan.py
```diff
@@ -25,6 +27,323 @@
...
+        return
+    geometry_key = (kernel_size, stride)
+    if geometry_key in _WARNED_INCOMPLETE_COVERAGE_GEOMETRIES:
+        return
+    _WARNED_INCOMPLETE_COVERAGE_GEOMETRIES.add(geometry_key)
+    warnings.warn(
+        "This convolution geometry leaves uncovered stride residues on axes "
+        f"{uncovered_axes}; some active fine coordinates can have zero rulebook degree. "
+        "This matches dense Torch sampling. Pass acknowledge_incomplete_coverage=True to suppress this warning.",
+        UserWarning,
```
swahtz: **☑️ todo:** ‏We should use a named, exported warning category rather than a generic `UserWarning`.  With this the only way a user can selectively filter for this warning is regex-matching the message text. A `class ConvolutionCoverageWarning(UserWarning)` exported from fvdb lets users (and CI) do `filterwarnings("error", category=fvdb.ConvolutionCoverageWarning)`. This is the established pattern — sklearn.exceptions.ConvergenceWarning, pandas.errors.PerformanceWarning, numpy's named warnings — and we have no custom categories yet, so this would set the precedent for the right direction.

src/fvdb/detail/ops/BuildGridForConv.cu
```diff
@@ -53,178 +181,246 @@ buildCoarseGridFromFineGridCPU(const GridBatchData &fineBatchHdl,
...
-                               stride,
-                               kernelVolume,
-                               outIJKAcc,
-                               outIJKBIdxAcc,
-                               outMaskAcc);
+JaggedTensor
+countThenFillConvIJKForGrid(const GridBatchData &batchHdl, ConvolutionGeometry const &geometry) {
+    const int64_t inputVoxelCount = batchHdl.totalVoxels();
+    const auto countOptions = torch::TensorOptions().dtype(torch::kInt32).device(batchHdl.device());
+    torch::Tensor counts    = torch::zeros({inputVoxelCount}, countOptions);
```
swahtz: **🤓 nitpick (non-blocking):** Technically this could just be `torch::empty` because each active voxel will write its slot.

swahtz (changes requested):
Thanks for doing this, it's a lot of work and it's great to have this fix!

Perhaps one other general thing we could add somewhere in the docs or a docstring is just a statement to clarify for users what equivalences can be drawn between fvdb-core and the 3 other sparse conv frameworks to save users time trying to discover it if they're porting work.  I drafted something with Claude if it helps:

fVDB's geometry equals spconv/torchsparse with `padding = (K−1)//2;` MinkowskiEngine corner-anchors even kernels `[0,K)`, so even-K topologies ported from ME are shifted by `floor((K−1)/2)` and fVDB additionally materializes the uncropped boundary support (5³ vs 4³ for 16³/K=S=4)

fvdb/convolution_plan.py
blackencino: I made this change, but disagree with your reasoning. "support" is a very established term of art with respect to kernels and convolution.

src/fvdb/detail/ops/BuildGridForConvTranspose.cu
blackencino: Hoisted unshifted subdivision before the staging check and added coverage for non-uniform `K=S(1,2,2)`.

fvdb/convolution_plan.py
blackencino: Add automatic test-state reset. Once-per-geometry, process-lifetime suppression remains intentional and is now documented.

src/fvdb/detail/ops/BuildGridForConv.cu
blackencino: Added the same-thread, read-before-next-build contract beside the `thread_local` state.

docs/api/convolution_plan.rst
blackencino: Moved the explanatory material into the `ConvolutionPlan` class docstring.

fvdb/convolution_plan.py
blackencino: Replaced the public policy name with `complete`

fvdb/convolution_plan.py
blackencino: Added and exported enums for topology polic, topology provenance, and phase policy.

tests/unit/test_conv_semantics.py
blackencino: Converted both new convolution-semantics files to `unittest.TestCase` classes

fvdb/convolution_plan.py
blackencino: Exported `ConvolutionCoverageReport` and `ConvolutionTransformCompatibility` and added their API documentation.

CHANGES.md
blackencino: Thanks

src/python/GridBatchOps.cpp
blackencino: Replaced the `py::dict` result with a bound, read-only `BuildGridForConvResourceStats` type and updated the stub.

fvdb/convolution_plan.py
blackencino: Added and exported `ConvolutionCoverageWarning` , with tests using the named category.

src/fvdb/detail/ops/BuildGridForConv.cu
blackencino: Changed the count tensor to `torch::empty`, every active voxel writes its slot before the tensor is read.

swahtz (approved):
Thanks for addressing the notes, lgtm!

PR #729, opened 2026-08-05
Add mGPU 3DGUT support

matthewdcong:
This PR adds mGPU support for Unscented Transform projection and world-space rasterization ops. Together, these enable support for training splats with distorted camera models following the 3DGUT paper.

In addition to implementing the mGPU versions of the ops, this PR also incorporates the shared memory optimizations in #555 for the UT projection. This is necessary in order to maintain the preferred sharding for mGPU but also reduces shared memory usage for the single-GPU case as well.

Adding more test coverage for mGPU in CI is an ongoing effort partially due to limitations in PyTorch. However, I have verified manually that this change passes the fvdb-reality-capture Python Gaussian splatting tests for the mGPU case.


matthewdcong:
Ongoing work on adding mGPU test coverage, e.g. https://github.com/openvdb/fvdb-reality-capture/pull/311

src/fvdb/detail/utils/gsplat/GaussianCameras.cuh
```diff
@@ -727,55 +727,73 @@ template <typename T> struct PerspectiveWithDistortionCamera {
...
     DistortionModel cameraModel           = DistortionModel::PINHOLE;
+    int64_t sharedCameraOffset            = 0;
 
-    Mat3 *__restrict__ worldToCamStartRotShared   = nullptr; // [C,3,3], optional
-    Vec3 *__restrict__ worldToCamStartTransShared = nullptr; // [C,3], optional
-    Mat3 *__restrict__ worldToCamEndRotShared     = nullptr; // [C,3,3], optional
-    Vec3 *__restrict__ worldToCamEndTransShared   = nullptr; // [C,3], optional
-    Mat3 *__restrict__ projectionMatsShared       = nullptr; // [C,3,3], optional
-    T *__restrict__ distortionShared              = nullptr; // [C,K], optional
+    Mat3 *__restrict__ worldToCamStartRotShared   = nullptr;
```
swahtz, line 732: **💅 polish:** ‏Let's restore the comments that indicate the shapes for ease-of-reading.

src/fvdb/detail/utils/gsplat/GaussianCameras.cuh
```diff
@@ -1131,56 +1165,86 @@ template <typename T> struct OrthographicWithDistortionCamera {
...
+    Vec3 *__restrict__ worldToCamStartTransShared = nullptr;
+    Mat3 *__restrict__ worldToCamEndRotShared     = nullptr;
+    Vec3 *__restrict__ worldToCamEndTransShared   = nullptr;
+    Mat3 *__restrict__ projectionMatsShared       = nullptr;
 
-  public:
-    /// @brief Returns dynamic shared-memory bytes required to cache camera state.
     inline __host__ __device__ size_t
-    numSharedMemBytes() const {
+    sharedMemoryBytesForCameraCount(const uint32_t cameraCount) const {
```
swahtz, line 1177: **🤓 nitpick (non-blocking):** ‏This could be a static function.

src/fvdb/detail/ops/gsplat/RasterizeWorldSpaceGaussiansBackward.cu
```diff
@@ -458,10 +464,308 @@ launchBackward(const torch::Tensor &means,
...
+                                                   0,
+                                                   blockCount,
+                                                   stream);
+    }
+    return {dMeans, dQuats, dLogScales, dFeatures, dOpacities};
+}
+
+template <typename ScalarType>
+void
+reduceGradientShards(std::vector<torch::Tensor> &localGradients, torch::Tensor &outputGradient) {
```
swahtz, line 545: **💅 polish:** ‏This looks like the same function as in `RasterizeScreenSpaceGaussiansBackward.cu`, I'd suggest putting this in a utility/header to DRY these up.

swahtz (approved):
A couple of non-blocking comments but otherwise looks good.  I think one other thing this PR should have would be some unit tests in `GaussianProjectionUTTest.cpp` and `GaussianRasterizeWorldSpaceTest.cpp` now that this functionality works for multi-GPU.

src/fvdb/detail/utils/gsplat/GaussianCameras.cuh
```diff
@@ -727,55 +727,73 @@ template <typename T> struct PerspectiveWithDistortionCamera {
...
+    T *__restrict__ distortionShared              = nullptr;
 
-  public:
-    /// @brief Returns dynamic shared-memory bytes required to cache camera state.
     inline __host__ __device__ size_t
-    numSharedMemBytes() const {
+    sharedMemoryBytesForCameraCount(const uint32_t cameraCount) const {
         size_t bytes = 0;
         bytes        = alignUpBytes(bytes, alignof(Mat3));
-        bytes += static_cast<size_t>(numCameras) * sizeof(Mat3); // worldToCamStartRotShared
```
swahtz, line 744: **💅 polish:** ‏Restore these  comments so we can follow the shared memory calculation

src/fvdb/detail/utils/gsplat/GaussianCameras.cuh
```diff
@@ -1131,56 +1165,86 @@ template <typename T> struct OrthographicWithDistortionCamera {
...
+        bytes += static_cast<size_t>(cameraCount) * sizeof(Mat3);
```
harrism, line 1180: **💡 suggestion:** ‏ Why not just use multiplication by 5 instead of 5 additions?  Also, I think commenting on what the storage is for like the 5 comments that you deleted would be helpful.

```suggestion
        bytes += 5 * static_cast<size_t>(cameraCount) * sizeof(Mat3);
```

src/fvdb/detail/ops/gsplat/RasterizeWorldSpaceGaussiansBackward.cu
```diff
@@ -346,72 +354,69 @@ rasterizeFromWorld3DGSBackwardKernel(
                                               gid * args.dFeatures.stride(1);
 #pragma unroll
                 for (uint32_t k = 0; k < NUM_CHANNELS; ++k) {
-                    atomicAdd_system(dFeaturesGaussianPtr + k * args.dFeatures.stride(2),
-                                     v_feat_local[k]);
+                    atomicAdd(dFeaturesGaussianPtr + k * args.dFeatures.stride(2), v_feat_local[k]);
```
harrism, line 357: Curious why we these were atomicAdd_system before.  This wasn't ever pinned host memory, right?

harrism (approved):
Looks good. Just one minor suggestion.

src/fvdb/detail/ops/gsplat/RasterizeWorldSpaceGaussiansBackward.cu
matthewdcong: Likely carryover from the screen space rasterization code where we needed `atomicAdd_system` previously because we had multiple GPUs scattering gradients to different Gaussians. It's no longer needed now because we scatter to a local buffer of gradients and then reduce across devices via NCCL.

src/fvdb/detail/utils/gsplat/GaussianCameras.cuh
matthewdcong: I'll restore the comments here but we can't just multiply by five because some of the datatypes are `Vec3` and some are `Mat3` and we need to make sure each is aligned.

matthewdcong:
> A couple of non-blocking comments but otherwise looks good. I think one other thing this PR should have would be some unit tests in `GaussianProjectionUTTest.cpp` and `GaussianRasterizeWorldSpaceTest.cpp` now that this functionality works for multi-GPU.

Yeah, multi-GPU tests are on the TODO list but there are a couple of limitations that prevent it them from being included in this PR. I've done some manual verification locally as I mentioned above.

Issue #414, opened 2026-01-14
PyTorch 2.7.0 and IsaacSim 5.1 Compatibility


darosaj:
I'm wondering if it would be possible to release a fvdb wheel built with the same python version (3.11) and PyTorch version (2.7.0) that IsaacSim5.1 requires, so that both could be pip installed in the same environment.

Thanks!

swahtz:
Hi @darosaj , I've been thinking a bit about this along with some other requests we've had for combinations using versions of PyTorch older than the current stable that we distribute.  Based on that, I've recently merged a script that makes it easy for people to build custom wheels of `fvdb-core`, using the same process we use in release, with whatever versions of PyTorch/Python/CUDA they like (assuming a PyTorch wheel exists for that combination).

https://fvdb-core.readthedocs.io/latest/installation.html#build-and-install-a-custom-wheel-from-source

Have a look, that should cover your use-case.

PR #731, opened 2026-08-06
Add Dockerized production wheel build script and unify publish workflows

swahtz:
## Description

Motivated by issues such as #414 and other related requests for built wheels of `fvdb-core` with alternative dependencies to what is currently published, this PR introduces mechanisms to make it much easier for users to generate these packages on their own and clarifies in the documentation what versions `fvdb-core` will build and test. 

Adds a user-facing script that builds a production `fvdb-core` wheel inside a Docker container from the local checkout (uncommitted changes included), and refactors the publish workflows to use it so the wheel recipe lives in exactly one place.

Now a user who needs a one-off installable wheel for a specific Python/Torch/CUDA/Compute Capability, can run:

`./docker/build_wheel.py --python 3.11 --torch 2.11.0 --cuda 13.0 --cuda-arch-list native`

To build a Docker container with those dependent softwares and automate a build ('native' tries to detect the current machine's compute arch but an arch string is also supported).

Documentation has been refactored such that the `README` only contains setup/build instructions related to development process and the published documentation contains all end-user installation instructions.

### New: `docker/build_wheel.py` + `docker/Dockerfile.wheel`

- Reproduces the official publish build: `nvidia/cuda:*-cudnn-devel-rockylinux8`, pinned `uv`-managed Python, gcc-toolset, Kitware CMake, torch pinned via constraints file against the matching `download.pytorch.org/whl/cuXXX` index, `./build.sh wheel strip_symbols verbose`, auditwheel manylinux repair with the exclude lists from `.github/versions.json`, and the `+pt<torch>.cu<cuda>` local version suffix.
- Written in stdlib-only Python (the host needs only Docker with BuildKit and `python3`).
- Defaults for `--python`, `--torch`, `--cuda`, and `--cuda-arch-list` come from `.github/versions.json`, but any Python/CUDA version may be requested: unlisted CUDA versions are used verbatim as the `nvidia/cuda` image tag (full `major.minor.patch` form); `--cuda-arch-list native` detects the host GPU's compute capability via `nvidia-smi` (no GPU needed otherwise).
- `--version-mode suffix|nightly|none`, `--version <exact>`, and `--skip-auditwheel` cover all CI version-stamping paths; `--jobs` and `--output-dir` for local control.
- The version stamp is applied to the image's copy of `pyproject.toml` only, and BuildKit's local exporter writes the wheel to `./dist/` owned by the invoking user — the working tree is never modified.

### Unified CI

`publish.yml` (`fvdb-build`) and `nightly-publish.yml` (`nightly-build`) no longer run as container jobs with duplicated dnf/uv/cmake/torch/sed/build/auditwheel steps — each is now a single `./docker/build_wheel.py` invocation (net −246/+80 lines). Existing behavior is preserved: S3 publishes get the local version suffix, TestPyPI re-runs get `.postN`, nightlies get `.devYYYYMMDD+pt.cu`, and auditwheel is skipped for non-release S3 publishes. The `auditwheel_excludes` arrays in `versions.json` (previously parsed but unused) are now the actual source of truth.

### Docs

New "Building a custom wheel with Docker" subsection in `docs/installation.rst`, and the README's Docker section now points wheel builders at `docker/build_wheel.py` instead of the old `docker ps`/`docker cp` extraction instructions (its dev-container instructions are unchanged).

## Testing

- CLI paths exercised: bad version-mode/jobs, mutually exclusive version flags, native arch detection, and the unlisted-CUDA fallback (e.g. `--cuda 12.6.3` resolves to the verbatim `nvidia/cuda` image tag and `cu126` torch index).
- Full end-to-end local builds (`--cuda-arch-list "12.0+PTX"`, run with both the initial bash version and the final Python script) produced `fvdb_core-0.6.0.dev0+pt211.cu130-cp312-cp312-manylinux_2_27_x86_64.whl` (26 MB, no bundled torch/CUDA libs), with a clean `git status` afterward.
- Wheel installed in a fresh `python:3.12-bookworm` container alongside torch 2.11.0+cu130: `import fvdb` works and a GPU grid construction succeeds.
- Both workflow files pass YAML validation.

## Note for reviewers

The build jobs now run directly on the EC2 runners (not in a job container), so the runner AMI needs git and Docker with BuildKit. Docker is already required for container jobs; a `workflow_dispatch` of publish with target `testpypi` is a good first real-world test.  A run with these arguments can be found here:  https://github.com/openvdb/fvdb-core/actions/runs/31128771936


matthewdcong:
Thanks for putting this together. What happens if you request a PyTorch and CUDA combination that does not have a prebuilt PyTorch wheel? e.g. PyTorch v2.13 and CUDA 13.1? I might have missed it, but if it's not in the docs we should add a section and an error message in the wheel build script.

swahtz:
Good catch — as of ffec05fd the script now checks this up front instead of failing minutes into the docker build with a buried resolver error.

Before starting the build, `build_wheel.py` consults the same PEP 503 index the build resolves against (`https://download.pytorch.org/whl/<tag>/torch/`), which makes it the canonical source of valid combinations with no matrix to maintain. Concretely:

- Unknown CUDA variant (your `cu131` example): `Error: cu131 is not a published PyTorch CUDA variant (no index at https://download.pytorch.org/whl/cu131/torch/). See https://pytorch.org/get-started/locally/ for available variants.`
- Known variant, unpublished torch version: `Error: PyTorch does not publish a torch==2.13.0 wheel for cu128; available versions for cu128: 2.7.0, 2.7.1, 2.8.0, 2.9.0, 2.9.1, 2.10.0, 2.11.0`
- The same fetch also validates the Python version against the wheels' cp tags: `Error: PyTorch does not publish a torch==2.11.0+cu130 wheel for Python 3.9; available Python versions: 3.10, 3.11, 3.12, 3.13, 3.14`

If the index is unreachable (e.g. building offline), the check degrades to a warning and an invalid combination fails during the build's dependency-install step as before. There's also a new paragraph in `docs/installation.rst` documenting this behavior.

harrism (approved):
This looks great. Excellent to factor this out so developers / users can build what they need using the same script we use in CI.

We should duplicate for reality-capture.

swahtz:
> This looks great. Excellent to factor this out so developers / users can build what they need using the same script we use in CI.
> 
> We should duplicate for reality-capture.

I would have thought `pip install` is all that's needed in reality-capture because there's no build (and so no custom dependencies to have to build for)?

I also opened #736 to see if we could DRY it up further with our pip-based tests but that requires also building the gtests…

PR #740, opened 2026-08-12
Add compatibility wrapper for StrEnum which was introduced in py310

matthewdcong:
`StrEnum` was introduced in Python 3.10 and we still support Python 3.10 (the Python version that ships with Ubuntu 22.04).


swahtz (approved):
lgtm, thanks!

PR #744, opened 2026-08-19
Update to tinyply 3.0

matthewdcong:
tinyply 3.0 significantly improves load times for PLY files. A 375M Gaussian PLY with SH3 loads in 4 minutes instead of 12, approximately a 3x speedup.

This also removes the need for the CMake minimum version patch. The only code changes necessary are to address a leaking namespace in tinyply 2.4 that was fixed in 3.0.


areidmeyer (approved):
Simple enough. Approved.

PR #738, opened 2026-08-11
PyTorch 2.13

swahtz:
This pull request updates fVDB&#39;s build, test, and documentation systems to support PyTorch 2.13, CUDA 13.0/13.2, and Python 3.10–3.15. It removes building/testing for CUDA 12.8 and PyTorch 2.11, and updates installation instructions and version matrices accordingly. The changes ensure that all CI workflows, publishing scripts, and documentation are aligned with the new supported versions.

**Build and Test Matrix Updates:**

- Updated `.github/versions.json` to add CUDA 13.2 (patch 13.2.1), remove CUDA 12.8, and set supported Python versions to 3.10–3.15 and PyTorch version to 2.13.
- Updated the publish matrix and version outputs to handle CUDA 13.2 and removed references to CUDA 12.8 throughout CI configuration files.

**CI Workflow Changes:**

- Renamed and refactored the main CUDA workflow file from `cu128.yml` to `cu132.yml`, updating all job names, container images, and pip install commands to use CUDA 13.2 and the new PyTorch version.

**Documentation Updates:**

- Updated the version matrix and installation instructions in `docs/installation.rst` to reflect support for PyTorch 2.13, CUDA 13.0/13.2, and Python 3.10–3.15, and to remove references to older versions.
- Added a new entry in `CHANGES.md` to document the updated support for PyTorch 2.13, CUDA 13.0/13.2, and Python 3.10–3.15.

**Decoupling stable-release install docs from in-development versions:**

Previously the pip install examples in `docs/installation.rst` were assembled from `fvdb_core_stable_version` (release cadence) plus the torch/CUDA tags in `.github/versions.json` (main's dev cadence) — so moving main to PyTorch 2.13 would have made the stable install commands point at wheels that don't exist (e.g. `0.5.1+pt213.*`). This PR decouples the two so main can target the next release without breaking the published-wheel instructions:

- `docs/conf.py` grows a stable release metadata block (fvdb version, PyTorch version, CUDA list, Python range) that drives the "Installation from pre-built wheels" examples; `.github/versions.json` keeps driving the nightly examples.
- The version matrix's 0.6 row is marked "(pre-release, nightly wheels only)" and auto-fills its PyTorch/Python/CUDA cells from `versions.json`.
- `devtools/update-doc-versions.sh` gains a `--versions-json` option that snapshots the released torch/cuda/python into the stable block and strips the pre-release marker for the released minor version.
[...]

PR #745, opened 2026-08-20
Weekly compatibility tests against oldest supported PyTorch

swahtz:
Implements the weekly testing regime described in the installation docs ([Notes on Testing, Compatibility, and Distribution](https://fvdb-core.readthedocs.io/latest/installation.html#notes-on-testing-compatibility-and-distribution)): test fvdb-core weekly against the oldest PyTorch version that publishes wheels for the lowest CUDA version implied by the "stable PyTorch's latest two CUDA versions" policy. For the PyTorch 2.13 matrix (#738) that is **PyTorch 2.9.1 with CUDA 13.0** (verified against the versions published on download.pytorch.org/whl/cu130).

**Changes:**

- **`.github/versions.json`**: new `weekly_compat` block recording the tested torch/cuda versions, so they get reviewed and updated alongside the rest of the version matrix each cycle.
- **`.github/workflows/load-versions.yml`**: new outputs `compat-torch-version`, `compat-torch-full-version`, `compat-cuda-version`, `compat-cuda-tag`, and `compat-cuda-patch` (the container-image patch is looked up from the existing `cuda.versions` map, so the compat CUDA must be one of the supported versions — which the policy guarantees).
- **`.github/workflows/weekly-compat.yml`**: new scheduled workflow (Saturdays 8am UTC + `workflow_dispatch`) mirroring the `cu130.yml` pipeline without the `pull_request_target` merge plumbing: provision EC2 CPU runner → build wheel/gtests against the compat PyTorch/CUDA pin → provision EC2 GPU runner → run gtests and Python unit tests → tear down runners.
- **`CHANGES.md`**: note the new weekly compatibility testing under 0.6.0.

**Notes:**

- Doc tests (`pytest --markdown-docs`) are intentionally omitted: the documentation examples pin the current in-development versions, not the compat versions.
- Passes `actionlint`, `zizmor --persona=regular`, and the repo's EC2 runner-token policy check locally.
- The `weekly_compat` values are a deliberate second source of truth from `torch.version`: at each new dev cycle (like #738), whoever bumps the matrix consults [PyTorch's release compatibility matrix](https://github.com/pytorch/pytorch/blob/main/RELEASE.md#release-compatibility-matrix) and updates `weekly_compat` per the policy.


.github/workflows/weekly-compat.yml
```diff
@@ -0,0 +1,417 @@
...
+
+jobs:
+  versions:
+    if: github.repository == 'openvdb/fvdb-core'
+    uses: ./.github/workflows/load-versions.yml
+
+  ##############################################################################
+  # START FVDB BUILD RUNNER
+  ##############################################################################
+  start-build-runner:
```
harrism, line 35: > _Analysis by Claude Code, run by @harrism. Findings marked "verified" were checked against the code and against current `main`._

**This job (and the three like it below) predates the shared runner workflows and should be rebased onto them.** *(verified against current `main`)*

`main` now has `./.github/workflows/start-ec2-runner.yml` and `stop-ec2-runner.yml` (#759, #760, #763). Every other workflow in the repo calls them. They provide, in one place:

* a sweep across all three us-east-2 AZs;
* three attempts with 90s/180s backoff;
* a `g5.xlarge` fallback when `g6.xlarge` is exhausted region-wide;
* termination of instances left behind by a failed attempt;
* outcome-based selection of the winning attempt.

This job has none of it, so one `InsufficientInstanceCapacity` refusal fails the entire weekly run with no retry. Measured over 2026-08-27..09-03, that condition occurred in 14 of 54 GPU runner starts, six of them with no capacity available anywhere in the region for minutes at a time (see #761).

The replacement is roughly:

```yaml
  start-build-runner:
    name: Start CPU-only EC2 runner for build
    needs: versions
    permissions:
      id-token: write  # Required for AWS OIDC
    uses: ./.github/workflows/start-ec2-runner.yml
    with:
      instance-type: m6a.8xlarge
      az-config: ${{ needs.versions.outputs.aws-cpu-az-config }}
      aws-role: ${{ needs.versions.outputs.aws-role }}
      aws-region: ${{ needs.versions.outputs.aws-region }}
    secrets:
      EC2_RUNNER_TOKEN: ${{ secrets.EC2_RUNNER_TOKEN }}
```

and the same shape for the GPU start at :191 and the two stop jobs at :178 and :409. Note the token must be forwarded by name — `secrets: inherit` is rejected by `check_runner_token_policy.py` since #759.

Worth doing before merge rather than after: the weekly job runs unattended, so it is the one place where a capacity failure is least likely to be noticed and retried by hand.

.github/versions.json
```diff
@@ -17,6 +17,14 @@
     "version": "2.13",
     "full_version": "2.13.0"
   },
+  "weekly_compat": {
```
harrism, line 20: > _Analysis by Claude Code, run by @harrism. Findings marked "verified" were checked against the code and against current `main`._

**These pins are described as derived but are hand-typed, and nothing checks the derivation.** *(verified)*

The comment on the next line defines the contract precisely — *"oldest PyTorch that publishes wheels for the lowest CUDA version in `cuda.versions`"* — but `torch 2.9.1` and `cuda 13.0` are literals. Nothing recomputes them from `.cuda.versions` / `.torch.version`, and `check_env_pin_consistency.py` does not cover this: that script only compares the conda `env/*.yml` files against each other and never reads `versions.json`.

Failure mode is the quiet one. When the CUDA floor rises to 13.2, `cuda.versions` changes, these two values do not, and the weekly job keeps happily building the obsolete 13.0 / 2.9 combination and reporting green. The workflow that exists to detect compatibility drift becomes the thing that has drifted — and because it is green, nobody looks.

The same numbers are also restated in prose in `CHANGES.md` and `docs/installation.rst:233`, so there are three copies to keep in step.

Two ways out, either fine:

* derive them at runtime in the `versions` job — `jq` can pick the lowest key of `.cuda.versions` and map it to the oldest torch that ships wheels for it; or
* keep the literals and add a check that recomputes and compares, in the spirit of `check_env_pin_consistency.py`.

The second is probably less work and fails loudly, which is the property that matters here.

.github/workflows/weekly-compat.yml
```diff
@@ -0,0 +1,417 @@
...
+# ("Notes on Testing, Compatibility, and Distribution"): build and test main
+# against the oldest PyTorch version that publishes wheels for the lowest
+# CUDA version our stable-PyTorch policy implies. The tested versions are
+# recorded in .github/versions.json under "weekly_compat".
+
+name: fVDB Weekly Compatibility Build and Test (oldest tested PyTorch)
+
+on:
+  schedule:
+    - cron: "0 8 * * 6"  # weekly, Saturday 8am UTC
```
harrism, line 14: > _Analysis by Claude Code, run by @harrism. Findings marked "verified" were checked against the code and against current `main`._

**A scheduled workflow that fails has no route to a human.** *(verified)*

Every other CI workflow in this repo is PR-triggered, so a failure is visible on somebody's PR within minutes. This one runs Saturday 08:00 UTC against nobody's branch. GitHub's default notification for a failed scheduled run goes only to the user who last edited the workflow file — so a breakage sits red until someone happens to open the Actions tab.

That matters more than usual here because the failure this job is designed to catch is a *slow* one: PyTorch 2.9 compatibility will break silently one dependency bump at a time, and the signal is only useful if it reaches somebody.

Cheapest fix is a final `if: failure()` job that posts to the team's Slack webhook or opens an issue. `nightly.yml` in fvdb-reality-capture has a pattern for this worth copying.

.github/workflows/weekly-compat.yml
```diff
@@ -0,0 +1,417 @@
...
+  workflow_dispatch:
+
+concurrency:
+  group: weekly-compat
+  cancel-in-progress: true
```
harrism: > _Analysis by Claude Code, run by @harrism. Findings marked "verified" were checked against the code and against current `main`._

**Speculative:** `cancel-in-progress: true` on a workflow that provisions EC2 instances can leak them.

If someone triggers this via `workflow_dispatch` while the scheduled run is mid-build, the scheduled run is cancelled. Cancellation does not run the `stop-*-runner` jobs unless they are reached — so an `m6a.8xlarge` and possibly a `g6.xlarge` can be left running until someone notices.

I have not confirmed a real occurrence, and `nightly-publish.yml` carries the same shape, so this may be an accepted repo-wide risk rather than something to fix here. Flagging because the blast radius is a large instance billing indefinitely rather than a failed run. If it is worth closing, `cancel-in-progress: false` on a once-weekly job costs almost nothing — concurrent runs are already unlikely.

harrism (commented):
> _Analysis by Claude Code, run by @harrism. Findings marked "verified" were checked against the code and against current `main`._

## Review: weekly compatibility tests

The workflow is mechanically sound and security-clean — cron is valid, every third-party action is SHA-pinned, no `pull_request_target`, permissions are minimal, every `needs.versions.outputs.*` reference resolves, and the `torch-2.9.1+cu130-cp312` wheel is live on download.pytorch.org. The runner-token policy checker passes.

Two things worth resolving before this merges, plus a user-facing bug this PR is the natural place to fix.

### 1. The branch predates the reusable-runner work (inline comment below)

This branched from `2c724d10` (#738), so it hand-rolls `machulav/ec2-github-runner` in four places. Since then #759/#760/#763 moved every other workflow onto `./.github/workflows/start-ec2-runner.yml` and `stop-ec2-runner.yml`, which carry the AZ sweep, retry-with-backoff, orphan cleanup and the `g5.xlarge` fallback. Merged as-is this becomes the one workflow with none of that — a single insufficient-capacity refusal fails the whole weekly run. It also re-adds a fifth copy of the token-handling code that #759 centralized. A rebase should be a net deletion here.

### 2. The `weekly_compat` pins are unguarded (inline comment below)

### 3. `pyproject.toml:15` declares `torch>=2.7`, two minors below the oldest version actually tested

Not a file this PR touches, so no inline comment — but this is the PR that codifies "oldest supported PyTorch", which makes it the right place. A user on torch 2.7 or 2.8 gets a satisfied dependency resolve and then an undefined-symbol failure at `import fvdb`. Raising the floor to `>=2.9` makes the declared floor and the tested floor the same number.

### Minor

`cu130.yml` runs an `fvdb-docs-test` job (`pytest --markdown-docs docs`) that has no counterpart here. Deliberate, or an oversight? If the docs examples use APIs whose behaviour differs on 2.9, this is exactly the workflow that would catch it.

The three `astral-sh/setup-uv@d4b2f3b...` pins lack the `# vX.Y.Z` trailing comment every other pinned action in the repo carries. Cosmetic, and inherited from `cu130.yml`.

---

Codex was also run over this diff and reported no findings; everything above came from manual verification against current `main`.

.github/workflows/weekly-compat.yml
swahtz: Done in 4a27132. Rebased onto current `main` and replaced all four hand-rolled runner jobs with `start-ec2-runner.yml` / `stop-ec2-runner.yml`, including the `g5.xlarge` fallback for the GPU runner. The workflow no longer references the runner action or AWS credentials directly.

Related: while looking at this I traced the orphaned-instance problem to cancellation mid-provisioning and opened #768 to fix it in the shared workflows. Once that merges, this workflow will pick up the deterministic `runner-label` convention in a one-line follow-up. It is not exposed in the meantime because it no longer cancels in-progress runs (see below).

.github/versions.json
swahtz: Agreed the "derived" wording was wrong; these are hand-maintained pins. In 4a27132:

- `load-versions.yml` now fails with an `::error` annotation if `weekly_compat.cuda` is not the lowest key of `cuda.versions` (numeric sort, so a future 13.10 orders after 13.2). Since every workflow loads versions, a stale pin breaks every PR immediately rather than letting the weekly job drift green.
- The comment in `versions.json` now says the pins are hand-maintained, that the CUDA value is enforced, and where to look when refreshing the torch value.

Deriving the torch pin at runtime is not really possible: "oldest torch that ships wheels for CUDA X" lives on download.pytorch.org, not in the repo. The guard on the CUDA half catches the drift scenario you describe, and the bump-time checklist covers the torch half.

.github/workflows/weekly-compat.yml
swahtz: We are going to rely on GitHub's built-in notifications for failed scheduled runs plus the Actions tab, rather than add an issue-on-failure job or a Slack integration. An issue per failure turns every transient capacity error into something to triage and close, and a Slack bot is more machinery than a once-weekly job warrants. Anyone else on the team who wants to be notified can subscribe via the repo notification settings.

.github/workflows/weekly-compat.yml
swahtz: Good call; flipped to `cancel-in-progress: false` in 4a27132 with a comment explaining why. For a once-weekly job there is no benefit to cancelling and the downside is an orphaned instance.

Your speculation was right in general, by the way: this is exactly how the orphaned instances we have been seeing were being created on the PR workflows, where a new push cancels the previous run during runner provisioning. #768 fixes that in the shared workflows.

swahtz:
Thanks for the review. Addressed in 4a27132 (the branch was rebased onto `main`, so the history changed):

- **Reusable runner workflows**: all four runner jobs now use `start-ec2-runner.yml` / `stop-ec2-runner.yml`.
- **`weekly_compat` pins**: `load-versions.yml` fails loudly if `weekly_compat.cuda` is not the lowest supported CUDA; comment reworded to say the pins are hand-maintained and how to refresh them.
- **`cancel-in-progress`**: now `false`.

Two items we are deliberately not taking:

- **Failure notifications**: GitHub's built-in notifications for failed scheduled runs are the intended channel. See the inline reply.
- **`torch>=2.7` in `pyproject.toml`**: that is the source-build floor, and it is intentionally lower than the oldest version we test in CI. fvdb-core still builds against 2.7 and 2.8, and raising the floor would block users building from source on those versions for no benefit. What we build and test in CI is not the same thing as what we support as a build target. The undefined-symbol failure you describe applies to prebuilt wheels, which are already served per torch version from the wheel index.

On the minor points: the docs-test omission is deliberate, as noted in the PR description, because the doc examples pin the current in-development versions. The `setup-uv` pins cannot carry a static `# vX.Y.Z` comment because the SHA is read from `versions.json`; that is inherited from `cu130.yml`.

Separately, investigating the `cancel-in-progress` point led to the root cause of the orphaned EC2 instances we have been seeing; that fix is in #768.

.github/workflows/load-versions.yml
```diff
@@ -107,6 +122,16 @@ jobs:
...
+          # weekly_compat.cuda is a hand-maintained pin that must track the
+          # lowest supported CUDA version. Fail loudly if cuda.versions moved
+          # and the compat pin did not, so the weekly job cannot drift into
+          # testing an obsolete combination while reporting green.
+          COMPAT_CUDA=$(jq -r '.weekly_compat.cuda' $CFG)
+          MIN_CUDA=$(jq -r '.cuda.versions | keys_unsorted | sort_by(split(".") | map(tonumber)) | .[0]' $CFG)
+          if [ "$COMPAT_CUDA" != "$MIN_CUDA" ]; then
+            echo "::error file=$CFG::weekly_compat.cuda is $COMPAT_CUDA but the lowest cuda.versions entry is $MIN_CUDA; update weekly_compat (cuda and torch) per docs/installation.rst"
+            exit 1
+          fi
```
harrism: This guard runs in every caller of `load-versions.yml`, and `publish.yml` calls it with `ref: ${{ inputs.branch || '' }}` (publish.yml:116-118): it checks out *that* branch's `versions.json` under *main's* loader. Dispatching a publish from main with `branch=release/v0.5` reads a `versions.json` with no `weekly_compat` block, so `COMPAT_CUDA` is the string `null`, `MIN_CUDA` is `12.8`, and the step exits 1 before any outputs are written; every downstream publish job then fails. Reproduced locally against the real `release/v0.5` file.

Suggest tolerating the missing key so the guard only bites where the block exists:

```suggestion
          COMPAT_CUDA=$(jq -r '.weekly_compat.cuda // empty' $CFG)
          MIN_CUDA=$(jq -r '.cuda.versions | keys_unsorted | sort_by(split(".") | map(tonumber)) | .[0]' $CFG)
          if [ -n "$COMPAT_CUDA" ] && [ "$COMPAT_CUDA" != "$MIN_CUDA" ]; then
            echo "::error file=$CFG::weekly_compat.cuda is $COMPAT_CUDA but the lowest cuda.versions entry is $MIN_CUDA; update weekly_compat (cuda and torch) per docs/installation.rst"
            exit 1
          fi
```

(or move the guard into `weekly-compat.yml`'s own versions-consuming step so only the weekly run is gated). With the key absent, `compat-cuda-patch=$(jq -r '.cuda.versions[.weekly_compat.cuda].patch' $CFG)` further down also errors inside jq ("Cannot index object with null"); harmless inside `$()`, but `// empty` belongs there too.

harrism (changes requested):
Re-review at 4a271324. Of the four earlier threads: the reusable start/stop runner workflows (with the g5.xlarge fallback) and `cancel-in-progress: false` resolve two outright, and the CUDA half of the compat pin is now guarded with the torch half documented as a bump-time checklist, which is fine. The failure-notification point was declined and that is a reasonable call for a weekly job. All the earlier threads can be resolved.

One new item, inline: the guard as written fails `publish.yml` dispatches against release branches. One-line fix.

Checked and clean: cron `0 8 * * 6` (Saturday 08:00 UTC); `contents: read` at the top with `id-token: write` only on the four reusable-workflow callers; the `github.repository` gate on `versions` cascades to skip everything downstream and both stop jobs use `always() && result != 'skipped'`; artifact names match between upload and download; the build/gtest/unit-test jobs are `cu130.yml` minus the `pull_request_target` plumbing with the tag parameterised; `torch-2.9.1+cu130-cp312-linux_x86_64` exists on the index; `8.6+PTX` runs on both the L4 and the A10G fallback; zizmor at the CI-pinned version is clean.

.github/workflows/load-versions.yml
swahtz: Good catch — same blast radius as the CUDA-pin guard. Applied your suggestion for the guard block, and fixed the `compat-cuda-patch` line the same way in 3cc466e7 (it was still doing an unguarded `.weekly_compat.cuda` index, though harmless inside `$()`, it now resolves to empty instead of erroring on stderr).

.github/workflows/load-versions.yml
```diff
@@ -115,6 +140,11 @@ jobs:
             echo "python-default=$(jq -r '.python.default' $CFG)"
             echo "torch-version=$(jq -r '.torch.version' $CFG)"
             echo "torch-full-version=$(jq -r '.torch.full_version' $CFG)"
+            echo "compat-torch-version=$(jq -r '.weekly_compat.torch.version' $CFG)"
+            echo "compat-torch-full-version=$(jq -r '.weekly_compat.torch.full_version' $CFG)"
+            echo "compat-cuda-version=$(jq -r '.weekly_compat.cuda' $CFG)"
+            echo "compat-cuda-tag=cu$(jq -r '.weekly_compat.cuda' $CFG | tr -d '.')"
+            echo "compat-cuda-patch=$(jq -r '.cuda.versions[.weekly_compat.cuda].patch // empty' $CFG)"
```
harrism: Cosmetic, not blocking: `// empty` handles a null result but not a jq runtime error, and indexing `.cuda.versions[null]` is the latter, so with `weekly_compat` absent this still prints `jq: error ... Cannot index object with null` on stderr (exit 5 inside the `$()`, step still exits 0, output empty as intended). Reproduced against `release/v0.5`'s file. If you want it silent:

```suggestion
            echo "compat-cuda-patch=$(jq -r '(.weekly_compat.cuda // empty) as $c | .cuda.versions[$c].patch // empty' $CFG)"
```

(The `aws-*-az-config` lines a few below emit a similar `Cannot iterate over null` for `release/v0.5`, whose `versions.json` has `aws.subnet` rather than `aws.subnets`; that is pre-existing and identical on main.)

harrism (approved):
Re-checked at 3cc466e7. The guard now reads `.weekly_compat.cuda // empty` and only fires when the key is present, and the `Parse versions.json` script was re-run locally against the real `release/v0.5` and `main` version files (no `weekly_compat` block): exit 0 in both, where 4a271324 exited 1. With the pin deliberately set to 13.2 it still exits 1, so the guard keeps its teeth. The `publish.yml` dispatch-against-a-release-branch scenario is unblocked. One cosmetic leftover inline; not blocking.

PR #746, opened 2026-08-21
Block PRs that drift env pins from fvdb-reality-capture

harrism:
## Problem

The nightly benchmarks in [openvdb/fvdb-reality-capture](https://github.com/openvdb/fvdb-reality-capture) build the fvdb-core wheel from **this repo's** `env/build_environment.yml`, then install it into their own benchmark environment. When those pins disagree, the wheel is compiled against one libtorch and loaded against another, and their nightly dies at `import fvdb`:

```
ImportError: .../site-packages/fvdb/libfvdb.so: undefined symbol:
  _ZN3c1010ValueErrorC1ENS_14SourceLocationENSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
```

This has happened three times. Most recently #738 (PyTorch 2.13) bumped `pytorch-gpu` from 2.11.0 to 2.13.0 while their benchmark env still pinned 2.11.0, breaking their nightly the same day. The two previous occurrences were fixed downstream in their `1b6956f` (2.8→2.10) and `6c2ede6` (2.10→2.11). A comment in the env file did not prevent the third.

## Change

Add a `downstream-env-drift` job to the code style workflow comparing `pytorch-gpu`, `cuda-version` and `python` against their benchmark env on `main`.

## Design notes

**It blocks rather than warns.** A GitHub `::warning::` is an annotation — it does not post to the PR conversation, and the check still reports green. A warning buried under passing CI is exactly the failure mode this is meant to prevent, so the check fails instead.

**Intentional bumps are expected to fail, and there is an escape hatch.** The companion change lives in another repository and cannot merge in the same commit, so a hard gate with no override would deadlock a legitimate PyTorch bump. Applying the **`allow-downstream-drift`** label acknowledges the drift and lets the PR proceed; the matching update then lands in fvdb-reality-capture. *(That label needs to be created in this repo before the escape hatch works.)*

**An unreadable pin fails.** Without that, renaming or reformatting a pin line would make both `grep -oP` results empty, compare equal, and pass — a guard that quietly stops guarding.

**Hosted in `codestyle.yml`** because it needs only `pull_request` and `contents: read`, so it works for fork PRs with no elevated permissions and no `pull_request_target`.

## Testing

Dry-ran all four paths against the real environment files from both repos:

| Case | Result |
|---|---|
| Real drift (2.13.0 vs 2.11.0), no label | exit 1, error annotation naming both versions |
| Real drift, with `allow-downstream-drift` | exit 0, notice annotation |
| Pins synced | exit 0 |
| Pin line renamed/reformatted | exit 1, "could not read" error |

## Sequencing

This check fails against fvdb-reality-capture `main` today, because the 2.11→2.13 drift is real and present. It goes green once their [#318](https://github.com/openvdb/fvdb-reality-capture/pull/318) merges — **that should land first**, otherwise this red-Xes every fvdb-core PR.

Companion PR adding the mirror-image fail-fast gate to their nightly: [fvdb-reality-capture#319](https://github.com/openvdb/fvdb-reality-capture/pull/319).

### DIFF
--- a/.github/workflows/codestyle.yml
+++ b/.github/workflows/codestyle.yml
@@ -92,3 +92,54 @@ jobs:
           set +e
           git grep -n "	" -- ':!*/codestyle.yml' ':!*.svg' ':!*.webp' ':!*.cmd' ':!*.png' ':!*.wlt' ':!*.jpg' ':!*.gif' ':!*.mp4' ':!*.pt' ':!*.pth' ':!*.nvdb' ':!*.npz' ':!*.gitmodules' ':!*/wip/*'
           test $? -eq 1
+
+  # The nightly benchmarks in openvdb/fvdb-reality-capture build the fvdb-core
+  # wheel from THIS repo's env/build_environment.yml and install it into their
+  # own benchmark environment. If the pins disagree, that wheel is compiled
+  # against one libtorch and loaded against another, and their nightly dies at
+  # `import fvdb` with an undefined-symbol ImportError. This has happened three
+  # times; a comment in the env file did not prevent the third.
+  #
+  # This gate blocks rather than warns: a GitHub `::warning::` is an annotation,
+  # so a passing (green) check hides it and the drift lands anyway.
+  #
+  # Intentional bumps are expected to fail here, because the companion change
+  # lives in another repository and cannot merge simultaneously. Apply the
+  # 'allow-downstream-drift' label to acknowledge and proceed, then land the
+  # matching update in fvdb-reality-capture.
+  downstream-env-drift:
+    name: Downstream env pins (fvdb-reality-capture)
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      with:
+        persist-credentials: false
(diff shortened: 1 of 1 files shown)


harrism:
Closing in favour of a prevention-based approach.

Per @swahtz's suggestion, openvdb/fvdb-reality-capture#319 now *derives* the `pytorch-gpu`/`cuda-version`/`python` pins from this repo's `env/build_environment.yml` via a `scripts/generate_benchmark_env.py` script, rather than detecting drift after the fact. The nightly regenerates from the exact fvdb-core commit it builds the wheel from, and a PR check keeps the committed file honest for `docker build` and local use.

With that in place this gate protects nothing: frc's CI can no longer use a mismatched pin, so blocking fvdb-core PRs adds friction without preventing a failure. The `allow-downstream-drift` label and the companion-PR search go away with it.

Worth noting separately: this repo's four env files (`build`, `dev`, `test`, `learn`) each pin PyTorch/CUDA/Python independently. They agree today (13.0 / 3.12 / 2.13.0), but nothing enforces that. A single source for those four would be a genuine improvement here, and would also make "which file does downstream derive from?" unambiguous.

PR #748, opened 2026-08-21
Use the batch bbox min for the default morton/hilbert encoding offset

swahtz:
## Summary

The default (`offset=None`) path of the eight morton/hilbert encoding functions in `fvdb/functional/_topology.py` computed its coordinate offset by materializing every active voxel coordinate as an `(N, 3)` tensor (`active_grid_coords`), running a device-wide min reduction, and syncing the result to the host — all to produce a 3-element offset. The batch bbox min is the same componentwise minimum and is already tracked as host-side metadata (`GridBatchData::mTotalBBox`), so the default offset now costs nothing: no kernel launch, no `(N, 3)` allocation, no device sync.

One-line change per site (×8: `morton`/`morton_zyx`/`hilbert`/`hilbert_zyx`, batch and single variants):

```python
offset = -torch.min(_fvdb_cpp.active_grid_coords(grid_data).jdata, dim=0).values   # before
offset = -grid_data.total_bbox[0]                                                  # after
```

### Performance

`GridBatch.morton()` wall time (RTX sm_120, median of 30):

| grid size | before | after | speedup |
|---|---|---|---|
| 400k voxels | 87 µs | 52 µs | 1.67× |
| 5M voxels | 290 µs | 162 µs | 1.79× |

The eliminated work scales with voxel count while the bbox read is O(1), so the win grows with grid size. Removing the device→host sync also makes the encoders launch-async-friendly.

### Correctness

A new regression test (`TestDefaultOffsetUsesBBoxMin`) pins the equivalence: default-offset codes are bitwise identical to explicitly passing `-min(active coords)` for all eight encoders, on CPU and CUDA, including sliced `GridBatch` views (whose bbox metadata is recomputed per view — verified). Full `test_morton_hilbert`, `test_basic_ops`, and `test_basic_ops_single` suites pass (444 tests).

### DIFF
--- a/fvdb/functional/_topology.py
+++ b/fvdb/functional/_topology.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 """Functional API for grid topology operations (coarsen, refine, dual, dilate, etc.)."""
+
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


matthewdcong (approved):
Nice one!

PR #750, opened 2026-08-21
Fix incorrect channel ordering in FusedSSIM prefetch

matthewdcong:
Previously, prefetching in FusedSSIM assumed a NHWC ordering which matches the storage ordering of the image on disk. However, the image is permuted during the training pipeline to NCHW in order to leverage the fused SSIM kernels. This incorrect prefetching led to a significantly amount of unnecessary UVM traffic and fixing it improves fused SSIM scalability dramatically. Measured on 2x RTX 3090s:

| Resolution | Inference: before | Inference: after | Improvement | Training: before | Training: after | Improvement | Backward: before | Backward: after | Improvement |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1920x1080 | 1.110 ms | 0.559 ms | 1.99x / 49.7% | 1.707 ms | 0.971 ms | 1.76x / 43.1% | 2.353 ms | 1.108 ms | 2.12x / 52.9% |
| 3840x2160 | 4.077 ms | 1.227 ms | 3.32x / 69.9% | 6.143 ms | 2.084 ms | 2.95x / 66.1% | 7.802 ms | 2.222 ms | 3.51x / 71.5% |
| 7680x4320 | 17.089 ms | 3.942 ms | 4.33x / 76.9% | 24.976 ms | 6.078 ms | 4.11x / 75.7% | 31.815 ms | 6.818 ms | 4.67x / 78.6% |

### DIFF
--- a/src/fvdb/detail/ops/gsplat/FusedSSIM.cu
+++ b/src/fvdb/detail/ops/gsplat/FusedSSIM.cu
@@ -649,12 +649,13 @@ appendImagePrefetchRanges(std::vector<void *> &prefetchPointers,
             // adjacent devices never issue prefetches for overlapping logical ranges.
             const size_t firstRow = firstTileRow * BLOCK_Y;
             const size_t lastRow  = std::min(lastTileRow * BLOCK_Y, static_cast<size_t>(H));
+            const size_t rowCount = lastRow - firstRow;
 
             for (int channel = 0; channel < CH; ++channel) {
                 const size_t elementOffset = batch * tensor.stride(0) + channel * tensor.stride(1) +
                                              firstRow * tensor.stride(2);
                 auto *pointer          = tensorData + elementOffset * scalarSize;
-                const size_t byteCount = (lastRow - firstRow) * tensor.stride(2) * scalarSize;
+                const size_t byteCount = rowCount * static_cast<size_t>(W) * scalarSize;
 
                 if (prefetchPointers.size() > firstTensorRange) {
                     auto *previousEnd =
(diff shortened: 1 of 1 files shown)


src/fvdb/detail/ops/gsplat/FusedSSIM.cu
```diff
@@ -582,45 +583,92 @@ fusedSSIMBackwardCUDA(double C1,
...
+            // Prefetch only the rows owned by this device. Halo reads remain demand-driven so
+            // adjacent devices never issue prefetches for overlapping logical ranges.
+            const size_t firstRow = firstTileRow * BLOCK_Y;
+            const size_t lastRow  = std::min(lastTileRow * BLOCK_Y, static_cast<size_t>(H));
+
+            for (int channel = 0; channel < CH; ++channel) {
+                const size_t elementOffset = batch * tensor.stride(0) + channel * tensor.stride(1) +
+                                             firstRow * tensor.stride(2);
+                auto *pointer          = tensorData + elementOffset * scalarSize;
+                const size_t byteCount = (lastRow - firstRow) * tensor.stride(2) * scalarSize;
```
swahtz: **💡 suggestion:** ‏Codex found this small gotchya:  When `H==1` and the tensor input is a stepped or sliced view with `stride(2) > W`, PyTorch may still report it as contiguous because singleton-dimension strides are ignored.  Computing `byteCount` from `stride(2)` can then pass a range larger than the tensor's allocation to `cudaMemPrefetchAsync`.  It suggested deriving the range length from the logical row width:  `rowCount * W * scalarSize`

swahtz (approved):
Looks great.  Small nit that a Codex review found but otherwise lgtm

PR #751, opened 2026-08-22
Speed up SH backward camera-gradient reduction

matthewdcong:
Instead of allocating a full view-direction gradient intermediate in unified memory, we instead allocate more compact device-local per-block sums. The per-block sums are written to after a block reduce in the SH kernel followed by a CUB `DeviceSegmentedReduce`. On dual RTX 3090s:

| Gaussians | Cameras | Before full op | After full op | Speedup |
| ---: | ---: | ---: | ---: | ---: |
| 1M | 1 | 1.671 ms | 0.948 ms | 43.3% |
| 1M | 4 | 4.374 ms | 3.347 ms | 23.5% |
| 10M | 1 | 12.774 ms | 5.799 ms | 54.6% |
| 10M | 4 | 22.928 ms | 15.813 ms | 31.0% |
| 100M | 1 | 135.826 ms | 57.659 ms | 57.5% |
| 100M | 4 | 260.342 ms | 156.974 ms | 39.7% |

### DIFF
--- a/src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
+++ b/src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
@@ -298,7 +298,6 @@ __global__ __launch_bounds__(DEFAULT_BLOCK_DIM) void
 computeShBackward(
     const int64_t offset,
     const int64_t count,
-    const int64_t C,
     const int64_t N,
     const int64_t K,
     const int64_t D,
@@ -316,16 +315,16 @@ computeShBackward(
     T *__restrict__ outDLossDMeans,   // [N, 3] optional
     T *__restrict__ outDLossDViewDirs // [C, N, 3] optional
 ) {
-    // parallelize over C * N * D
-    auto idx = blockIdx.x * blockDim.x + threadIdx.x; // cidx * N * D + gidx * D + c
-    if (idx >= count * C * D) {
+    // Each block handles one camera and parallelizes over count * D.
+    const auto idx = blockIdx.x * blockDim.x + threadIdx.x; // gidx * D + c
+    if (idx >= count * D) {
         return;
     }
 
-    const auto eid = idx / D;              // cid * N + gid
-    const auto cid = eid / count;          // camera index
-    const auto gid = eid % count + offset; // gaussian index
-    const auto c   = idx % D;              // render channel
+    auto cid        = blockIdx.y;    // camera index
+    const auto gidx = idx / D;
+    const auto gid  = gidx + offset; // gaussian index
+    const auto c    = idx % D;       // render channel
     if (radii != nullptr &&
         (radii[(cid * N + gid) * 2 + 0] <= 0 || radii[(cid * N + gid) * 2 + 1] <= 0)) {
         return;
@@ -360,7 +359,7 @@ computeShBackward(
     if (dLossDViewDirPtr != nullptr) {
         auto activeThreads = cooperative_groups::coalesced_threads();
         auto pairThreads =
-            cooperative_groups::labeled_partition(activeThreads, static_cast<int>(eid));
+            cooperative_groups::labeled_partition(activeThreads, static_cast<int>(gidx));
         dLossDViewDir.x =
             cooperative_groups::reduce(pairThreads, dLossDViewDir.x, cooperative_groups::plus<T>());
         dLossDViewDir.y =
@@ -704,10 +703,10 @@ dispatchEvaluateSphericalHarmonicsBwd<torch::kCUDA>(
                 dLossDSh0Coeffs, dLossDShNCoeffs, dLossDMeans, dLossDWorldToCamMatrices);
         }
 
-        computeShBackward<scalar_t><<<NUM_BLOCKS, DEFAULT_BLOCK_DIM, 0, stream>>>(
+        const dim3 shGrid(GET_BLOCKS(N * D, DEFAULT_BLOCK_DIM), C);
+        computeShBackward<scalar_t><<<shGrid, DEFAULT_BLOCK_DIM, 0, stream>>>(
             0,
             N,
-            C,
             N,
             K,
             D,
@@ -1029,12 +1028,11 @@ dispatchEvaluateSphericalHarmonicsBwd<torch::kPrivateUse1>(
             std::tie(elementOffset, elementCount) = deviceChunk(N, deviceId);
 
             if (elementCount > 0) {
-                const auto NUM_BLOCKS = GET_BLOCKS(C * elementCount * D, DEFAULT_BLOCK_DIM);
+                const dim3 shGrid(GET_BLOCKS(elementCount * D, DEFAULT_BLOCK_DIM), C);
 
-                computeShBackward<scalar_t><<<NUM_BLOCKS, DEFAULT_BLOCK_DIM, 0, stream>>>(
+                computeShBackward<scalar_t><<<shGrid, DEFAULT_BLOCK_DIM, 0, stream>>>(
                     elementOffset,
                     elementCount,
-                    C,
                     N,
                     K,
                     D,
--- a/src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
+++ b/src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
@@ -373,8 +373,8 @@ computeShBackward(
     T *__restrict__ outDLossDViewDirBlockSums // [C, gridDim.x, 3] optional, unpacked only
 ) {
     // Each block handles one camera and parallelizes over count * D.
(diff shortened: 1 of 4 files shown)


src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -742,19 +764,41 @@ dispatchEvaluateSphericalHarmonicsBwd<torch::kCUDA>(
...
+                    tempStorageBytes,
+                    dLossDViewDirBlockSums.data_ptr<scalar_t>(),
+                    reducedViewDirectionGradients.data_ptr<scalar_t>(),
+                    C,
+                    NUM_BLOCKS.x,
+                    stream));
+                C10_CUDA_CHECK(cudaFreeAsync(tempStorage, stream));
+
+                const auto viewMatrixBlocks = GET_BLOCKS(C, DEFAULT_BLOCK_DIM);
+                viewDirectionGradientsToWorldToCamMatrices<scalar_t>
```
swahtz, line 783: **🚩 issue:** ‏Previously we used to derive the local camera ID with `numWorldToCamMatrices` whereas the new kernel derives `localCameraId` from the thread index over `[0, C)` (C = `numCameras`).  In these changes, nothing on this path guarantees `dLossDWorldToCamMatrices` actually has `C` rows.  The `TORCH_CHECK_VALUE(worldToCamMatrices.size(0) == C, ...)` at L684 sits inside the `shDegreeToUse > 0` guard (L648), while this gradient branch runs for any `K > 1` (L697). So `shDegreeToUse == 0 + computeDLossDWorldToCamMatrices + worldToCamMatrices.size(0) < numCameras` writes rows [0, C) into a `[size(0), 4, 4]` tensor — an out-of-bounds device write.

I'd suggest hoisting the shape check to the entry of the K > 1 branch (L697), outside the `shDegreeToUse > 0` guard — and mirroring it at L910 in the PrivateUse1 dispatcher.

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -1050,25 +1087,55 @@ dispatchEvaluateSphericalHarmonicsBwd<torch::kPrivateUse1>(
...
+                        tempStorageBytes,
+                        dLossDViewDirBlockSums,
+                        reducedViewDirectionGradients,
+                        C,
+                        NUM_BLOCKS.x,
+                        stream));
+                    C10_CUDA_CHECK(cudaFreeAsync(tempStorage, stream));
+
+                    const auto viewMatrixBlocks = GET_BLOCKS(C, DEFAULT_BLOCK_DIM);
+                    viewDirectionGradientsToWorldToCamMatrices<scalar_t>
```
swahtz, line 1090: Same unvalidated `size(0) == C` corner as the kCUDA launch (see comment above) — here it also overruns the per-device cudaMallocAsync buffer from L952. A `TORCH_CHECK(worldToCamMatrices.size(0) == C)` at L910 covers this dispatcher.

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -313,169 +368,121 @@ computeShBackward(
...
+    T *__restrict__ outDLossDMeans,           // [N, 3] optional
+    T *__restrict__ outDLossDViewDirs,        // [C, N, 3] optional, packed only
+    T *__restrict__ outDLossDViewDirBlockSums // [C, gridDim.x, 3] optional, unpacked only
 ) {
-    // parallelize over C * N * D
-    auto idx = blockIdx.x * blockDim.x + threadIdx.x; // cidx * N * D + gidx * D + c
-    if (idx >= count * C * D) {
-        return;
+    // Each block handles one camera and parallelizes over count * D.
+    const auto idx = blockIdx.x * blockDim.x + threadIdx.x; // gidx * D + c
```
swahtz: **🚩 issue:** ‏`auto` is going to produce an `uint32_t` here but it's likely we could have values that overflows that range.  Especially in the case of a large channel count (at 32 channels, ~134M Gaussian will hit it).
 
```suggestion
    const int64_t idx = static_cast<int64_t>(blockIdx.x) * blockDim.x + threadIdx.x; // gidx * D + c
```

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
swahtz: **🚩 issue:** ‏Related to the comment above, we should fix this potential overflow issue here too.


```suggestion
    const int64_t idx = static_cast<int64_t>(blockIdx.x) * blockDim.x + threadIdx.x; // cidx * N * D + gidx * D + c
```

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -742,19 +764,41 @@ dispatchEvaluateSphericalHarmonicsBwd<torch::kCUDA>(
...
                         numWorldToCamMatrices,
                         worldToCamMatrices.data_ptr<scalar_t>(),
                         reducedViewDirectionGradients.data_ptr<scalar_t>(),
                         dLossDWorldToCamMatrices.data_ptr<scalar_t>());
                 C10_CUDA_KERNEL_LAUNCH_CHECK();
             } else {
-                reduceUnpackedViewDirectionGradientsToWorldToCamMatrices<AssignOutput, scalar_t>
-                    <<<numWorldToCamMatrices, DEFAULT_BLOCK_DIM, 0, stream>>>(
-                        N,
+                size_t tempStorageBytes = 0;
```
swahtz: **💅 polish:** ‏ This CUB temp storage is allocated with raw cudaMallocAsync/cudaFreeAsync instead of PyTorch's caching allocator but we have some  FVDB_CUB_WRAPPER / FVDB_CUB_WRAPPER_ASYNC  macros in `IntersectGaussianTiles.cu` and `BuildSparseGaussianTileLayout.cu` to accomplish this.  I'd suggest this PR refactors into a util header we could re-use here (and in another instance below).

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
```diff
@@ -1050,25 +1087,55 @@ dispatchEvaluateSphericalHarmonicsBwd<torch::kPrivateUse1>(
...
                 if (computeDLossDWorldToCamMatrices) {
-                    const auto numWorldToCamMatrices = worldToCamMatrices.size(0);
-                    reduceUnpackedViewDirectionGradientsToWorldToCamMatrices<SystemAtomicAddOutput,
-                                                                             scalar_t>
-                        <<<numWorldToCamMatrices, DEFAULT_BLOCK_DIM, 0, stream>>>(
-                            N,
-                            elementOffset,
-                            elementCount,
+                    size_t tempStorageBytes = 0;
+                    C10_CUDA_CHECK(reduceViewDirectionGradientBlockSums<scalar_t>(
```
swahtz: **💅 polish:** ‏ Similar suggestion as above for using `FVDB_CUB_WRAPPER` here for a repeated pattern.

swahtz (changes requested):
Overall looks good to me, just a few minor checks/fixes.

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
matthewdcong: This should use `FVDB_CUB_WRAPPER_ASYNC` as opposed to `FVDB_CUB_WRAPPER` since it is the `kPrivateUse1` path. Changed.

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
matthewdcong: Good catch. This should use the caching allocator via `FVDB_CUB_WRAPPER`. Fixed.

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
matthewdcong: Simplified and covered by removing the unnecessary `shDegreeToUse > 0` guard on the shape asserts.

src/fvdb/detail/ops/gsplat/EvaluateSphericalHarmonicsBackward.cu
matthewdcong: Fixed, see above

swahtz (approved):
Let's go! 🚀

Issue #741, opened 2026-08-15
Generative Transposed Convolution


lishiyu005:
Dear developers,

I greatly appreciate this good work. 

Do you provide 3D generative transposed convolution operation that allows to create new possible coordinates as in Minkowski Engine: https://github.com/NVIDIA/MinkowskiEngine/blob/b71caef9b45ab32c02d2c1b5f48f13c1eb5df526/MinkowskiEngine/MinkowskiConvolution.py#L539

If not do you plan to add a shape completion task similar to https://github.com/NVIDIA/MinkowskiEngine/blob/b71caef9b45ab32c02d2c1b5f48f13c1eb5df526/examples/completion.py?

Thank you very much.

swahtz:
Hi @lishiyu005 ,

Thanks very much, we're glad that you find it useful!

Yes, we do supply functionality that is equivalent to the MinkowskiEngine generative transposed convolution idea.  In fact, fixing up some issues in this functionality was part of the recent PR #726 and will be in the imminent v0.6 release, so I'd recommend you build from source or install nightly builds to try this out.  

`ConvolutionPlan.from_grid_batch_transposed(kernel_size, stride, source_grid, target_grid=None)` selects the `COMPLETE` topology policy and builds the target as `source_grid.conv_transpose_grid(kernel_size, stride)`.   New voxels are generated everywhere any kernel tap can reach, which is [the "generative" behavior ](https://github.com/openvdb/fvdb-core/blob/2c724d102f0b7f40bc7f928256b9570b57d5d89c/fvdb/convolution_plan.py#L761).

The target grid's voxels can be masked by building a GridBatch from a masked ijk by gathering features with `inject_from`.  Those masked features can be a classifier of which voxels to mask off and can be differentiable.  However, we don't have a one-call layer to do this, though it would be very useful… in fact we [had examples that did this for the older API as part of a shape completion example](https://github.com/openvdb/fvdb-core/blob/2c724d102f0b7f40bc7f928256b9570b57d5d89c/examples/wip/structure_prediction_net.py#L101) but they're old enough now that they don't structurally work but the idea is the same.

Let me go and write this layer and integrate it into an example, as you suggest, since the logic for doing this is common to these types of problems but also perhaps a bit tricky to do right.  I can then add it to our `nn` module and as an example script which will update our old structure prediction example.

swahtz:
[...]

PR #753, opened 2026-08-31
Add fvdb.nn.Prune and generative shape completion / VAE examples

swahtz:
Addresses #741.

## Summary

Scope note: an earlier revision also batched `GridBatch.pruned_grid` on CUDA through the leaf-mask builder from #757. That commit is reverted here so this PR is only the layer and the examples; batched pruning follows on the NanoVDB `TopologyBuilder` engine in #785.

- **`fvdb.nn.Prune`** — the `MinkowskiPruning` analog: prunes grid topology and its aligned features by a per-voxel boolean mask (`pruned_grid` + `rmask`, differentiable through kept rows). Registered in docs; 13 unit tests.
- **`examples/shape_completion.py` / `examples/shape_vae.py`** — one-to-one analogs of MinkowskiEngine's `completion.py` and `vae.py`, built on the generative transposed convolution from #726 (`from_grid_batch_transposed(..., target_grid=None)`) plus `Prune`.
- **`notebooks/04_shape_vae.ipynb`** — trains the VAE and explores its latent space: reconstructions, prior sampling, latent interpolation (ballet flat → boot), and an optional ipywidgets explorer. Shapes render as lit voxel surfaces.
- **Dataset**: the VAE trains on the 254-mesh GSO "Shoe" subset (CC-BY 4.0, voxel-foundation/fvdb-example-data#1) via a new `load_gso_shoes()` loader and data-revision bump; completion uses the dragon + happy meshes.
- Removes `examples/wip/structure_prediction_net.py` (targeted a removed API surface; superseded by the new examples).

## Design notes

- Per-level GT targets come from a `conv_grid(2, 2)` pyramid queried with `coords_in_grid` — not `coarsened_grid`, whose block-centroid lattice differs from what `conv_transpose_grid(2, 2)` inverts.
- ME's teacher forcing (`keep |= target`) and per-level BCE are reproduced verbatim; eval decodes without forcing.
- The VAE's dense-neck seed adds a learned per-voxel positional embedding — a uniform broadcast latent leaves neck voxels spatially indistinguishable after k2 transposes.

## Training-loop profile

The first CI run failed on the notebook's training cell exceeding nbmake's 300 s per-cell default, which prompted a profile of the VAE step at batch 16 on an RTX PRO 6000.

- Data loading is not a factor: minibatch indexing is 1% of the step.
- Learning rate is not a factor: 3e-4, 1e-3, and 3e-3 converge to the same loss and per-level IoU at every checkpoint over 1500 iterations, and the loss is still falling at 1500. The model is undertrained rather than misconfigured.
- After #757 batched the transposed-conv grid builds, `pruned_grid` is the remaining per-grid loop: 28 ms of a 51 ms decoder forward at batch 16 (one NanoVDB `PruneGrid` per member plus `mergeGridHandles`). Batching it is deferred to the NanoVDB-based batched topology engine tracked in #785, so this PR leaves `pruned_grid` as it is on main; the training step is about 89 ms.
- What remains is launch-bound rather than any single slow op: 17 plan builds (~21 ms, the stride-2 ones spend most of their time in host-side coverage validation), the gather-scatter conv forward and backward (~35 ms across ~940 small matmuls per step), and about 18k kernel launches per step in total. Those are library follow-ups, not example changes.

`NUM_ITERATIONS` drops from 1500 to 1000. Two thirds of the time keeps most of the result (final loss 0.27 vs 0.25), and at ~89 ms per step the notebook cell runs in about 90 s, inside nbmake's 300 s limit. The notebook reads the constant from the example, so one change covers both.

## Results / verification

- Completion IoU ≈ 0.75 (500 iters, ~40 s); VAE reconstruction IoU ≈ 0.36 on 8 shapes (1000 iters, ~66 s including data load), prior samples decode to plausible shoes.
- Prune-related tests across `test_basic_ops`, `test_basic_ops_single`, `test_prune_single_voxel`, `test_sliced_batch`, `test_inject`, `test_nn_modules`: 83 passed.
- Notebook re-executed with committed outputs; `clang-format` and `black --line-length 120` clean.


swahtz:
Added `notebooks/04_shape_vae.ipynb`: trains the shape VAE from `examples/shape_vae.py` and explores its latent space interactively — reconstructions (IoU 0.999), prior sampling, batched latent interpolation (car → dragon), and an ipywidgets latent explorer (blend two shapes + prior noise) with a static fallback when ipywidgets is unavailable. Executed outputs are stored so it renders on GitHub; passes `pytest --nbmake` in ~87 s on GPU. `ipywidgets` added to the `fvdb_learn` environment.

swahtz:
Dataset upgrade: the shape VAE (example + notebook) now trains on the 254-model Google Scanned Objects **Shoe** subset (CC-BY 4.0, added to fvdb-example-data in voxel-foundation/fvdb-example-data#1) instead of the 6 bundled meshes — one category with real intra-class variation, mirroring the role of ModelNet40 chairs in ME's `vae.py`. Adds `fvdb.utils.examples.load_gso_shoes()` and bumps the pinned data revision; training runs random minibatches over the pre-voxelized `GridBatch` with a once-built `conv_grid` pyramid sub-indexed per batch. The notebook now interpolates a ballet flat into a tall boot.

CI note: `pytest tests/test_examples.py -k shape` now takes ~4:07 (was ~2:02) due to the larger dataset and minibatch training.

Perf note: generative training is bound by per-iteration generated-topology **grid construction** (`conv_transpose_grid`; the kernel maps themselves are cheap) — profiled and filed as #755 rather than worked around here, so the examples stay semantically 1:1 with MinkowskiEngine.

Rishit-dagli (commented):
are there any plans for a similar shape vae example but for 4d volume?

fvdb/nn/modules.py
```diff
@@ -261,6 +261,58 @@ def forward(
...
+            mask (JaggedTensor): Boolean keep-mask with one entry per active voxel of ``grid``.
+                Shape: ``(batch_size, num_voxels_b)``.
+
+        Returns:
+            pruned_data (JaggedTensor): Features of the surviving voxels, aligned with ``pruned_grid``.
+            pruned_grid (GridBatch): A new :class:`fvdb.GridBatch` containing only the voxels
+                where ``mask`` is ``True``.
+        """
+        pruned_grid = grid.pruned_grid(mask)
+        pruned_data = data.rmask(mask.jdata)
```
harrism, line 321: > _Analysis by Claude Code, run by @harrism._

**`forward` never checks that `data`'s jagged partitioning matches `grid`/`mask`, so a mismatch produces silently misaligned features instead of an error.**

Neither callee catches it:

* `rmask` validates only the *flat* mask (`JaggedTensor.cpp:584-590` — device, `dim() == 1`, `size(0) == mData.size(0)`, bool dtype). It never looks at offsets.
* `pruneGrid` validates only `mask.num_tensors() == batchSize()` (`BuildPrunedGrid.cu:151-155`).

So with `grid` and `mask` partitioned `[2, 2]` and `data` partitioned `[1, 3]` — same total of 4 voxels — an all-true mask passes both checks and returns `pruned_data` still split `[1, 3]` alongside a `pruned_grid` split `[2, 2]`. Nothing raises. A downstream `BatchNorm` or convolution then operates on rows attributed to the wrong grid in the batch, and the only symptom is worse training.

This is easy to hit when `data` comes from a different branch of a U-Net than the grid it is being pruned against, which is exactly the pattern the new examples use.

A check in `forward` costs nothing next to the kernel launches below it:

```python
if data.jdata.shape[0] != grid.total_voxels:
    raise ValueError(
        f"data has {data.jdata.shape[0]} rows but grid has {grid.total_voxels} voxels"
    )
if not torch.equal(data.joffsets, grid.joffsets):
    raise ValueError("data and grid must have the same per-grid partitioning")
```

harrism (commented):

## Review: `fvdb.nn.Prune` + shape completion / VAE examples

Two defects below. The layer's core logic is sound — gradient flow through `rmask` is autograd-tracked and `test_prune_gradient` pins the exact expected gradient; all-false and all-true masks are covered by `test_prune_forward_all_false_all_true`, and `pruneGrid` preserves `grid_count` with zero-voxel grids; the voxel-order alignment uses the same `pruned_grid` + `rmask(mask.jdata)` idiom already established in `ClipGrid.cu`. The examples' fvdb API usage checks out against the real signatures, and the encoder/decoder channel pyramid in `shape_completion.py` has no off-by-one.

docs/api/nn.rst
```diff
@@ -4,6 +4,7 @@ Neural Network Layers and Blocks
 .. autoclass:: fvdb.nn.MaxPool
 .. autoclass:: fvdb.nn.AvgPool
 .. autoclass:: fvdb.nn.UpsamplingNearest
+.. autoclass:: fvdb.nn.Prune
```
harrism, line 7: > _Analysis by Claude Code, run by @harrism._

**`Prune`'s argument documentation renders nowhere.**

`Prune` is the only `fvdb.nn` class that documents its Args and Returns solely on `forward` rather than in the class docstring. That combines badly with two things already in the repo:

1. `_trace_fvdb_nn_forward` (`modules.py:15-30`) rebinds `forward` to a plain `_forward(self, *args, **kwargs)` with no `functools.wraps`, so `Prune.forward.__doc__` is `None` and the signature degrades to `(self, *a, **k)`.
2. This `.. autoclass::` has no `:members:`, and `docs/conf.py:143` sets only `undoc-members`.

Net effect: the Args/Returns block added in this PR appears neither in the rendered docs nor in `help(fvdb.nn.Prune.forward)`. The decorator is pre-existing and every other class survives it by documenting itself at class level — `Prune` is the first one it fully de-documents.

Either fix works. Moving the Args/Returns into the class docstring matches the convention of the surrounding classes and needs no other change. Adding `functools.wraps(old_forward)` to the decorator fixes it for every future class too, at the cost of touching shared code in this PR.

notebooks/04_shape_vae.ipynb
```diff
@@ -0,0 +1,519 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "8ea3da0a",
+   "metadata": {},
+   "source": [
+    "# Generative Shape VAE: Interactive Sampling Demo\n",
+    "\n",
+    "This notebook trains the sparse voxel shape VAE from [`examples/shape_vae.py`](../examples/shape_vae.py) and then explores its latent space interactively:\n",
```
harrism: Suggest spelling out Variational AutoEncoder and linking to a definition somewhere.

swahtz:
> are there any plans for a similar shape vae example but for 4d volume?

I don't think we currently have a prescribed way to approach that (I'm assuming changing topology over time).  I'd be happy to include an example based on what you're working on if you'd like to include it, perhaps in `fvdb-examples`?

fvdb/nn/modules.py
swahtz: Good catch, fixed in 650fd265. `forward` now validates both `data` and `mask` against `grid` (tensor count, total rows, then per-grid `joffsets`) and raises `ValueError` on mismatch. The mask case matters too: `pruneGrid` indexes each per-grid mask by voxel index without bounds checks, so an oversized per-grid mask was undefined behaviour rather than a silent misalignment. The offsets compare is skipped for a zero-grid batch, where `GridBatch.joffsets` is empty but a matching `JaggedTensor` reports `[0]`. Covered by `test_prune_rejects_mismatched_partitioning`.

docs/api/nn.rst
swahtz: Fixed both ways in 650fd265. The decorator now uses `functools.wraps`, so every `fvdb.nn` module keeps its `forward` docstring and signature (`AvgPool`, `MaxPool`, and `UpsamplingNearest` also document forward args on `forward` and were losing them too). `Prune` additionally documents Args/Returns/Raises in the class docstring, matching the surrounding classes; I confirmed with a Sphinx build using the repo's autodoc settings that all three sections render on the `Prune` entry. `test_prune_forward_keeps_docstring` pins the decorator behaviour.

notebooks/04_shape_vae.ipynb
swahtz: Done in 650fd265. The title and first sentence now spell out Variational Autoencoder and link to Kingma & Welling (arXiv 1312.6114).

swahtz:

The `fvdb.nn.Prune` layer, the shape-completion and shape-VAE examples and the notebook in this PR do not depend on the engine; they can be re-opened at any time on top of today's per-member `pruned_grid`, with the batched Prune pass following on the mask engine.

swahtz:
Reopened, this shouldn't have been closed, we should just remove the batched pruned grid component.

swahtz:
Reopened with the batched `pruned_grid` pass removed: `ad8586fb` reverts `1abf3d52` (the `Prune` op in `BatchedTopologyBuilder.cuh`, the `BuildPrunedGrid.cu` rewrite and its builder tests) and `d3deac26` merges current main (#773). The branch now touches only `fvdb.nn.Prune`, the two examples, the notebook, docs and tests; `pruned_grid` stays on main's per-member path. Batched pruning will come back on the NanoVDB TopologyBuilder engine in #785.

fvdb/nn/modules.py
```diff
@@ -261,6 +263,74 @@ def forward(
...
+        self._check_partitioned_like(grid, data, "data")
+        self._check_partitioned_like(grid, mask, "mask")
+        pruned_grid = grid.pruned_grid(mask)
+        pruned_data = data.rmask(mask.jdata)
+        return pruned_data, pruned_grid
+
+    @staticmethod
+    def _check_partitioned_like(grid: GridBatch, tensor: JaggedTensor, name: str) -> None:
+        if tensor.num_tensors != grid.grid_count:
+            raise ValueError(f"{name} has {tensor.num_tensors} tensors but grid has {grid.grid_count} grids")
```
harrism, line 329: Nested `data` (`ldim == 2`) slips through: `num_tensors` is the leaf count and `joffsets` the flat leaf boundaries, so neither sees the outer list structure, and `rmask` propagates `mListIdx`/`mNumOuterLists` unchanged. Concretely: `grid = GridBatch.from_ijk(JaggedTensor([ijk, ijk]))` (two grids, two voxels each), `data = JaggedTensor([[t0, t1]])` (one outer list holding both), `mask = grid.jagged_like(torch.ones(4, dtype=torch.bool))`. Offsets are `[0, 2, 4]` on both sides, every check passes, and the result has `pruned_grid.grid_count == 2` but `len(pruned_data) == 1`, with `pruned_data[0]` holding both grids' rows. The mask side is already rejected in C++ (`BuildPrunedGrid.cu` requires `ldim() == 1`); the data side is not.

```suggestion
    def _check_partitioned_like(grid: GridBatch, tensor: JaggedTensor, name: str) -> None:
        if tensor.ldim != 1:
            raise ValueError(f"{name} must be a flat list of tensors (ldim == 1), got ldim == {tensor.ldim}")
        if tensor.num_tensors != grid.grid_count:
```

plus one nested case in `test_prune_rejects_mismatched_partitioning`.

harrism (commented):
Re-review at d3deac26. Both earlier threads are addressed: `_check_partitioned_like` compares tensor count, row count and `joffsets` against the grid (int64 on the same device on both sides, so no false positives for grid-built masks or sliced views), covered by `test_prune_rejects_mismatched_partitioning`; `functools.wraps` plus the class docstring fix the rendering, pinned by `test_prune_forward_keeps_docstring`. The batched-prune experiment is reverted net-zero over every file it touched, and the branch merges clean onto main. `test_sample.py`'s fp32 reference is a strengthening, not a loosening.

One gap left, inline; a one-line guard and a test case, then this is good to go. Two non-blocking notes: `notebooks/04_shape_vae.ipynb` still carries the output from the reverted batched-prune build (`1000/1000 [01:05...]`) while the PR body now says ~90 s, so the stored cell misrepresents what ships; and the `torch.equal` on CUDA offsets returns a Python bool, so each `Prune.forward` adds two device-to-host syncs (eight per decoder step at four levels). `pruned_grid` already syncs on the mask sum, so it is marginal, but worth knowing given the launch-bound framing.

fvdb/nn/modules.py
swahtz: Confirmed and fixed in 653e0280 — I reproduced your exact case first (checks all passed, `pruned_grid.grid_count == 2` with `len(pruned_data) == 1`). `_check_partitioned_like` now requires `ldim == 1` before the count/row/offset checks, applied to both `data` and `mask` (the mask side keeps its C++ rejection in `BuildPrunedGrid.cu`, but now fails with the same Python `ValueError` as the data side instead of reaching the op). The nested case is added to `test_prune_rejects_mismatched_partitioning`; all 18 prune tests pass on CPU+CUDA, float32/64.

fvdb/nn/modules.py
swahtz: On the sync remark — verified and improved in 9c5033ff. `joffsets` are CUDA tensors on a CUDA batch, so the `torch.equal` did sync as you said. It turns out `jagged_like` and `ConvolutionPlan.execute` outputs share the grid's offsets tensor *object* (checked for same-topology convs and generative transposes), so the check now short-circuits on identity and the equal only runs for tensors built independently of the grid. The decoder hot path — features from plan outputs, masks from `jagged_like` — is back to zero extra syncs while foreign tensors still get the full comparison.

fvdb/nn/modules.py
```diff
@@ -261,6 +263,80 @@ def forward(
...
+
+    @staticmethod
+    def _check_partitioned_like(grid: GridBatch, tensor: JaggedTensor, name: str) -> None:
+        if tensor.ldim != 1:
+            raise ValueError(f"{name} must be a flat list of tensors (ldim == 1), got ldim == {tensor.ldim}")
+        if tensor.num_tensors != grid.grid_count:
+            raise ValueError(f"{name} has {tensor.num_tensors} tensors but grid has {grid.grid_count} grids")
+        if tensor.jdata.shape[0] != grid.total_voxels:
+            raise ValueError(f"{name} has {tensor.jdata.shape[0]} rows but grid has {grid.total_voxels} voxels")
+        # jagged_like and ConvolutionPlan outputs share the grid's offsets tensor, so an
```
harrism: Nit: the matmul backend returns `data.jagged_like(out)` (`fvdb/convolution_plan.py:1069-1070`), so a plan's output shares its *input's* offsets object rather than the grid's. Same object whenever the input came from the grid, which is the case that matters here, but the comment would be exact as "share the offsets tensor of the input they were built from".

harrism (approved):
Re-verified at 9c5033ff. The `ldim` guard is applied to both `data` and `mask` ahead of the count/row/offset checks, and the added nested case in `test_prune_rejects_mismatched_partitioning` fails without it (it builds the exact same-offsets nested layout and asserts on a real `prune` call). The identity short-circuit on the offsets object is a better answer to the sync note than leaving it: it holds for `jagged_like` and for `ConvolutionPlan` outputs whose input came from the grid, a false positive is impossible, and a miss only costs the `torch.equal`. All 18 prune tests pass on CPU and CUDA against this head's Python with the unchanged C++ ops.

Two non-blocking leftovers: the committed output in `notebooks/04_shape_vae.ipynb` (`1000/1000 [01:05...]`) is still from the reverted batched-prune build while the body says ~90 s, so re-execute or leave knowingly; and one comment wording, inline.

One thing found on the way that is not this PR's: `JaggedTensor.from_data_offsets_and_list_ids` accepts a permuted single-column `jlidx` without canonicalizing it, producing a flat tensor whose `lshape` disagrees with its offsets; fed to `Prune` that yields `pruned_grid.num_voxels == [0, 2]` against `pruned_data.lshape == [2, 0]`. Pre-existing constructor invariant, will file separately.

fvdb/nn/modules.py
swahtz: Right — verified against the matmul backend (grid-built input chains to the grid's offsets, a foreign input with equal-but-distinct offsets does not). Comment reworded to "share the offsets tensor of the input they were built from" in 80bff4b4.

Issue #692, opened 2026-07-16
ray_implicit_intersection returns a spurious near-origin crossing when the ray starts far from the SDF band


swahtz:
## Summary

`GridBatch.ray_implicit_intersection` / `Grid.ray_implicit_intersection` returns an
**incorrect** intersection for a ray that starts far from the SDF's active narrow band. The
reported hit collapses to a spurious point ~¼ of a voxel from the ray origin instead of the
true zero-crossing. Moving the ray origin close to the surface returns the correct result.

This silently breaks ray-marching an SDF from a distant camera/emitter — a common setup
(e.g. `mitre_example/depth_to_sdf.py`, level-set raytracing/rendering).

## Reproduction (self-contained, no data files)

Builds an eikonal SDF the same way a TSDF-fusion pipeline does (`integrate_tsdf` +
`reinitialize_sdf`) on a synthetic sloped depth image, then traces the central pixel's ray
straight down `+z` from several start heights:

```python
import math
import torch
import fvdb

device = "cuda" if torch.cuda.is_available() else "cpu"
H = W = 96
vs = 460.0            # voxel size (metres)
trunc = vs * 4.0      # TSDF truncation distance

# Synthetic depth image: surface depth ramps across the frame (terrain-like relief),
# viewed by a pinhole camera at the world origin looking down +z.
u = torch.arange(W, device=device).float()
depth = (460.0 + (u / (W - 1)) * (5500.0 - 460.0)).unsqueeze(0).repeat(H, 1)  # (H, W)
fx = fy = W / (2.0 * math.tan(math.radians(30.0)))
cx = cy = W / 2.0

proj = torch.tensor([[fx, 0, cx], [0, fy, cy], [0, 0, 1]], device=device).unsqueeze(0)
cam_to_world = torch.eye(4, device=device).unsqueeze(0)
grid = fvdb.GridBatch.from_zero_voxels(device=device, voxel_sizes=vs, origins=0.0)
tsdf0 = grid.jagged_like(torch.zeros(grid.total_voxels, device=device))
w0 = grid.jagged_like(torch.zeros(grid.total_voxels, device=device))
grid, tsdf, _ = grid.integrate_tsdf(trunc, proj, cam_to_world, tsdf0, w0, depth.unsqueeze(0))
sdf = grid.jagged_like(tsdf.jdata * trunc)
sdf = grid.reinitialize_sdf(sdf, band=max(1, round(trunc / vs)))

[...]

PR #754, opened 2026-09-01
Use fused NanoVDB `ReadAccessor::getDimAndActive` in the HDDA iterators

swahtz:
## Summary

OpenVDB PR [#2220](https://github.com/AcademySoftwareFoundation/openvdb/pull/2220) added the fused `ReadAccessor::getDimAndActive` API to NanoVDB: a single Root→Internal→Leaf descent (or accessor-cache lookup) that returns both the HDDA step size and the active state of a voxel, packed into one register. This PR bumps the NanoVDB pin to that merge commit (`9df9ec06`) and adopts the fused call in both HDDA iterators, replacing the separate `getDim` + `isActive` pair that previously cost two tree walks per iteration.

This restores the methodology originally prototyped alongside #663, which had to be reverted before merge because the fused accessor only existed in a fork at the time.

## Changes

**`HDDASegmentIterator::nextSegment`**
- One `getDimAndActive` call per iteration under the default `ActiveExact` policy, which is byte-for-byte equivalent to the separate calls. `ActiveExact` is required here because the iterator reads `active` unconditionally, including at coarse-tile levels.
- When the HDDA re-aligns to a different level, `HDDA::update` snaps `mVoxel` to the new grid, so the result is re-queried after the update to keep `active` consistent with the voxel the TimeSpan logic reads. The two-descent cost is thus only paid on (rare) level-change iterations.

**`HDDAValueIteratorImpl::nextVoxel`**
- The realign-loop `getDim` queries and the post-convergence `isActive` fuse into one `getDimAndActive` per pass; each re-query runs at the voxel `HDDA::update` just snapped to, so on loop exit the result always describes the current voxel. The force-unrolled ≤3-pass structure is unchanged.
- Policy is selected per instantiation: `HDDALeafVoxelIterator` (`LeafOnly=true`) gates on `dim == 1` before consulting `active`, which is exactly the `ActiveOnLeafOnly` contract, so it keeps `getDim`'s skip-flag fast path. `HDDAActiveValueIterator` (`LeafOnly=false`) must yield active coarse tiles, so it uses `ActiveExact`.

No callers change — `ray_implicit_intersection`, `uniform_ray_samples`, `voxels_along_rays`, and `segments_along_rays` consume the iterators through their public API.

## Benchmarks

RTX PRO 6000 Blackwell, 512×512 pinhole-camera rays, fp32, median of 20 CUDA-event-timed iterations (top/bottom-2 trimmed), NanoVDB example grids. Output checksums are bit-identical to main across all workloads — this is a pure performance change.

| Op | Dataset | main (ms) | fused (ms) | Speedup |
|---|---|---|---|---|
| `voxels_along_rays` | dragon | 4.902 | 2.911 | **1.68×** |
| `voxels_along_rays` | emu | 8.343 | 5.031 | **1.66×** |
| `voxels_along_rays` | wdas_cloud | 24.140 | 16.970 | **1.42×** |
| `voxels_along_rays` | crawler | 18.920 | 13.939 | **1.36×** |
| `segments_along_rays` | crawler | 7.544 | 5.795 | **1.30×** |
| `segments_along_rays` | emu | 3.703 | 2.900 | **1.28×** |
| `segments_along_rays` | wdas_cloud | 12.676 | 9.945 | **1.27×** |
| `segments_along_rays` | dragon | 1.634 | 1.351 | **1.21×** |
| `uniform_ray_samples` | wdas_cloud | 12.798 | 10.033 | **1.28×** |
| `uniform_ray_samples` | dragon | 1.669 | 1.385 | **1.21×** |
| `ray_implicit_intersection` | emu | 0.587 | 0.478 | **1.23×** |
| `ray_implicit_intersection` | crawler | 1.901 | 1.630 | **1.17×** |
| `ray_implicit_intersection` | dragon | 0.384 | 0.374 | **1.03×** |

`ray_implicit_intersection` gains the least because its leaf-only iterator already skips coarse tiles cheaply (and now additionally benefits from `ActiveOnLeafOnly` keeping the skip-flag fast path).

## Test plan

- `pytest tests/unit/test_ray_marching.py tests/unit/test_sample.py` — 455 passed, 4 skipped.
- `pytest tests/unit/test_basic_ops.py -k "ray or implicit or voxels_along or segments"` — 56 passed, 1 skipped, including the #692 zero-prefix regression test.
- Benchmark outputs checksum-identical to main on all 13 workloads above.

### DIFF
--- a/src/cmake/get_nanovdb.cmake
+++ b/src/cmake/get_nanovdb.cmake
@@ -4,7 +4,7 @@
 CPMAddPackage(
     NAME nanovdb
     GITHUB_REPOSITORY AcademySoftwareFoundation/openvdb
-    GIT_TAG 7946f17edb443fe46076a22ea933e52a23453c24
+    GIT_TAG 9df9ec060dd6a982073c06931c5e93cbb208393e
     SOURCE_SUBDIR nanovdb/nanovdb
     DOWNLOAD_ONLY YES
 )
--- a/src/fvdb/detail/utils/nanovdb/HDDAIterators.h
+++ b/src/fvdb/detail/utils/nanovdb/HDDAIterators.h
@@ -11,6 +11,7 @@
 #include <c10/util/Half.h>
 
 #include <iostream>
+#include <type_traits>
 
 namespace nanovdb {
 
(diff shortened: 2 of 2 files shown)


harrism (approved):
Nice speedup!

PR #756, opened 2026-09-01
Enforce agreeing toolchain pins across the conda envs

harrism:
## Problem

`env/` holds several conda environments that each pin `python`, `cuda-version` and `pytorch-gpu` independently, with nothing keeping them in agreement:

| File | python | cuda-version | pytorch-gpu |
|---|---|---|---|
| `build_environment.yml` | 3.12 | 13.0 | 2.13.0 |
| `dev_environment.yml` | 3.12 | 13.0 | 2.13.0 |
| `learn_environment.yml` | 3.12 | 13.0 | 2.13.0 |
| `test_environment.yml` | 3.12 | 13.0 | 2.13.0 |
| `release_base_environment.yml` | — | — | — |

They agree today, but that's coincidence, not a guarantee — four files have to be edited in lockstep by hand on every bump.

A disagreement wouldn't be cosmetic. `build_environment.yml` is the environment the wheel is compiled in, so it fixes the ABI: someone working from a different env file with a different PyTorch builds or loads a wheel against a mismatched libtorch and gets an undefined-symbol `ImportError` at `import fvdb` — the same failure that broke fvdb-reality-capture's nightly three times across repos. fvdb-reality-capture also derives its benchmark environment from `build_environment.yml` ([frc#319](https://github.com/openvdb/fvdb-reality-capture/pull/319)), so a disagreement here leaves *"which PyTorch does fvdb-core use?"* without a single answer.

## Change

A codestyle job enforcing:

> If two environment files pin the same key, they must pin the same value.

The rule is **permissive about presence and strict about agreement** — a file may omit a key (`release_base_environment.yml` pins none), so adding a slimmer environment later doesn't require touching the script.

`--fix` propagates `build_environment.yml`'s values to the others, keeping a version bump to a one-file edit plus one command:

```
python3 .github/scripts/check_env_pin_consistency.py env         # check
python3 .github/scripts/check_env_pin_consistency.py env --fix   # propagate
```

Follows the existing `check_runner_token_policy.py` precedent: policy script in `.github/scripts/`, companion pytest file, both run from the workflow.

## Why this shape

This replaces [#746](https://github.com/openvdb/fvdb-core/pull/746) (closed), which tried to gate fvdb-core PRs against fvdb-reality-capture's pins. That was cross-repo and awkward — it needed a label escape hatch and a companion-PR search precisely because the fix lived in another repository and couldn't merge in the same commit.

This check is **entirely in-repo, so it cannot deadlock**: whatever breaks it is fixed in the same PR. No token, no label, no cross-repo API call.

## Testing

Nine unit tests cover agreement, omitted keys, disagreement on each of the three keys, surgical `--fix` (unrelated content preserved), and both error paths. Verified against the real `env/` directory: passes as-is, exits 1 with a per-key breakdown when a pin is broken, and `--fix` restores it.

### DIFF
--- a/.github/scripts/check_env_pin_consistency.py
+++ b/.github/scripts/check_env_pin_consistency.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python3
+# Copyright Contributors to the OpenVDB Project
+# SPDX-License-Identifier: Apache-2.0
+"""Repo-specific CI policy for toolchain pins shared across the conda envs.
+
+``env/`` holds several conda environments (build, dev, test, learn, ...) that
+each pin the toolchain independently. Nothing has kept them in agreement, yet a
+disagreement is not cosmetic:
+
+  * ``build_environment.yml`` is the environment the wheel is compiled in, so it
+    determines the resulting ABI.
+  * A developer working from ``dev_environment.yml`` with a different PyTorch
+    would build or load a wheel against a mismatched libtorch and hit an
+    undefined-symbol ImportError.
+  * fvdb-reality-capture derives its benchmark environment from
+    ``build_environment.yml``; if the env files here disagree, "which PyTorch
+    does fvdb-core use?" has no single answer.
+
+The rule enforced here is deliberately permissive about *presence* and strict
+about *agreement*:
+
+    If two environment files pin the same key, they must pin the same value.
+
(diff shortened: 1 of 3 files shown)


swahtz (approved):
lgtm

PR #757, opened 2026-09-01
Batched leaf-mask topology construction for generated conv grids

### LINKED ISSUE #755: Generated-topology grid construction dominates per-iteration ConvolutionPlan cost (serial per-grid build + merge)
## Summary

When training generative sparse networks (topology changes every iteration, so `ConvolutionPlan`s must be rebuilt per step), plan construction dominates wall-clock and leaves the GPU underutilized (~35% GPU, one CPU core pegged in the shape-VAE example from #753). Profiling shows the cost is **not** the kernel-map build — it is the construction of the *generated topology grid* itself.

## Measurements

Shape-VAE example (#753) on the 254-shoe GSO dataset, batch 16 (~110k voxels), resolution 64, 4-level encoder/decoder:

- One training iteration: 165 ms, of which **69.6 ms is ConvolutionPlan construction** (21 plans) and 22 ms per-minibatch `conv_grid` pyramid rebuild.
- `cProfile` of a single `ConvolutionPlan.from_grid_batch_transposed(2, 2, grid_batch16)`: **24 of 26 ms is `_fvdb_cpp.conv_transpose_grid`** (generated target-grid construction). The transpose kernel map (`gs_build_transpose_topology`) is only 0.5 ms; a k3 s1 same-topology kernel map (`gs_build_topology`) is 0.8 ms.
- Plan-construction time scales ~linearly with batch voxel count (65 / 126 / 182 ms per iteration at batch 16 / 32 / 48), so batching the training step does not amortize it.

## Root cause

`dispatchBuildGridForConvTranspose` (and the analogous `conv_grid` path) constructs the generated grid **one batch member at a time** and then merges:

- `src/fvdb/detail/ops/BuildGridForConvTranspose.cu`: even on the fast leaf-mask subdivision path (K=S=2, which this workload hits), `perItemGridHandle` loops over the batch building one NanoVDB grid handle per member, each with its own stream work, then `nanovdb::cuda::mergeGridHandles`. At batch 16 that is ~1.5 ms of fixed per-grid overhead x 16, serialized.

So per-iteration cost is dominated by per-grid fixed overhead in generated-topology construction, not by topology size or kernel-map work.

## Suggested direction

Batched generated-grid construction: build the strided/subdivided support for all batch members in one pass (single staging + single grid-batch build) instead of per-member handle construction + merge. This would directly speed up:

- `GridBatch.conv_grid` / `conv_transpose_grid`
- `ConvolutionPlan.from_grid_batch(_transposed)` with generated targets (`target_grid=None`)
- any per-iteration generative training loop (the #741 / #753 use case; MinkowskiEngine and spconv build their equivalents fully on-device, which is why they sustain higher GPU utilization on the same workload)

Secondary (much smaller): `from_grid_batch(1, 1, g, g)` already short-circuits execution to `_MatmulBackend`, but construction still pays ~1.2 ms of Python-side validation/coverage setup per call; a fast path for the identity-plan case would tidy up per-iteration classifier heads.

## Repro

```python
import torch, cProfile
from fvdb import ConvolutionPlan
from fvdb.utils.examples import load_gso_shoes
import fvdb

meshes = load_gso_shoes(limit=16)
v = fvdb.JaggedTensor([(m[0] - m[0].amin(0)) / m[0].amax() * 0.96 + 0.02 for m in meshes])
f = fvdb.JaggedTensor([m[1].int() for m in meshes])
g = fvdb.GridBatch.from_mesh(v, f, voxel_sizes=1/64, origins=0.0)
cProfile.run('ConvolutionPlan.from_grid_batch_transposed(2, 2, g)', sort='cumulative')
# -> ~24ms in _fvdb_cpp.conv_transpose_grid, ~0.5ms in gs_build_transpose_topology
```

swahtz:
# Batched leaf-mask topology construction for generated conv grids

Fixes #755.

## Summary

`conv_grid` / `conv_transpose_grid` with generated targets (and `refined_grid` / `coarsened_grid` / the stride-1 dilate-pad paths generally) built their output **one batch member at a time**: each member paid a full NanoVDB `RefineGrid`/`CoarsenGrid`/`DilateGrid`/`PadGrid` build (3+ stream synchronizations each, a host-side speculative root refinement with a pageable D2H readback, a `GridHandle` constructor with 2 blocking memcpys + raw `cudaMalloc`/`cudaFree`), plus a host-built proxy grid + H2D per **empty** member, and finally `nanovdb::cuda::mergeGridHandles` added another synchronization **per grid**. That is ~1.5 ms of fixed overhead x B members x 21 plans per training iteration in generative workloads where topology changes every step (#741 / #753) — construction cost scaled linearly in batch size and left the GPU at ~35% utilization.

This PR batches the leaf-mask topology machinery across the whole grid batch (extending #712's leaf-mask direction — no return to coordinate staging), in four self-contained commits:

### 1. Batched topology builder + refine/coarsen passes

New `src/fvdb/detail/utils/nanovdb/BatchedTopologyBuilder.cuh`: one batched pass builds ALL grids at once. Per pass:

- one emission kernel over all source leaves of all members produces candidate output leaves as (root-tile sort key, in-tile node key, origin, 512-bit mask) slots, segmented per grid — mask bit math is NanoVDB's own (`RefineLeafMasksFunctor::refineMask`, `CoarsenLeafMasksFunctor::coarsenMask`);
- two stable segmented radix sorts put each grid's slots in **canonical NanoVDB node order** (root tiles by the PointsToGrid offset-shifted key, then x-major upper/lower child offsets), so produced grids enumerate voxels identically to every other build path (pinned by elementwise `torch.equal` tests);
- head-flag + scan passes dedup nodes and derive per-grid counts and parent linkage; root tiles are derived **on device** from the unique upper keys (the host-side speculative `refineRoot` and its readback disappear);
- **one** `cudaStreamSynchronize` reads back per-grid node counts, one buffer is allocated, and batched kernels (transcribed from `tools::cuda::TopologyBuilder`'s functors with `(gridIndex, localIndex)` indexing) write every grid's headers (`mGridIndex=g, mGridCount=B`), nodes, leaf `mOffset`/`mPrefixSum`, and bboxes. Empty members become valid empty grids inline.

`fineGridHandleFromCoarseCUDA` / `coarseGridHandleFromFineCUDA` route through it for all batch sizes; multi-pass factors (4, 8) chain passes, mirroring the previous per-pass semantics. Checksums are disabled on the output (matching `ops::contiguousGridHandle` and `mergeGridHandles`). CPU / PrivateUse1 paths and non-power-of-two coordinate fallbacks unchanged; masked subdivision still prunes upstream, then batched-refines.

### 2. Batched box-dilate passes (stride-1 K>1, k3s2)

A `BoxDilate` pass computes the Minkowski sum with an axis-aligned unit box via pure per-axis bit shifts (scatter formulation, up to 27 target leaves per source leaf, deduplicated by the same back-end). This covers `DilateGrid`'s 26-neighbor dilation (`[-1,1]^3`) and both `PadGrid` octants (`{-1,0}^3`, `{0,1}^3`), so the stride-1 uniform-K conv/conv-transpose paths and the k3s2 transpose (refine + negative pad) become batched pass sequences. The per-member `perItemGridHandle` drivers are deleted from both conv builders.

### 3. Identity-plan fast path (issue's secondary item)

`ConvolutionPlan.from_grid_batch[_transposed](1, 1, g[, g])` short-circuits to the matmul backend without building tensor-valued transform diagnostics (which are trivially exact when source and target share `GridBatchData`), preserving channel-pair validation, backend-name rejection, and the general path for distinct-but-equal-looking grids: ~1.2 ms -> **0.08 ms** per call.

### 4. Grid-construction cheap wins

- `makeGridBatchData`: `leafBatchIndices` via one `repeat_interleave` instead of B x `torch::full` + `torch::cat` (B+1 dispatches on every grid construction).
- `voxelSizesTensor` / `voxelOriginsTensor`: accessor fill instead of per-element ATen indexing (6 dispatches per grid per call, on every plan construction's transform validation).

## Measurements

RTX PRO 6000 Blackwell, synthetic shell batches (~7.8k voxels/member, resolution 64), median of 20 CUDA-event-timed iterations (`src/benchmarks/convolution/benchmark_conv_grid_build.py`):

| op | before B=16 | after B=16 | before B=48 | after B=48 |
|---|---|---|---|---|
| `conv_transpose_grid` k2s2 | 9.0 ms | **0.80 ms** | 25.6 ms | **0.94 ms** |
| `conv_grid` k2s2 | 8.7 ms | **0.80 ms** | 23.4 ms | **0.94 ms** |
| `conv_grid` k3s1 | 9.8 ms | **0.83 ms** | 25.1 ms | **1.00 ms** |
| plan `from_grid_batch(2,2,g)` | 8.9 ms | **1.27 ms** | 24.8 ms | **1.62 ms** |
| plan `from_grid_batch_transposed(2,2,g)` | 10.2 ms | **1.94 ms** | 29.4 ms | **3.27 ms** |
| 4-level plan-pyramid rebuild (8 plans + 3 conv_grids) | 114 ms | **17.4 ms** | 309 ms | **27.0 ms** |
| identity plan `from_grid_batch(1,1,g,g)` | ~1.2 ms | **0.08 ms** | — | — |

Construction cost is now near-flat in batch size (B=1: within noise of the old single-grid path). The issue's per-iteration plan-construction share (~70 ms of a 165 ms shape-VAE iteration at B=16) drops to single-digit milliseconds.

Follow-ups (out of scope, same back-end): an ijk emission front-end to resolve `BuildGridFromIjk.cu`'s per-member FIXME (`from_ijk`/`from_points`/shifted-geometry fallbacks), and `dilated_grid`/`BuildPaddedGrid`'s standalone per-member loops.

## Test plan

- New `tests/unit/test_batched_topology_builder.py` (18 tests): elementwise (`torch.equal` on `ijk.jdata`, `num_voxels`, per-member bboxes) equivalence against `from_ijk`-built expected topologies — pinning canonical node order — plus per-member coordinate-set equality against the CPU paths, across: mixed member sizes, empty members (first/middle/last/all), coordinates straddling +-4096 root-tile boundaries and negative octants (where the sort-key and stored `Tile::key` encodings order differently), single-grid and 16-grid batches, factors 2 and 4, masked refine, `conv_grid`/`conv_transpose_grid` K in {2,3,4,5} at stride 1, k3s2 transpose, k2s2 vs per-member, and a refine->coarsen round trip.
- Existing suites, all green (724 passed, 1 skipped): `pytest unit/test_conv_semantics_integration.py unit/test_conv_default.py unit/test_conv_transpose_default.py unit/test_batching.py unit/test_basic_ops.py unit/test_sliced_batch.py unit/test_conv_semantics.py unit/test_conv_ground_truth.py unit/test_nn_modules.py` — includes the elementwise ijk order pins, sliced-view coverage, resource-stats path pinning, and the matmul/identity plan contract tests.
- Benchmark: `python src/benchmarks/convolution/benchmark_conv_grid_build.py` (numbers above; `--gso` runs the issue's verbatim GSO repro).

### DIFF
--- a/fvdb/convolution_plan.py
+++ b/fvdb/convolution_plan.py
@@ -716,6 +716,19 @@ class ConvolutionPlan:
         if backend_name == "pred_gather_igemm":
             _validate_pred_gather_igemm_admission(kernel_size, stride, channel_pairs, transposed=False)
             _validate_pred_gather_igemm_grid_admission(source_grid, target_grid)
+        elif backend_name in ("default", "gather_scatter"):
+            identity_plan = cls._identity_matmul_plan(
+                kernel_size,
+                stride,
+                source_grid,
+                target_grid,
+                channel_pairs,
+                resolved_policy,
+                topology_provenance,
+                transposed=False,
+            )
+            if identity_plan is not None:
+                return identity_plan
         if target_grid is None:
             target_grid = source_grid.conv_grid(kernel_size, stride)
 
@@ -808,6 +821,19 @@ class ConvolutionPlan:
         if backend_name == "pred_gather_igemm":
             _validate_pred_gather_igemm_admission(kernel_size, stride, channel_pairs, transposed=True)
             _validate_pred_gather_igemm_grid_admission(source_grid, target_grid)
+        elif backend_name in ("default", "gather_scatter"):
+            identity_plan = cls._identity_matmul_plan(
+                kernel_size,
+                stride,
+                source_grid,
+                target_grid,
+                channel_pairs,
+                resolved_policy,
+                topology_provenance,
+                transposed=True,
+            )
+            if identity_plan is not None:
+                return identity_plan
         if target_grid is None:
             target_grid = source_grid.conv_transpose_grid(kernel_size, stride)
 
@@ -1163,6 +1189,66 @@ class ConvolutionPlan:
     # ============================================================
     #                 Private methods
     # ============================================================
+
+    @classmethod
+    def _identity_matmul_plan(
+        cls,
+        kernel_size: torch.Tensor,
+        stride: torch.Tensor,
+        source_grid: GridBatch,
+        target_grid: GridBatch | None,
+        channel_pairs: tuple[tuple[int, int], ...],
+        resolved_policy: ConvolutionTopologyPolicy,
+        topology_provenance: ConvolutionTopologyProvenance,
+        transposed: bool,
+    ) -> "ConvolutionPlan | None":
+        """Identity (K == S == 1, shared grid data) short circuit to the matmul backend.
+
+        Per-iteration classifier heads rebuild identity plans constantly (issue #755); the general
+        path spends ~1 ms per call building tensor-valued transform diagnostics that are trivially
+        satisfied when source and target share their GridBatchData. Returns None when the fast
+        path does not apply (the caller then follows the general path, keeping full validation
+        for distinct-but-equal-looking grids and incompatible transforms).
+        """
+        if kernel_size.tolist() != [1, 1, 1] or stride.tolist() != [1, 1, 1]:
+            return None
+        if target_grid is None:
+            # conv_grid / conv_transpose_grid are the identity at K == S == 1: the generated
+            # target is the source grid itself, preserving public and data identity.
+            target_grid = source_grid
+        elif not _get_grid_data(source_grid).is_same(_get_grid_data(target_grid)):
+            return None
+        # Preserve the general path's construction-time channel-pair validation.
+        for channel_pair in channel_pairs:
+            if len(channel_pair) != 2 or channel_pair[0] <= 0 or channel_pair[1] <= 0:
+                raise ValueError("channel_pair must be a tuple of two positive integers")
+        geometry = _fvdb_cpp.ConvolutionGeometry(kernel_size, stride)
+        # Shared grid data makes every registration diagnostic exact by construction.
+        compatibility = ConvolutionTransformCompatibility(
+            fine_grid_count=source_grid.grid_count,
+            coarse_grid_count=source_grid.grid_count,
+            same_batch_size=True,
+            same_device=True,
+            scale_compatible=True,
+            registration_integer=True,
+            registration_zero=True,
+            compatible=True,
+            registration_offset=torch.zeros((source_grid.grid_count, 3), dtype=torch.float64),
+        )
+        backend = _MatmulBackend()
+        return cls(
+            source_grid,
+            target_grid,
+            geometry,
+            channel_pairs,
+            transposed,
+            backend,
+            compatibility,
+            resolved_policy,
+            topology_provenance,
+            _CoverageReportCache(backend, source_grid, target_grid),
+            False,
+        )
 
     @staticmethod
     def _build_backend(
--- a/fvdb/convolution_plan.py
+++ b/fvdb/convolution_plan.py
@@ -1223,6 +1223,16 @@ class ConvolutionPlan:
             target_grid = source_grid
         elif not _get_grid_data(source_grid).is_same(_get_grid_data(target_grid)):
             return None
+        # Shared grid data makes the registration diagnostic exact only for a finite transform.
+        # The general path rejects non-finite origins and zero voxel sizes through a non-finite
+        # registration offset, so the fast path must reject them too.
+        grid_data = _get_grid_data(source_grid)
+        if not bool(
+            torch.isfinite(grid_data.origins).all()
+            and torch.isfinite(grid_data.voxel_sizes).all()
+            and (grid_data.voxel_sizes != 0).all()
+        ):
+            raise ValueError("Convolution grid transform must have finite origins and finite, nonzero voxel sizes.")
         # Preserve the general path's construction-time channel-pair validation.
         for channel_pair in channel_pairs:
             if len(channel_pair) != 2 or channel_pair[0] <= 0 or channel_pair[1] <= 0:
--- a/fvdb/convolution_plan.py
+++ b/fvdb/convolution_plan.py
@@ -1209,6 +1209,11 @@ class ConvolutionPlan:
         satisfied when source and target share their GridBatchData. Returns None when the fast
         path does not apply (the caller then follows the general path, keeping full validation
         for distinct-but-equal-looking grids and incompatible transforms).
+
+        Applies to ``backend="default"`` and ``backend="gather_scatter"`` alike. An explicit
+        ``gather_scatter`` request at the identity is honoured as matmul, matching the contract
+        ``_build_backend`` has always applied on the general path. With no reduction over kernel
+        taps the two backends compute the same per-voxel matmul, so the result is identical.
         """
         if kernel_size.tolist() != [1, 1, 1] or stride.tolist() != [1, 1, 1]:
             return None
--- a/src/fvdb/detail/utils/nanovdb/BatchedTopologyBuilder.cuh
+++ b/src/fvdb/detail/utils/nanovdb/BatchedTopologyBuilder.cuh
@@ -1,0 +1,1208 @@
+// Copyright Contributors to the OpenVDB Project
+// SPDX-License-Identifier: Apache-2.0
+//
+// Batched leaf-mask topology construction: builds ALL grids of a batch in one pass.
+//
+// NanoVDB's morphology builders (RefineGrid / CoarsenGrid via TopologyBuilder) are single-grid by
+// construction: each getHandle() call performs several stream synchronizations (source-tree
+// readback, host-side speculative root refinement, node-count readback for allocation sizing) and
+// produces one single-grid handle, which fvdb then merges per batch member -- with another
+// synchronization per grid inside nanovdb::cuda::mergeGridHandles. For generated-topology
+// workloads that rebuild grids every training iteration (issue #755), that per-member fixed
+// overhead -- not the topology size -- dominates wall clock and serializes the GPU.
+//
+// This header rebuilds the factor-2 refine (subdivision) and coarsen passes so one invocation
+// covers the whole batch:
+//
+//   emit    one kernel over all source leaves of all members emits candidate output leaves as
+//           (root-tile sort key, in-tile node key, leaf origin, 512-bit activity mask) slots,
+//           segmented per grid;
+//   sort    two stable segmented radix sorts put each grid's slots in canonical NanoVDB node
+//           order (root tiles by the PointsToGrid offset-shifted key, then upper/lower child
+//           offsets); invalid slots sort to the segment tails;
+//   dedup   head-flag + scan passes derive the unique leaf/lower/upper nodes, their per-grid
+//           counts, and parent linkage (coarsen additionally OR-combines duplicate leaf masks);
+//   size    ONE host synchronization reads back the per-grid node counts; per-grid byte offsets
+//           are computed on the host and a single output buffer is allocated;
+//   build   batched kernels write every grid's GridData/TreeData/RootData (mGridIndex = g,
+//           mGridCount = B), root tiles, upper/lower/leaf nodes, leaf mOffset/mPrefixSum, and
+//           bounding boxes. Empty members become valid empty grids inline (no host proxy grids).
+//
+// The mask bit math is NanoVDB's own (RefineLeafMasksFunctor::refineMask /
+// CoarsenLeafMasksFunctor::coarsenMask); the header/bbox/prefix-sum stages are transcriptions of
+// tools::cuda::TopologyBuilder's functors with (gridIndex, localIndex) indexing. Checksums are
+// disabled on the output, matching ops::contiguousGridHandle and mergeGridHandles behavior.
+//
+// Scratch is allocated through torch (the caching allocator) on the caller's current stream. The
+// only stream synchronization per pass is the node-count readback that sizes the output buffer.
+
+#ifndef FVDB_DETAIL_UTILS_NANOVDB_BATCHEDTOPOLOGYBUILDER_CUH
+#define FVDB_DETAIL_UTILS_NANOVDB_BATCHEDTOPOLOGYBUILDER_CUH
+
+#include <fvdb/GridBatchData.h>
+#include <fvdb/TorchDeviceBuffer.h>
+
+#include <nanovdb/NanoVDB.h>
+#include <nanovdb/util/MorphologyHelpers.h>
+#include <nanovdb/util/cuda/Morphology.cuh>
+
+#include <c10/cuda/CUDAException.h>
+#include <c10/cuda/CUDAGuard.h>
+#include <torch/types.h>
+
+#include <cub/cub.cuh>
+
+#include <algorithm>
+#include <cstdint>
+#include <limits>
+#include <vector>
+
+namespace fvdb {
+namespace detail {
+namespace batched {
+
+using BuildT = nanovdb::ValueOnIndex;
+using GridT  = nanovdb::NanoGrid<BuildT>;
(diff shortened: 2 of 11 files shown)


src/fvdb/detail/utils/nanovdb/BatchedTopologyBuilder.cuh
```diff
@@ -0,0 +1,1208 @@
...
+            lowerParent[lowerIdx]   = upperRank[j] - 1;
+        }
+        if (upperFlag[j]) {
+            upperHeadSlot[upperRank[j] - 1] = uint32_t(j);
+        }
+    }
+}
+
+/// Coarsen only: OR every duplicate slot's mask contribution into its unique leaf's head slot.
+static __global__ void
```
harrism, line 563: > _Analysis by Claude Code, run by @harrism._

**The docstring says "Coarsen only", but the call site runs this for dilate too.**

At `:1049` the guard is:

```cpp
if (pass.op != TopologyPassSpec::Op::Refine) { // coarsen/dilate have duplicate producers
```

which is correct — box-dilation genuinely produces duplicate output leaves when neighbouring source leaves dilate into the same target, so their mask bits must be OR-ed together exactly as coarsening's are. Refine is the only op with an injective leaf mapping, so excluding just Refine is the right condition.

The risk is the comment, not the code. Someone reading `/// Coarsen only:` and tidying the guard to `== Op::Coarsen` would silently drop dilate's duplicate contributions, and the symptom would be a handful of missing voxels at leaf boundaries in stride-1 convolutions — no crash, no assertion, and not something the elementwise tests would obviously localize.

Suggest `/// Coarsen and dilate: OR every duplicate slot's mask contribution into its unique leaf's head slot. (Refine is excluded because its leaf mapping is injective.)`

src/fvdb/detail/utils/nanovdb/BatchedTopologyBuilder.cuh
```diff
@@ -0,0 +1,1208 @@
...
+        // must be masked out for the occupancy test.
+        nanovdb::Mask<3> fineMask;
+        uint64_t *w       = fineMask.words();
+        uint64_t occupied = 0;
+        for (int i = 0; i < 4; ++i) {
+            w[i] = srcWords[i + bi * 4] >> (4 * bk + 32 * bj);
+            occupied |= w[i] & 0x000000000f0f0f0fUL;
+        }
+        if (!occupied) {
+            em.tileKey[s] = kInvalidTileKey;
```
harrism, line 264: > _Analysis by Claude Code, run by @harrism._

**Dead slots leave `nodeKey`, `origin` and `mask` uninitialized, and both the sort and the gather read them.**

This early-out writes only `tileKey` before `continue`ing, but the emission arrays come from `torch::empty`, so the other three stay whatever was in device memory. Two later reads do not filter on `tileKey` first:

* sort pass 1 (`:951`) sorts **all** `numSlots` entries by `em.nodeKey`, dead ones included;
* `gatherSorted` (`:448`) unconditionally reads `nodeKey`, all three `origin` components and all 8 `mask` words for every slot.

The output is not affected — dead slots carry `kInvalidTileKey`, sort to the end of their segment, and are skipped by `computeHeadFlags` and `combineDuplicateMasks` — so this is not a correctness bug. But it is a genuine uninitialized read: `compute-sanitizer --tool initcheck` will flag it, which makes it noise in front of any real finding later, and it is formally UB.

**One line per emission kernel fixes the sort half.** In all three kernels the early-out is the same two lines; add the `nodeKey` write:

```cpp
        if (!occupied) {
            em.tileKey[s] = kInvalidTileKey;
            em.nodeKey[s] = 0;   // sort pass 1 reads every slot's node key
            continue;
        }
```

Same edit at `:283` (`emitCoarsenedLeaves`, the `srcLeaf.valueMask().isOff()` guard) and `:401` (`emitBoxDilatedLeaves`).

That still leaves `gatherSorted` reading uninitialized `origin` and `mask`. Rather than zeroing 11 more words in each kernel, the cheaper fix is one guard in `gatherSorted` — skip the payload copy when the slot is dead, since nothing downstream reads it:

```cpp
const uint32_t s = perm[j];
out.nodeKey[j]   = em.nodeKey[s];
if (em.tileKey[s] == kInvalidTileKey) { continue; }   // payload is never read for dead slots
```

(`out.tileKey` is already the sorted key array, so it does not need writing here.)

tests/unit/test_batched_topology_builder.py
```diff
@@ -0,0 +1,218 @@
...
+            self.assertTrue(
+                torch.equal(conv.ijk.unbind()[b], conv_single.ijk.jdata),
+                f"conv_grid k2s2 member {b}",
+            )
+            self.assertTrue(
+                torch.equal(convt.ijk.unbind()[b], convt_single.ijk.jdata),
+                f"conv_transpose_grid k2s2 member {b}",
+            )
+
+    @parameterized.expand([("mixed_sizes",), ("empty_members",), ("tile_boundaries",)])
```
harrism: > _Analysis by Claude Code, run by @harrism._

**The box-dilate path is the only one never tested with an all-empty batch.**

`test_refined_grid_matches_expected_and_cpu` and `test_coarsened_grid_matches_expected_and_cpu` both parameterize over `_tricky_batches().keys()`, which picks up `all_empty` automatically. This one hardcodes three names and omits it — so `all_empty` is defined at `:75` and reaches the refine and coarsen paths but never the stride-1 / k3s2 route.

That inverts the risk ordering: `emitBoxDilatedLeaves` is the newest and most intricate of the three emission kernels (the `shiftWordY` / `shiftWordZ` word arithmetic and the one-sided even-K passes), and it is the one whose zero-slot behaviour goes unexercised. The machinery looks correct — `numBlocks` clamps to 1 and the `numSlots > 0` guards are in place — which is exactly why a test is worth having: it pins behaviour that currently holds by construction.

The fixture data already exists, so this is:

```python
@parameterized.expand([(name,) for name in _tricky_batches().keys()])
```

matching the two tests above.

harrism (commented):

## Review: batched leaf-mask topology construction

Reviewed the builder directly rather than through an automated pass. **No correctness defects found in the build path.** Three things below, none blocking.

Two premises I checked rather than assumed, since they carry the whole design and are the parts a reviewer would otherwise take on trust:

**The canonical-order claim holds, verified by execution.** Everything depends on `tileSortKey` reproducing the root-tile order other CUDA paths already produce. I built grids through fvdb on an RTX PRO 6000 Blackwell and compared the actual `ijk` order against the order the PR's two keys predict, for origin-only coordinates, negative octants, `+-4096` straddling, and far-negative coordinates. Exact match in every case — including the one the PR itself calls out, where the sort key and the stored `Tile::key` order differently.

**The bit packing is safe.** `tileSortKey` produces 20-bit fields in 21-bit slots, so no component can bleed into the next even at `INT32_MIN`/`INT32_MAX`; the `+2^31` offset makes it unsigned-monotone across negatives; and `kInvalidTileKey = ~0` genuinely sorts after every representable valid key. `nodeSortKey` packs a 15-bit upper offset above a 12-bit lower offset with no overlap.

Also checked and found correct: the two-pass LSD radix sort (stable by 27-bit node key, then 64-bit tile key) gives the right lexicographic order; `computeHeadFlags` short-circuits before any `[j-1]` access so there is no underflow at `j == 0` and no cross-segment comparison, and because invalid slots sort last a valid slot can never follow an invalid one inside a segment; `combineDuplicateMasks`'s `leafHeadSlot[leafRank[j] - 1]` indexing resolves to the right head; `segmentNodeOffsets` bounds are right for `numGrids + 1` threads; and `numBlocks` clamping to 1 means an all-empty batch cannot produce a zero-block launch.

The measurements and the four-commit split make this easy to follow — the separation between the builder, the routing, the identity short-circuit, and the metadata wins is worth keeping in the merge.

fvdb/convolution_plan.py
```diff
@@ -716,6 +716,19 @@ def from_grid_batch(
         if backend_name == "pred_gather_igemm":
             _validate_pred_gather_igemm_admission(kernel_size, stride, channel_pairs, transposed=False)
             _validate_pred_gather_igemm_grid_admission(source_grid, target_grid)
+        elif backend_name in ("default", "gather_scatter"):
```
harrism, line 719: > _Analysis by Claude Code, run by @harrism._

**Question rather than a defect: this also overrides an *explicit* `backend="gather_scatter"` request.**

For `"default"` the substitution is unambiguously right — the caller has expressed no preference, and at K == S == 1 with shared grid data a matmul plan is the correct choice.

The part worth confirming is that `"gather_scatter"` is in the same branch. `expert_config["backend"] = "gather_scatter"` is an explicit, deliberate request, and it now silently yields a `_MatmulBackend`. Note the two neighbouring cases are *not* treated this way: `"dense"` raises, and `"pred_gather_igemm"` goes through its own admission checks and keeps the backend the caller asked for. So `gather_scatter` is the one explicitly-named backend that can come back as something else.

Whether that matters depends on what the knob is for. The result should be numerically identical — at K == 1 there is no reduction over kernel taps, so both paths are the same per-voxel matmul — so this is not a correctness concern. But `expert_config` is the escape hatch people reach for when they are benchmarking backends against each other or working around a backend-specific problem, and both of those uses break quietly: a benchmark comparing `gather_scatter` to `matmul` at K == 1 would measure matmul twice and report them identical, with nothing in the plan to indicate why. Downstream code branches on the choice too (`isinstance(plan._backend, _MatmulBackend)` at `:917`).

Three ways to resolve it, in increasing strictness — any is fine, including the first:

1. Keep the behaviour and say so in the docstring: *"applies to `default` and `gather_scatter`; an explicit `gather_scatter` request is honoured as matmul at the identity, which is numerically equivalent."*
2. Restrict the short circuit to `backend_name == "default"`, so an explicit request is always honoured and only the unspecified case is optimized.
3. Keep it, but record the original request on the plan so a caller or a benchmark can tell the substitution happened.

Mainly flagging because the fast path is invisible at the call site: a caller sees `from_grid_batch(1, 1, g, g)` return quickly and has no signal that their backend choice was reinterpreted.

src/fvdb/detail/utils/nanovdb/BatchedTopologyBuilder.cuh
swahtz: Fixed in c539a87e. The docstring now names coarsen and box-dilate and says why refine is excluded.

src/fvdb/detail/utils/nanovdb/BatchedTopologyBuilder.cuh
swahtz: Fixed in c539a87e, both halves as you suggested. Dead slots write a zero node key in all three emission kernels and gatherSorted skips the payload copy for them. Ran compute-sanitizer initcheck over refine, coarsen, and box-dilate on a batch with an empty member. No uninitialized reads are attributed to the builder's kernels. The remaining reports are the GridHandle constructor's host readback of the grid header, and they also appear on the from_ijk PointsToGrid path, so they predate this PR.

tests/unit/test_batched_topology_builder.py
swahtz: Fixed in c539a87e. The test now parameterizes over all tricky batches, which also brings in single_grid and larger_batch. 21 tests pass.

fvdb/convolution_plan.py
swahtz: Good catch on the visibility. The substitution itself predates this PR though. The general path in `_build_backend` has returned `_MatmulBackend` for `gather_scatter` at K == S == 1 with shared grid data since #726, so the fast path only skips the diagnostics and returns the same backend the slow path would. Went with your option 1 and documented it in the `_identity_matmul_plan` docstring in c539a87e.

harrism (approved):
Nice optimization! We should consider what parts of this should move up to nanovdb next.

PR #762, opened 2026-09-08
Fix reinitialize_sdf inactive-interior bug; rename retopologize_sdf to rebuild_narrow_band

swahtz:
Fixes a correctness bug in `reinitialize_sdf` / `retopologize_sdf` (introduced in #669) for any input whose interior is inactive, documents the op's actual input contract, rejects anisotropic voxels instead of silently mis-scaling, and **renames `retopologize_sdf` → `rebuild_narrow_band`** (breaking, no alias).

## Bug

An `IndexGrid` has a single background slot (value-index 0). The eikonal solver filled it with `+band*vx`, so **every inactive neighbour read as "outside"**. That is correct for a filled solid (the occupancy → SDF pipeline the op was written for), but wrong for a narrow band whose interior is inactive — a pruned level set, an imported OpenVDB grid, or **the op's own narrow-band output**. The innermost active voxels saw `+band*vx` across the face, `|∇φ|` blew up to ~`2·band`, and a phantom interface grew inward from the band edge.

The zero crossing survives (the frozen Peng sign pins it), so meshes looked right and this went unnoticed. But the inner half of the band was wrong and the rebuild was not idempotent.

Measured on a sphere (R = 12 vx, band = 3), before the fix:

| input | inner-half-band mean \|err\| | min(φ) | sign flips (\|truth\| > 0.5 vx) |
|---|---|---|---|
| solid (interior active) | 0.076 vx | −2.92 | 0 |
| narrow band (interior inactive) | **3.579 vx** | **+0.29** (truth −2.94) | **2120** |
| rebuild applied to its own output | 0.767 vx | **−0.95** (was −2.99) | |

## Fix

**`ReinitializeSdf.cu`** — resolve inactive neighbours per read instead of via slot 0: `faceValue()` returns `copysign(band*vx, φ_center)`, so inactive space continues the sign of the adjacent active voxel (−B beside interior, +B beside exterior). All three stencil kernels (sign, Godunov, smoothing) go through it. No-op for filled solids, whose inactive neighbours only ever touch positive voxels.

**`_sdf.py`** — `pad=True` had the same flaw one level up: it injected a constant `+band*vx` into every freshly dilated voxel, so padding a hollow band seeded its interior as exterior. It now pads one layer at a time and seeds each new voxel with the sign of its 26-neighbours (batched path applies per-grid `joffsets`, since `GridBatch.neighbor_indexes` returns per-grid indices).

After the fix the narrow-band case matches the solid case exactly (0.076 vx, no flips) and a second pass reproduces `min(φ) = −2.99`.

## Rename: `retopologize_sdf` → `rebuild_narrow_band`

"Retopologize" is mesh-remeshing jargon and says nothing about what the op produces — a *narrow-band* SDF on a rebuilt grid — so it's neither discoverable nor descriptive. The new name follows OpenVDB's `levelSetRebuild`. Affected: `Grid.rebuild_narrow_band`, `GridBatch.rebuild_narrow_band`, `fvdb.functional.rebuild_narrow_band_{single,batch}`. Signature and return type are unchanged. **No deprecation alias is kept** (the op shipped in 0.5.x); `CHANGES.md` records it under 0.6.0 as breaking. `reinitialize_sdf` keeps its name — "reinitialization" is the standard level-set term for exactly this PDE.

## Documentation audit

Every docstring/comment describing this functionality was checked against the (fixed) behaviour:

- `Grid.rebuild_narrow_band` / `GridBatch.rebuild_narrow_band` still said new voxels are "seeded as exterior … for a hollow shell pass `pad=False`". That advice was already incorrect (`pad` never governed the in-kernel read); replaced.
- All `reinitialize_sdf` docstrings + the C++ header now state the input contract: only the **sign** of `field` is trusted; the surface is a sign change between *active* voxels (one voxel of each sign across the crossing suffices; a single-sign grid has no surface, so an occupancy mask needs a positive exterior layer first); inactive neighbours continue the adjacent sign (solids and hollow bands both valid); voxels that are **exactly 0** have a zero frozen sign and are left untouched (a no-data pass-through that `ray_implicit_intersection` relies on — #692), but their signed neighbours see them as an interface and smoothing blends them, so prune them first when possible; voxels must be isotropic; CUDA only, float32/float64.
- `SmoothingMode` now says smoothing is followed by a re-redistance and therefore *moves* the zero crossing, and that `MEAN_CURVATURE` erodes ~1-voxel features.
- `RayImplicitIntersection.cu` / `test_basic_ops.py` said `reinitialize_sdf` "leaves every active voxel outside the band at 0" — overstated; only exactly-0 input voxels stay 0 (verified: 150/150 with `smooth=0`, 2/150 with `smooth=2`).

## Isotropy check

The solve uses one voxel size but read only `voxelSizeAt(b)[0]`, so an anisotropic grid returned distances scaled by the aspect ratio along y/z. Now a per-grid `TORCH_CHECK_VALUE` (1e-6 relative tolerance, so float32 `0.1` → double round-off is accepted) raises `ValueError` naming the grid and its `voxel_size` before any GPU work. This is a behaviour change: callers feeding anisotropic grids were getting wrong answers and will now get an exception.

## Test plan

Built in the `fvdb` conda env (CUDA), `cd tests && pytest unit/test_sdf.py -v`:

- [x] 10 pre-existing tests unchanged (modulo the rename) and passing
- [x] `test_reinitialize_narrow_band_with_inactive_interior` — inner half-band error < 0.25 vx, comparable to outer half, no sign flips, deepest voxels reach the clamp
- [x] `test_rebuild_idempotent` — second pass with `pad=False` and `pad=True` reproduces the band
- [x] `test_pad_seeds_interior_of_hollow_band` — padded interior layers negative, exterior positive
- [x] `test_batch_pad_matches_single` — batched sign-aware padding matches the single-grid path
- [x] `test_rebuild_accepts_column_field` — `(N, 1)` scalar fields with the default `pad=True`, single and batched, match the flat `(N,)` result (review finding: the padding mask kept the trailing singleton dim; the field is now canonicalized to flat storage first)
- [x] `test_single_sign_field_has_no_surface` — an all-negative (occupancy) field has no surface: `reinitialize_sdf` returns the constant `-band*vx` and `rebuild_narrow_band` an empty band (correct for a fully-interior tile; the active-region boundary is not a surface). The occupancy recipe — one exterior layer seeded positive, then rebuild — recovers the boundary surface.
- [x] `test_anisotropic_voxels_rejected` — anisotropic raises `ValueError`; float32-round-off isotropic does not
- [x] API surface check: `retopologize_sdf*` absent from `Grid`, `GridBatch`, `fvdb.functional`; `rebuild_narrow_band*` present
- [x] `pytest unit/test_basic_ops.py -k ray_implicit_intersection` — 56 passed (comments touched there)
- [x] `clang-format --Werror` clean on the touched CUDA lines

## Notes

- The `feature/prune-layer-generative-examples` branch carries a rewrite of this kernel onto the register VBM decode; `faceValue` and the isotropy check will need re-applying there when it rebases.
- `_fvdb_cpp.pyi` / pybind registration needed no change (the C++ entry point `reinitialize_sdf` is unchanged; the rename is Python-side only).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,16 @@ fVDB Version History
 ## Version 0.6.0 - In Development
 
 - **PyTorch 2.13** fVDB updated to build, test and publish with PyTorch 2.13, CUDA 13.0/13.2 and Python 3.10-3.15 support.
+- **Breaking:** `retopologize_sdf` is renamed `rebuild_narrow_band` (`Grid`, `GridBatch`, and
+  `fvdb.functional.rebuild_narrow_band_{single,batch}`); no alias is kept. The old name was mesh-remeshing jargon
+  and did not say that the op produces a *narrow-band* SDF on a rebuilt grid.
+- Fixed `reinitialize_sdf` treating every inactive voxel as exterior. An `IndexGrid` has a single background slot,
+  so a narrow band whose interior is inactive (a pruned level set, an imported OpenVDB grid, or
+  `rebuild_narrow_band`'s own output) grew a phantom interface one voxel inside the true surface; the zero crossing
+  survived, but the inner half of the band was wrong and `rebuild_narrow_band` was not idempotent. Inactive
+  neighbours now continue the sign of the adjacent active voxel, and `pad=True` seeds new voxels from their
+  neighbours' sign instead of as exterior. `reinitialize_sdf` also now raises `ValueError` for anisotropic voxels
+  instead of silently using only `voxel_size[0]`.
 - **Breaking:** Unified sparse convolution and transposed-convolution geometry around the componentwise Torch-phase
   relation ``fine_ijk = stride * coarse_ijk + tap_ijk - padding_before``, where
   ``padding_before = floor((kernel_size - 1) / 2)`` and each zero-based tap component satisfies
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,10 @@ fVDB Version History
   neighbours now continue the sign of the adjacent active voxel, and `pad=True` seeds new voxels from their
   neighbours' sign instead of as exterior. `reinitialize_sdf` also now raises `ValueError` for anisotropic voxels
   instead of silently using only `voxel_size[0]`.
+  Occupancy inputs must include an explicit positive exterior layer around negative occupied voxels:
+  the active-region boundary no longer implicitly defines a surface, so rebuilding an all-negative
+  occupancy field yields an empty narrow band. Both SDF operations now require finite scalar values
+  with shape `(N,)` or `(N, 1)` and reject other shapes and NaN/Inf values with `ValueError`.
 - **Breaking:** Unified sparse convolution and transposed-convolution geometry around the componentwise Torch-phase
   relation ``fine_ijk = stride * coarse_ijk + tap_ijk - padding_before``, where
   ``padding_before = floor((kernel_size - 1) / 2)`` and each zero-based tap component satisfies
--- a/fvdb/functional/_sdf.py
+++ b/fvdb/functional/_sdf.py
@@ -113,6 +113,21 @@ def reinitialize_sdf_batch(
     Returns:
         sdf (JaggedTensor): The re-initialized SDF, same per-voxel ordering as ``field``.
 
+    Note:
+        * Only the **sign** of ``field`` is trusted; magnitudes are rebuilt. With ``smooth=0`` the
+          input's zero crossing is preserved (to sub-voxel accuracy); smoothing moves the surface to
+          its de-staircased position and then re-redistances.
+        * Inactive neighbours read as ``+/-band*vx`` with the sign of the adjacent active voxel, so
+          both filled solids (interior active) and narrow bands whose interior is inactive are valid
+          inputs. Voxels with no data are best left *inactive* rather than given a value.
+        * Voxels whose value is exactly ``0`` have a zero frozen sign and are left at ``0`` by the
+          redistance (a no-data pass-through relied on by the ray-implicit-intersection op, which
+          treats exact ``0`` as a gap). Their signed neighbours, however, see them as an interface and
+          are redistanced toward them, and smoothing blends them -- prune such voxels first when you
+          can.
+        * Each grid is assumed isotropic (only ``voxel_sizes[:, 0]`` is used). CUDA only;
+          ``float32`` or ``float64``.
+
     .. seealso:: :func:`reinitialize_sdf_single`, :func:`retopologize_sdf_batch`
     """
     result = _fvdb_cpp.reinitialize_sdf(
@@ -145,6 +160,12 @@ def reinitialize_sdf_single(
 
     Returns:
         sdf (torch.Tensor): The re-initialized SDF, shape ``(num_voxels,)``.
+
+    Note:
+        See :func:`reinitialize_sdf_batch` for the input contract: only the sign of ``field`` is
+        trusted, inactive neighbours continue the sign of the adjacent voxel (filled solids and
+        narrow bands with an inactive interior are both valid), exact-``0`` voxels are a no-data
+        pass-through that neighbours see as an interface, and voxels are assumed isotropic.
 
     .. seealso:: :func:`reinitialize_sdf_batch`, :func:`retopologize_sdf_single`
     """
@@ -202,6 +223,10 @@ def retopologize_sdf_batch(
         out_grid (GridBatch): The pruned (or, with ``prune=False``, the padded/original) grid batch.
         sdf (JaggedTensor): The narrow-band SDF, aligned with ``out_grid``.
 
+    Note:
+        Applying this to its own output reproduces it (up to a voxel layer at the band edge). See
+        :func:`reinitialize_sdf_batch` for the input contract.
+
     .. seealso:: :func:`retopologize_sdf_single`, :func:`reinitialize_sdf_batch`
     """
     # per-grid narrow-band half-width; voxel size may vary across the batch
@@ -255,6 +280,10 @@ def retopologize_sdf_single(
     Returns:
         out_grid (Grid): The pruned (or, with ``prune=False``, the padded/original) grid.
         sdf (torch.Tensor): The narrow-band SDF, aligned with ``out_grid``.
+
+    Note:
+        Applying this to its own output reproduces it (up to a voxel layer at the band edge). See
+        :func:`reinitialize_sdf_batch` for the input contract.
 
     .. seealso:: :func:`retopologize_sdf_batch`, :func:`reinitialize_sdf_single`
     """
--- a/fvdb/functional/_sdf.py
+++ b/fvdb/functional/_sdf.py
@@ -121,9 +121,8 @@ def reinitialize_sdf_batch(
           each sign across the crossing is sufficient (two or more per side gives the best sub-voxel
           accuracy). A grid whose active values are all one sign has no surface: the result is the
           constant ``-/+band*vx`` and :func:`rebuild_narrow_band_batch` returns an empty band, which
-          is correct for e.g. a tile that lies entirely inside an object. An occupancy mask must
-          therefore be given a positive exterior layer first (``dilated_grid(1)`` + ``inject`` with
-          ``default_value=+1``) -- its boundary is not treated as a surface.
+          is correct for e.g. a tile that lies entirely inside an object. The boundary of the active
+          region is not itself a surface.
         * Inactive neighbours read as ``+/-band*vx`` with the sign of the adjacent active voxel, so
           both filled solids (interior active) and narrow bands whose interior is inactive are valid
           inputs. Voxels with no data are best left *inactive* rather than given a value.
--- a/fvdb/functional/_sdf.py
+++ b/fvdb/functional/_sdf.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 """Functional API for signed-distance-field (SDF) re-initialization and narrow-band rebuild."""
+
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
@@ -21,6 +22,17 @@ if TYPE_CHECKING:
 def _to_cpp_smoothing(smoothing: SmoothingMode) -> "_fvdb_cpp.SmoothingMode":
     """Convert a public :class:`fvdb.SmoothingMode` to the bound C++ enum (matched by member name)."""
     return getattr(_fvdb_cpp.SmoothingMode, smoothing.name)
+
+
+def _validated_scalar_field(field: torch.Tensor, num_voxels: int) -> torch.Tensor:
+    """Validate before padding, where NaN is reserved for newly activated voxels."""
+    if field.dim() not in (1, 2) or (field.dim() == 2 and field.shape[1] != 1):
+        raise ValueError(f"field must be a scalar field with shape (N,) or (N, 1), got {tuple(field.shape)}")
+    if field.shape[0] != num_voxels:
+        raise ValueError(f"field must have one value per voxel, got {field.shape[0]} values for {num_voxels} voxels")
+    if not torch.isfinite(field).all().item():
+        raise ValueError("field must contain only finite values; leave no-data voxels inactive")
+    return field.reshape(-1)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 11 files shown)


fvdb/functional/_sdf.py
```diff
@@ -91,7 +166,13 @@ def reinitialize_sdf_single(
     Returns:
         sdf (torch.Tensor): The re-initialized SDF, shape ``(num_voxels,)``.
 
-    .. seealso:: :func:`reinitialize_sdf_batch`, :func:`retopologize_sdf_single`
+    Note:
+        See :func:`reinitialize_sdf_batch` for the input contract: only the sign of ``field`` is
+        trusted, inactive neighbours continue the sign of the adjacent voxel (filled solids and
+        narrow bands with an inactive interior are both valid), exact-``0`` voxels are a no-data
+        pass-through that neighbours see as an interface, and voxels must be isotropic.
```
harrism: **💡 suggestion:** ‏If we are going to refer to the reader somewhere else for the details of the input contract, and it's only 50 lines away, we don't need to also summarize the contract here. That just invites comment drift. 


```suggestion
        See :func:`reinitialize_sdf_batch` for the input contract.
```

src/fvdb/detail/ops/ReinitializeSdf.cu
```diff
@@ -70,12 +94,13 @@ signFusedKernel(const OnIndexGridT *grid,
...
+                ScalarT bandWidth,
                 ScalarT *sign) {
     VBM_FACES_BEGIN();
-    ScalarT xm = field[faceIndex[0]], xp = field[faceIndex[1]], ym = field[faceIndex[2]],
-            yp = field[faceIndex[3]], zm = field[faceIndex[4]], zp = field[faceIndex[5]];
+    const ScalarT phiCenter = field[centerIndex];
+    VBM_FACE_VALUES(field, phiCenter, bandWidth);
     ScalarT gradX = (xp - xm) / (2 * voxelSize), gradY = (yp - ym) / (2 * voxelSize),
-            gradZ = (zp - zm) / (2 * voxelSize), phiCenter = field[centerIndex];
+            gradZ = (zp - zm) / (2 * voxelSize);
```
harrism: **💡 suggestion:** ‏One assignment per line, please (avoid comma operator):

```
ScalarT gradX = ...;
ScalarT gradY = ...;
ScalarT gradZ = ...;
```

(Unless using structured binding.)

harrism (approved):
Looks great. Just two small suggestions.

fvdb/functional/_sdf.py
swahtz: Done — the note now just points at `reinitialize_sdf_batch`.

src/fvdb/detail/ops/ReinitializeSdf.cu
swahtz: Done — split into three separate assignments.

fvdb/functional/_sdf.py
```diff
@@ -23,6 +23,61 @@ def _to_cpp_smoothing(smoothing: SmoothingMode) -> "_fvdb_cpp.SmoothingMode":
...
+    ``+/-band_width`` according to the sign of its already-present 26-neighbours.
+
+    An IndexGrid cannot mark inactive space as interior or exterior, so a constant exterior seed
+    would turn the inside of a hollow narrow band into "outside". Growing one layer at a time and
+    copying the sign of the neighbours lets the padding continue the field inward and outward.
+    """
+    for _ in range(band):
+        dilated = grid.dilated_grid(1)
+        padded = inject_single(dilated, grid, field, default_value=float("nan"))
+        is_new = torch.isnan(padded)
```
harrism, line 59: **🚩 issue:** ‏What about pre-existing NANs in the grid? This conflates them with new cells added by dilation.

fvdb/functional/_sdf.py
```diff
@@ -58,7 +113,27 @@ def reinitialize_sdf_batch(
     Returns:
         sdf (JaggedTensor): The re-initialized SDF, same per-voxel ordering as ``field``.
 
-    .. seealso:: :func:`reinitialize_sdf_single`, :func:`retopologize_sdf_batch`
+    Note:
+        * Only the **sign** of ``field`` is trusted; magnitudes are rebuilt. With ``smooth=0`` the
```
harrism: > _Analysis by Claude Code, run by @harrism._

**"Only the sign of `field` is trusted" is not accurate — magnitudes affect the result.**

`signFusedKernel` (`ReinitializeSdf.cu:105-109`) computes the frozen Peng sign from `phiCenter` **and** the magnitude-derived central gradient, and `phi` is initialized to the raw field rather than to `±band_width`. So the input magnitudes feed both the sign field and the initial condition of the redistancing.

The PR's own tests show the difference: `test_reinitialize_preserves_good_sdf` tolerates `0.25·vx`, while `test_reinitialize_recovers_from_step` — same signs, but `±band_width` magnitudes — needs `0.6·vx` to pass. If only the sign mattered those two would converge to the same answer.

This matters because the sentence invites a caller to pass a raw occupancy mask or arbitrary ±1 values and expect the same output as a well-formed SDF. They will instead get a slower-converging and less accurate result.

Suggest something closer to: *"the sign of `field` is authoritative; magnitudes need not be a true distance but do affect convergence and accuracy — a rough SDF converges better than a binary mask."*

The same sentence appears at `fvdb/grid.py:1531` and `fvdb/grid_batch.py:1090` and would want the same edit.

fvdb/functional/_sdf.py
```diff
@@ -136,36 +214,39 @@ def retopologize_sdf_batch(
...
+        :func:`reinitialize_sdf_batch` for the input contract.
+
+    .. seealso:: :func:`rebuild_narrow_band_single`, :func:`reinitialize_sdf_batch`
     """
     # per-grid narrow-band half-width; voxel size may vary across the batch
     band_width = band * grid.voxel_sizes[:, 0]
+    # Canonicalize scalar fields to flat per-voxel storage: reinitialize_sdf accepts (N, 1) (and
+    # returns (N,)), and the padding masks below must be one-dimensional.
+    if field.jdata.dim() != 1:
+        field = field.jagged_like(field.jdata.reshape(-1))
```
harrism: > _Analysis by Claude Code, run by @harrism._

**`reshape(-1)` silently accepts a multi-channel field instead of rejecting it.**

The comment above says the intent is to canonicalize `(N, 1)` to `(N,)`, which is right. But `reshape(-1)` also happily flattens `(N, 3)` into `3N` values — so a caller who passes a vector field (a gradient, a normal map, an RGB volume) gets no error here. The failure surfaces later inside `jagged_like` or the padding masks as a length mismatch, with a message that says nothing about channels.

A guard makes the contract explicit and the error message useful:

```python
if field.jdata.dim() != 1:
    if field.jdata.dim() != 2 or field.jdata.shape[1] != 1:
        raise ValueError(
            f"field must be a scalar field with shape (N,) or (N, 1), got {tuple(field.jdata.shape)}"
        )
    field = field.jagged_like(field.jdata.reshape(-1))
```

`rebuild_narrow_band_single` has the same shape at `:301` (`field = field.reshape(-1)`).

src/fvdb/detail/ops/ReinitializeSdf.cu
```diff
@@ -108,11 +134,11 @@ godunovFusedKernel(const OnIndexGridT *grid,
...
                    const ScalarT *sign,
                    ScalarT voxelSize,
+                   ScalarT bandWidth,
                    ScalarT *rhs) {
     VBM_FACES_BEGIN();
-    ScalarT center = field[centerIndex], sgn = sign[centerIndex];
-    ScalarT xm = field[faceIndex[0]], xp = field[faceIndex[1]], ym = field[faceIndex[2]],
-            yp = field[faceIndex[3]], zm = field[faceIndex[4]], zp = field[faceIndex[5]];
+    const ScalarT center = field[centerIndex], sgn = sign[centerIndex];
+    VBM_FACE_VALUES(field, center, bandWidth);
```
harrism: > _Analysis by Claude Code, run by @harrism. This one is speculative — flagging for your judgement rather than asserting a bug._

**`VBM_FACE_VALUES` is passed the live RK-stage `center` rather than the frozen `sgn`.**

The fix in this PR is right, and for the right reason: because `upwind()` selects `Max(back, 0)` / `Min(fwd, 0)` according to `sgn`, continuing the centre's sign into an inactive face makes that face *downwind*, so it contributes zero. That is the correct outflow boundary condition, not a heuristic — nice.

The question is which sign is used to do the continuing. `center` here is the current RK stage value, which evolves; `sgn` is the Peng sign frozen once in `signFusedKernel`. If a stage value crosses zero next to inactive space, the synthesized boundary value flips by `2·bandWidth` and that face becomes *upwind* instead of downwind — feeding the phantom value into `gradMag` rather than cancelling it.

Whether this is reachable in practice I could not determine statically. It is self-limiting: the Peng sign is small near a zero crossing, so the RHS is damped exactly where the flip could occur, and a stage value only crosses zero if the surface is moving through that voxel. It may be unreachable for any input you care about.

Passing `sgn` instead of `center` would remove the coupling entirely and make the boundary condition depend only on the frozen sign field, which is what the surrounding `upwind` calls already use. If there is a reason to prefer the live value, a one-line comment saying so would stop the next reader asking the same question.

CHANGES.md
```diff
@@ -4,6 +4,16 @@ fVDB Version History
 ## Version 0.6.0 - In Development
 
 - **PyTorch 2.13** fVDB updated to build, test and publish with PyTorch 2.13, CUDA 13.0/13.2 and Python 3.10-3.15 support.
+- **Breaking:** `retopologize_sdf` is renamed `rebuild_narrow_band` (`Grid`, `GridBatch`, and
+  `fvdb.functional.rebuild_narrow_band_{single,batch}`); no alias is kept. The old name was mesh-remeshing jargon
+  and did not say that the op produces a *narrow-band* SDF on a rebuilt grid.
+- Fixed `reinitialize_sdf` treating every inactive voxel as exterior. An `IndexGrid` has a single background slot,
```
harrism, line 10: > _Analysis by Claude Code, run by @harrism._

**This entry describes the fix but not the behaviour change a user will actually notice.**

The bullet explains the inactive-interior bug well. What it does not say is that an all-one-sign field — a raw occupancy mask, the most likely thing a user reaches for — now produces an **empty** `rebuild_narrow_band` result, where before it produced a surface at the active-region boundary.

That surface was accidental (it came from the same inactive-voxel misreading being fixed here) and the docstrings and tests do cover the new behaviour. But someone upgrading who was relying on it — knowingly or not — sees an empty grid and no explanation in the changelog. That is the kind of thing that turns into a bug report against the release rather than a `git log` search.

Suggest an extra sentence on this bullet, roughly: *"As a consequence, a field with no sign change (e.g. a raw occupancy mask) now yields an empty narrow band instead of a surface at the active-region boundary; seed such inputs with a signed field first."*

harrism (changes requested):

## Review: inactive-interior fix + rename

**The core fix is correct and the rename is complete.** Two things worth confirming, since they are the parts a reviewer is most likely to take on trust:

- The fix is right *for the right reason*, not just empirically: because `upwind()` selects `Max(back, 0)` / `Min(fwd, 0)` by `sgn`, continuing the centre's sign into an inactive face makes that face downwind, so it contributes zero to `gradMag`. That is the correct outflow BC for the Hamilton-Jacobi problem.
- The rename is clean. `grep -ri retopolog` over the whole tree hits only the historical 0.5.0 `CHANGES.md` entries and the intentional 0.6.0 migration note — no missed call sites, stubs, docs, or exports. `fvdb/_fvdb_cpp.pyi:934` exposes only `reinitialize_sdf`, unaffected.

One thing worth recording so it is not re-raised later: `sign = phiCenter / sqrt(...)` does not produce `0/0 → NaN` for a block of exactly-zero no-data voxels (the `integrate_tsdf` fill this PR documents), because of the `+ 1e-12` in the radicand at `ReinitializeSdf.cu:109`. Safe in float32 too.

Requesting changes on the documentation-accuracy items rather than the mathematics — the inline comments have detail.

### Blocking-ish

1. **`_sdf.py:117`** (and `grid.py:1531`, `grid_batch.py:1090`) — "Only the **sign** of `field` is trusted" is not accurate; magnitudes feed both the frozen Peng sign and the initial condition. The PR's own two tests need different tolerances (`0.25·vx` vs `0.6·vx`) for inputs that differ only in magnitude, which is the proof.
2. **`_sdf.py:239`** and **`:301`** — `reshape(-1)` silently accepts a multi-channel field where it means to accept `(N, 1)`.
3. **`CHANGES.md:10`** — does not mention that an all-one-sign field now yields an empty result.

### Minor / your call

4. **`_sdf.py:34`** — `is_new = torch.isnan(padded)` conflates pre-existing NaNs with newly dilated voxels (your comment, included here).
5. **`ReinitializeSdf.cu:346`** — the comment `// ping-pong (stage[0] already = bandWidth)` still refers to the slot-0 boundary mechanism this PR removes. Not commentable inline as the line is unchanged context.
6. **`ReinitializeSdf.cu:141`** — speculative, on whether the frozen `sgn` would be safer than the live stage `center` for the synthesized face value. Detail inline; may well be unreachable.
7. `__all__` in `functional/__init__.py:272-275` is no longer alphabetical. The repo has no isort/ruff config so CI will not care.

fvdb/functional/_sdf.py
swahtz: Addressed in 46f367bc. Existing NaNs really could be mistaken for new padding and overwritten with +B. Both operations now require finite inputs: rebuild validates before padding, and the C++ redistance entry point rejects NaN/Inf too. This keeps NaN reserved for the internal padding marker; callers should leave no-data voxels inactive.

Added regression coverage for NaN and both infinities through the single-grid and batched APIs, with padding enabled and disabled.

fvdb/functional/_sdf.py
swahtz: Updated in 46f367bc across the functional API, Grid, GridBatch, and the C++ header. The contract now says the field must represent the intended inside/outside regions and zero crossings; magnitudes need not be accurate distances, but affect convergence, accuracy, and sub-voxel surface location.

Also qualified surface preservation as the aim of redistancing, subject to discretization error, rather than implying that the frozen sign explicitly pins every interpolated crossing.

fvdb/functional/_sdf.py
swahtz: Addressed in 46f367bc. Both rebuild paths now validate (N,) or (N, 1) and the voxel count before flattening or padding. The C++ reinitialize entry point also explicitly validates the scalar layout.

The ordinary (N, 3) case previously failed later with a less useful error, while shapes such as (1, N) could actually be flattened and accepted. Added regressions for those cases, (N, 1, 1), and incorrect voxel counts. The existing column-field parity test still passes.

src/fvdb/detail/ops/ReinitializeSdf.cu
swahtz: Changed to the frozen sign in 46f367bc, with a CUDA regression covering both sign orientations. The stage-flip case is reachable for a rough signed field: with h=1 and B=3, set the centre to +0.1, its +x neighbour to +3, its four +/-y and +/-z neighbours to -3, and leave -x inactive. The central gradient cancels, so the frozen Peng sign is approximately +1 despite the small centre value. The first RK stage takes the centre to about -1.253625.

Using the live centre then flips the missing-face value to -B and introduces an upwind contribution. Using the frozen sign keeps that face downwind for the clamped stage values. The regression checks the resulting RK3 centre value against the independently evaluated update, for both polarities.

This change is limited to Godunov reads. Initial sign computation and smoothing still use the current field, and the comments/docstrings now make that distinction explicit. It prevents this boundary feedback; it does not promise preservation of every under-resolved feature.

CHANGES.md
swahtz: Added in 46f367bc. The changelog now explicitly states that rebuilding an all-negative occupancy field yields an empty narrow band, and gives the migration recipe: add a positive exterior layer around the negative occupied voxels before reconstruction. It also records the new scalar-shape and finite-value validation.

swahtz:
Addressed the latest review in 46f367bc, with replies on each inline thread. Also clarified the unused slot-0 comment and reordered the SDF exports in __all__.

Rebuilt and installed in the fvdb conda environment, then ran the CUDA SDF suite from tests/: 21 tests and 18 subtests passed, including the frozen-sign boundary regression and invalid-input checks. Black, clang-format 18, and whitespace checks passed.

harrism (approved):
Thanks for addressing all the feedback!

Issue #761, opened 2026-09-03
CI: reuse acquired GPU runners instead of demanding three per PR event


harrism:
## Problem

PR CI goes red when EC2 refuses `g6.xlarge` in us-east-2. Every PR event fires three
workflows — `tests.yml`, `cu130.yml`, `cu132.yml` — and each demands its own GPU runner
within seconds of the others. When the region can't serve three at once, the losers fail
the whole workflow.

## Data (54 GPU-runner start jobs, both repos, 2026-08-27 .. 09-03)

39 succeeded, 14 failed. Every failure was `InsufficientInstanceCapacity`; none had
another cause. Splitting the failures by whether capacity existed nearby:

| | count |
|---|---|
| a **sibling job on the same commit** succeeded within ±5 min | 6/14 |
| **no** GPU start succeeded anywhere within ±5 min | 6/14 |
| other (success nearby, different commit) | 2/14 |

So roughly half self-inflicted contention, half genuine scarcity. Two supporting facts:

* **We race ourselves.** The self-contention cases are three adjacent run IDs from one
  event. Example, commit `8dfa68e`: Conda succeeded 04:32:01, cu130 **failed** 04:33:12,
  cu132 succeeded 04:33:17 — capacity existed 5 s after the refusal.
* **Scarcity is real too.** All six no-capacity-anywhere cases are fvdb-reality-capture,
  mostly 09-01/09-02 while it still requested us-east-2b only. Whole-region (all three AZ)
  exhaustion was observed 6 times: 2× on 08-31, 4× on 09-02. 09-03 was clean.

Already landed: multi-AZ fallback (#705, openvdb/fvdb-reality-capture#326) and
retry-with-backoff (#760, openvdb/fvdb-reality-capture#329). Neither reduces peak demand.

## Suggested approach: reuse the runners we did get

Merge the three workflows into one run, attempt three starts with `continue-on-error`,
then assign the three test tasks across whichever labels actually came up:

```yaml
  start:                       # 3 attempts, failures tolerated
    strategy: { fail-fast: false, matrix: { n: [1, 2, 3] } }
    uses: ./.github/workflows/start-ec2-runner.yml

  plan:                        # round-robin tasks over the labels that exist
    needs: start               # e.g. [{task: conda, label: L1},
                               #       {task: cu130, label: L1},
                               #       {task: cu132, label: L2}]
  test:
    needs: plan
[...]

PR #763, opened 2026-09-08
CI: build PR wheels for SM 8.6 and fall back to g5.xlarge for GPU runners

harrism:
Addresses the **scarcity half** of #761. The contention half (runner reuse) is deliberately not here — see the note at the bottom.

## Why

Of 14 `InsufficientInstanceCapacity` failures between 2026-08-27 and 09-03, **six had no successful `g6.xlarge` launch anywhere within ±5 minutes** — the pool was empty, not contended. Retrying the same instance type harder cannot fix those. Drawing on a second pool can.

## `g5.xlarge` is a viable runner

Nothing in the tree requires SM 8.9:

* CUTLASS convolution backend targets `arch::Sm80` (`PredGatherIGemm.cu`)
* the only capability gate is `__CUDA_ARCH__ >= 800` (`AtomicAdd.cuh`)
* Flash Attention needs >= 8.0

`g5.xlarge` is A10G / SM 8.6 and satisfies all of it.

## The blocker was the PR build architecture

`arch_list_pr` was `"8.9+PTX"`. On an SM 8.6 device that has **no usable cubin** — binary compatibility runs forward across minor revisions, not backward — and its embedded PTX is `compute_89`, which cannot be JIT-compiled *down* to `sm_86`. A g5 runner would fail outright, not run slowly.

`"8.6+PTX"` covers both: g5 matches exactly, and g6 loads the same cubin because 8.6 is binary compatible with 8.9. **No PTX JIT is involved on either**, so there is no runtime cost to this.

It also closes a fidelity gap. `arch_list_publish` contains no 8.9 — so users on Ada already run the 8.6 cubin, and publish validation exercises exactly that path on `g6.xlarge` every release. PR CI was testing an 8.9-native binary that no user ever receives.

## The fallback

`start-ec2-runner.yml` gains an optional `fallback-instance-type`:

* attempt 1 — primary (`g6.xlarge`), sweeping all three AZs
* attempt 2 — the fallback (`g5.xlarge`), **with no backoff first**: a different capacity pool has no reason to be waited out
* attempt 3 — back to the primary, after the usual backoff

With no fallback configured the behaviour is byte-for-byte what it was. Wired up at all four `g6.xlarge` call sites (`tests.yml`, `cu130.yml`, `cu132.yml`, `publish.yml`).

## Testing

`actionlint` clean; token-policy checker OK on 17 workflows; 26 policy unit tests pass; all workflows and `versions.json` parse.

**This cannot be verified pre-merge.** `tests.yml` calls `load-versions.yml` without a `ref`, so `versions.json` is read from the *base* branch — under `pull_request_target` this PR's own run will still build 8.9 from `main`. The change takes effect on the first PR after merge. Both edits are small and independently revertible for that reason.

## Not in scope

The other half of #761 — three workflows on one PR event each demanding a runner within seconds, when the pool cannot serve three. Fixing that means reusing the runners we did get, which changes job names and therefore the nine required status checks in ruleset `main branch protection`. That needs the team's agreement first.


swahtz (approved):
Looks great, thanks for doing this.

PR #765, opened 2026-09-09
Add CPU implementation for computeGaussianNanInfMask

matthewdcong:
For debugging, we often want to generate and write debug PLYs on the CPU. However, Gaussian PLY output calls `computeGaussianNanInfMask` which currently only has a `kCUDA` implementation. This leads to unnecessary round-trip traffic between the CPU and GPU which we can shortcut by simply implementing the operator on the CPU.


src/fvdb/detail/ops/gsplat/ComputeGaussianNanInfMask.cu
```diff
@@ -222,8 +219,60 @@ dispatchComputeGaussianNanInfMask<torch::kCPU>(const fvdb::JaggedTensor &means,
...
+    }
+
+    const auto N = means.rsize(0);
+    auto outValid =
+        torch::empty({N}, torch::TensorOptions().dtype(torch::kBool).device(means.device()));
+
+    AT_DISPATCH_V2(means.scalar_type(),
+                   "computeGaussianNanInfMaskCPU",
+                   AT_WRAP([&] {
+                       const auto meansAccessor     = means.jdata().accessor<scalar_t, 2>();
```
swahtz, line 242: **🚩 issue:** ‏We should have a check that all these `JaggedTensors` are on the CPU before calling `accessor` in the dispatch.  We use `means`' device but we should make sure these all live on the host as part of the validation phase.

swahtz (changes requested):
All looks pretty straightforward.  Just one check I think we should add.

src/fvdb/detail/ops/gsplat/ComputeGaussianNanInfMask.cu
matthewdcong: Added

swahtz (approved):
Thanks, let's ship it!

PR #768, opened 2026-09-10
CI: stop leaking EC2 runners when a run is cancelled mid-provisioning

swahtz:
Fixes the orphaned EC2 instances we have been seeing. Investigation prompted by the `cancel-in-progress` discussion on #745.

## Root cause

Every PR workflow uses a per-PR concurrency group with `cancel-in-progress: true`. When a new push cancels a run while `start-ec2-runner.yml` is in flight, GitHub kills the job after the instance has launched but before the runner registers. The `select` step never runs, so the job's outputs are empty. The stop job still runs (`always()`), but it receives an empty label and id, and its `RunnerLabel` tag lookup finds nothing because the PR workflows never tagged their instances. The instance is orphaned.

Evidence from cancelled Build-and-Test runs since 2026-08-15:

| | |
|---|---|
| Cancelled runs | 44 |
| Start-runner jobs killed mid-flight | 37 |
| Killed 30s+ after the launch call (instance certainly running) | 9 |
| Stop jobs that failed after a cancelled start | 37 of 37 |

Example: run 34444278482 (2026-09-10). Start job log ends `Attempting to start EC2 instance ... The operation was canceled.` Stop job log shows `label:` and `ec2-instance-id:` blank and `ERROR: No instance found with RunnerLabel=`.

This predates #760. The August runs with the hand-rolled jobs show the same pattern. #760's cleanup steps only cover attempts whose `outcome == 'failure'`; a cancelled step never reaches them.

## Fix

Two independent routes to the instance, so either alone is enough:

- **`start-ec2-runner.yml`**: `if: cancelled()` steps after the attempts terminate any attempt that launched an instance and did not fail. The action publishes its outputs immediately after `RunInstances`, so the id survives the kill even though the job's outputs do not.
- **Tag every instance** with `RunnerLabel` (when a label is given), `GitHubRunId` and `GitHubRepository`. Tags are composed inside the reusable workflow, with caller-supplied keys winning on collision, so callers only need to pass `runner-label`. `cu130.yml`, `cu132.yml` and `tests.yml` now pass a deterministic `<wf>-<build|gpu>-<run_id>-<run_attempt>` label to both the start and stop jobs, matching what `publish.yml` and `nightly-publish.yml` already do. `stop-ec2-runner.yml` can then recover the instance by tag when the start job produced no outputs.

`stop-ec2-runner.yml` now treats "no instance with that tag" as a warning rather than a failure. With the cleanup above, that is the expected outcome when the start job was cancelled before, or without, launching anything.

## Not changed

`cancel-in-progress` stays on for the PR workflows. Letting stale builds run to completion would cost more instance hours than the orphans do. #745 sets it to `false` for the once-weekly job only, and that branch will pick up the `runner-label` convention once this merges.

## Verification

`actionlint` (with CI's shellcheck excludes) is clean repo-wide. `zizmor --persona=regular` reports the same low-severity findings as `main`. `check_runner_token_policy.py` passes and its tests pass. The tag-composition `jq` was exercised locally with and without caller-supplied tags.


harrism (approved):
The 3-way code duplication a little annoying, but its fine. Looks good, thanks.

PR #592, opened 2026-04-09
Fix tutorial docs: move from wip, fix broken APIs, add CI testing

harrism:
Closes #591

## Summary

- **Move `docs/wip/` → `docs/tutorials/`** — these are no longer WIP; the directory name now matches its purpose
- **Fix all broken API references** across 10 tutorial files (78 code blocks total):
  - `grid_to_world` → `voxel_to_world`, `world_to_grid` → `world_to_voxel`
  - `grid_to_world_matrices` → `voxel_to_world_matrices`
  - Remove `VDBTensor` (deleted class) — rewrite `basic_convolution.md` and `simple_unet.md` to use explicit `(JaggedTensor, ConvolutionPlan)` API
  - `SparseConv3d(transposed=True)` → separate `SparseConvTranspose3d` class
  - `fvnn.ReLU` → `torch.nn.ReLU` (removed from fvdb.nn)
  - `fvdb.save_gridbatch` / `fvdb.load_gridbatch` → `grid.save_nanovdb` / `fvdb.functional.load_nanovdb`
  - `fvdb.jrand` / `fvdb.jzeros` / etc. → `JaggedTensor.from_rand` / `JaggedTensor.from_zeros` / etc.
- **Add tutorials to `docs/index.rst`** toctree so they're discoverable in Sphinx docs
- **Update TEACHME cross-references** from `docs/wip/` to `docs/tutorials/`

## Testing

- **69 of 78 tutorial code blocks** are now tested via `pytest --markdown-docs` (88.5%)
- 9 `notest` blocks remain, all genuinely untestable (require MinkowskiEngine, nerfacc, or `nanovdb_convert` CLI)
- **234 API references validated** by `test_teachme_api_refs.py` (209 inline prose + 25 notest block refs)
- `--doc-dirs` option added so the API ref test can target specific directories
- CI workflow updated: removes stale `--ignore-glob="**/wip/**"`, adds API ref validation step

## Test plan

- [x] `pytest --markdown-docs docs/tutorials/` — 69 code blocks pass
- [x] `pytest --markdown-docs docs/TEACHME/` — 43 code blocks pass (unchanged)
- [x] `pytest tests/test_teachme_api_refs.py -v` — 234 API refs validated
- [x] `pytest tests/test_teachme_api_refs.py --doc-dirs docs/tutorials` — 140 refs (tutorials only)
- [x] Unit tests unaffected by new `conftest.py`
- [x] CI workflow adds API ref validation step alongside existing markdown-docs tests

[...]

PR #449, opened 2026-02-06
build: add black/clang-format helpers

fwilliams:
## Summary 
 - Add build.sh subcommands for CI-aligned Python (black) and C++/CUDA (clang-format) formatting.
 - Add scripts to run black and clang-format-18 (with helpful errors if missing).\n\n## Notes\n- Intended as a standalone formatting/tooling PR off main.

PR #769, opened 2026-09-10
build: add 'format' subcommand and fix run_clang_format.sh

swahtz:
## Summary

Adds `./build.sh format`, a single command that formats the repo the way CI checks it, and fixes the existing clang-format script.

- `./build.sh format` runs clang-format over `src/` and then black over the repo. `./build.sh format check` reports violations without rewriting files and exits non-zero, matching the CI codestyle jobs.
- Black reads its settings from the `[tool.black]` section in `pyproject.toml` added in #592, so this PR duplicates no black flags.
- `src/scripts/run_clang_format.sh` previously had its shebang after the license header, only worked when invoked from `src/scripts`, and did not pass `-style=file`. It now walks `src/` with the same extensions CI uses, prefers `clang-format-18` (which `env/dev_environment.yml` provides), warns when another version is used, and supports `check` mode.

Supersedes #449. Per the discussion there, the standalone black script is dropped in favor of the pyproject config.

## Testing

- `./build.sh format bogus` and `./build.sh format check extra` reject the arguments.
- `./build.sh format check` runs clang-format in dry-run mode and reports violations, then runs black in check mode. Exit code is non-zero if either reports a problem.
- `./build.sh format` rewrites files in place. With a local clang-format 21 the version warning fires and the reformat differs from CI, which is why the warning exists.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -13,6 +13,8 @@ usage() {
   echo "                   wheel      - Build the Python wheel."
   echo "                   ctest      - Run tests (requires tests to be built)."
   echo "                   docstest   - Run pytest markdown documentation tests."
+  echo "                   format     - Run clang-format and black over the repo (matches CI)."
+  echo "                                Pass 'check' to report violations without rewriting files."
   echo ""
   echo "Options:"
   echo "  -h, --help     Display this help message and exit."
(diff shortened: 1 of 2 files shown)


harrism:
Looks good. I still think pre-commit would be a better way to drive all the style checks, but this is a good step.

harrism (approved):

Issue #770, opened 2026-09-11
Finish adopting NanoVDB's injectable CUDA memory resources (openvdb #2232): move grid storage off the dual-space TorchDeviceBuffer


swahtz:
## Why now

Upstream has merged the whole "introduce + deprecate" arc of [openvdb #2232](https://github.com/AcademySoftwareFoundation/openvdb/issues/2232), shipping in OpenVDB 13.1:

- [#2288](https://github.com/AcademySoftwareFoundation/openvdb/pull/2288) single-space `GridHandle<cuda::Buffer<std::byte, R>>`
- [#2292](https://github.com/AcademySoftwareFoundation/openvdb/pull/2292) `cuda::copyTo` cross-space transfers, pinned handles
- [#2293](https://github.com/AcademySoftwareFoundation/openvdb/pull/2293) `reset()` via `destroy()`, single-space NodeManager
- [#2301](https://github.com/AcademySoftwareFoundation/openvdb/pull/2301) `DeviceBuffer` deprecated; `cuda/HandleStorage.h` lets every tool emit single-space output; `VoxelBlockManagerHandle` gains single-space accessors; `TopologyBuilder` stages its root via `allocateProcessedRoot`/`uploadProcessedRoot`
- [#2319](https://github.com/AcademySoftwareFoundation/openvdb/pull/2319) `DistributedPointsToGrid` per-device resources; `UnifiedBuffer` deprecated

One release later upstream deletes `hasDeviceDual`, the dual `GridHandle` accessors and the `BufferT::create` fallback. `nanovdb::GridHandle<fvdb::TorchDeviceBuffer>` is our grid storage everywhere (`GridBatchData.h`), and it depends on exactly those things. We should migrate inside the deprecation window rather than at the removal.

## Where we are

- #732 routed builder scratch through `BuilderResource` (= `TorchResource`, PyTorch's caching allocator) at all 13 CUDA builders plus our own `PadGrid.cuh`.
- #752 bumped the pin past #2288 and converted the small ready-now pieces (VBM metadata buffers via an adapter, SaveNanoVDB staging). It is now behind `main` (which moved to the #2220 pin in #754) and predates #2292/#2293/#2301/#2319.
[...]

PR #752, opened 2026-08-23
Update NanoVDB for single-space GridHandle and widen Torch-pool coverage

swahtz:
## What this PR does

Follow-up to #732. Upstream has since merged openvdb [#2286](https://github.com/AcademySoftwareFoundation/openvdb/pull/2286) (small-builder `ResourceT` seams — already in our previous pin, and our only relevant call site, `indexToGrid` in `SaveNanoVDB.cu`, was already routed) and [#2288](https://github.com/AcademySoftwareFoundation/openvdb/pull/2288) (single-space device-only `GridHandle<cuda::Buffer<T,R>>`, step 3 part 1 of openvdb [#2232](https://github.com/AcademySoftwareFoundation/openvdb/issues/2232)). This PR advances the pin past #2288 and widens Torch-allocator coverage where it is reachable today.

### 1. Pin bump: `7946f17e` → `e679862f`
The merge commit of upstream #2288. The new `hasDeviceSingle`/`hasHostSingle` traits are *detected*, defaulting to false, so `TorchDeviceBuffer`'s dual-trait specialization compiles and behaves unchanged.

### 2. ReinitializeSdf: VBM metadata buffers through the Torch pool
The VoxelBlockManager's `firstLeafID`/`jumpMap` buffers were allocated with `nanovdb::cuda::DeviceBuffer` — outside torch's pool (and, with `NANOVDB_USE_SYNC_CUDA_MALLOC` builds, via synchronous `cudaMalloc`). They now allocate through `BuilderResource` via a local `TorchVbmBuffer` adapter over the single-space `nanovdb::cuda::Buffer`, stream-ordered on the reinit stream.

**Why an adapter and not `cuda::Buffer` directly:** `VoxelBlockManagerHandle`'s device accessors are gated on `BufferTraits<BufferT>::hasDeviceDual` and `static_cast` their pointers from a `void*`-returning `deviceData()`, and `buildVoxelBlockManager` allocates through the legacy `BufferT::create(bytes, guide, device, stream)` static interface. Bringing the VBM handle into the single-space world is an upstream follow-up on the #2232 roadmap (noted in the adapter's doc comment).

### 3. SaveNanoVDB: device staging onto `cuda::Buffer<…, BuilderResource>`
[...]

PR #773, opened 2026-09-11
Advance NanoVDB pin to openvdb #2319 and adopt single-space cuda::Buffer where it now fits (#770 steps 1-2)

harrism:
Supersedes #752 (carrying its commit, with @swahtz as author). Steps 1–2 of #770, plus a latent-bug fix the new pin's validation exposed.

## Summary

Advances the NanoVDB pin to `31d1e596` (merge of openvdb #2319) — past the whole "introduce + deprecate" arc of openvdb #2232 that has landed so far — and uses the single-space `nanovdb::cuda::Buffer` directly at the three device-only scratch/output sites where #2301 removed the reasons for an adapter or a dual-space type.

**Grid storage is unchanged.** `GridBatchData`'s `GridHandle<TorchDeviceBuffer>` stays as it is; that flip is #770 steps 3–5 and deliberately not in this PR.

## Commit 1 — normalize grid headers before constructing `GridHandle`s (new)

A pre-existing bug, surfaced by the pin: NanoVDB #2292/#2301 make `GridHandle` **validate a host buffer's grid chain at construction** — every header must be valid, carry index *i* of count *N*, and have a size that fits. Two fvdb sites copied a header out of a multi-grid batch into a single-grid buffer and built a handle around it without resetting those fields:

* `SaveNanoVDB.cu` `indexToGridHost()` — `*dstGrid->data() = *srcGrid->data()` kept the source's `mGridIndex`, `mGridCount` and (index-grid) `mGridSize`. The CUDA `tools::cuda::indexToGrid` it ports resets index/count (`IndexToGrid.cuh:167-168`); the port didn't.
* `LoadNanovdb.cpp` `copyIndexGridToHandle()` — set `mGridCount = 1` but left `mGridIndex` at the batch position.

`copyIndexGridToHandle` also copies the grid only up to its first blind-metadata record yet left `mBlindMetadataCount`/`Offset` claiming it, and every one of these header writes left the file's checksum stale; both are now normalized too (checksum marked disabled, as `MakeContiguous`/`ConcatenateGrids` do after their fixups).

Harmless before (the fields were overwritten later or never read); at the new pin, saving N>1 grids to a standard type fails with *"invalid host buffer (grid 1 of N)"* and loading grid i>0 of a tensor-grid file fails with *"inconsistent grid index/count (grid 0 of 1)"* — **275 of 622 `test_io.py` cases**. Nine lines, correct at the old pin too, so it leads the series and every commit passes.

## Commit 2 — pin + `-Werror` fallout (swahtz's #752, rebased)

* Pin `9df9ec06` → `31d1e596`. #752's original target `e679862f` is now *behind* `main` (which moved to `9df9ec06` in #754), so the conflict was resolved forward inside the commit to keep it building.
* `PadGrid.cuh`: #2301 makes `cuda::DeviceBuffer` a `[[deprecated]]` alias of `DualDeviceBuffer` **and removes `TopologyBuilder::mProcessedRoot` outright**. The `getHandle()` default template argument takes the non-deprecated spelling; `padRoot()` is ported to `allocateProcessedRoot()` / `uploadProcessedRoot()` mirroring upstream `DilateGrid::dilateRoot`. Net effect: the root upload is now asynchronous (pinned staging) and its device copy allocates through `BuilderResource`. A now-dead `cudaGetDevice` is removed.
* #752's three allocator improvements are carried unchanged: SaveNanoVDB staging buffers on `cuda::Buffer<…, BuilderResource>`, `TorchDeviceBuffer::create` forwarding the builder's stream to `raw_alloc_with_stream`, and the VBM adapter (removed in commit 2).

## Commit 3 — single-space buffers where NanoVDB now allows them

| site | before | after |
|---|---|---|
| `ReinitializeSdf.cu` VBM metadata | local `TorchVbmBuffer` adapter (−79 lines) | `VoxelBlockManagerHandle<cuda::Buffer<std::byte, BuilderResource>>` directly |
| `BuildPrunedGrid.cu` per-leaf mask scratch | `TorchDeviceBuffer` + two `reinterpret_cast`; `PruneGrid` on legacy stream 0 | `cuda::Buffer<Mask<3>, BuilderResource>`, typed `data()`; `PruneGrid` on the fill stream |
| `SaveNanoVDB.cu` `indexToGrid` output | `GridHandle<TorchDeviceBuffer>` via a `TorchDeviceBuffer(0, device)` guide | `GridHandle<cuda::Buffer<std::byte, BuilderResource>>` via a default-constructed pool buffer |

Why each is safe now, verified against the pinned headers rather than assumed:

* The adapter existed because `VoxelBlockManagerHandle`'s accessors were gated on `hasDeviceDual` and `buildVoxelBlockManager` allocated through `BufferT::create`. #2301 adds `BufferHasDeviceSingle` accessors and routes allocation through `createDeviceStorage(bytes, proto, …)`; `BufferTraits<cuda::Buffer<T,R>>::hasDeviceSingle` is `!is_host_accessible_resource<R>`, true for `BuilderResource`, which is default-constructible and stream-ordered — so the `nullptr` prototype yields `Buffer(stream, bytes, noInit)` on the reinit stream. `reset()` releases via `destroy()`.
* `noInit` for the mask scratch is behaviour-preserving: `TorchDeviceBuffer` never zeroed the allocation, and `InjectPredicateToMaskFunctor` writes every word (zero, then `setOnAtomic`) before `PruneGrid` reads it.
* `PruneGrid` takes a `cudaStream_t` defaulting to 0 and fvdb never passed one, so its mask reads and its two internal `cudaStreamSynchronize`s were not ordered after the fill kernel whenever torch's current stream is a non-blocking one. Pre-existing; closed here because the new comment on that block would otherwise be false.
* `IndexToGrid::getHandle` allocates through `createDeviceStorage<BufferT>(size, &pool, …)` at this pin, so a default-constructed pool supplies the resource and no device guide is needed.

With this, **no `TorchDeviceBuffer` is allocated in `SaveNanoVDB.cu`**; the remaining mentions read the grid batch's own handle and belong to the storage flip.

## Not in scope

openvdb #2319's `UnifiedBuffer` deprecation (fvdb never uses it); openvdb #2322 (open) which, after #757, touches only `PadGrid.cuh` — the one file this PR already reworks; everything under #770 steps 3–6.

## Verification

On an RTX PRO 6000 Blackwell (sm_120), torch 2.13.0 / CUDA 13.0 env built from `env/build_environment.yml` + the `test_environment.yml` delta:

* `./build.sh install gtests` — `-Werror=all-warnings` clean, zero diagnostics.
* gtests via `ctest`: **47/47 passed**.
* `pytest unit/test_sdf.py unit/test_io.py unit/test_basic_ops.py unit/test_basic_ops_single.py unit/test_prune_single_voxel.py unit/test_sliced_batch.py unit/test_inject.py unit/test_batched_topology_builder.py`: **1270 passed, 1 skipped, 18 subtests passed**.
* Without commit 1, the same run at this pin is 995 passed / **275 failed**, all in `test_io.py` — that is the regression test for the header fix.

Each commit builds on its own; commit 1 is also correct at the old pin. Note for anyone reproducing: `./build.sh ctest` currently picks up the CPM `blosc-subbuild`'s `CMakeCache.txt` (`find … -print -quit`) and reports no tests; run `ctest` in `build/<tag>/src` directly. Pre-existing, not touched here.

`clang-format` 18 / `black` 24 check: clean.

## Review

Codex-reviewed (read-only, full diff + tree) before this revision. It confirmed the `PruneGrid` stream omission and the blind-metadata/checksum staleness above — all folded in. Two further pre-existing items it raised are **not** fixed here and are noted for follow-up: `patchGridWithBlindShape` (and the CUDA export path) rewrite checksum-covered `GridData` fields after the checksum is computed, so exported standard-type files carry a stale checksum; and `TorchDeviceBuffer::create` treats a null stream as "unspecified", which makes the legacy default stream 0 indistinguishable from "use torch's current stream" — this is the design documented in the header comment (commit 2), flagged for @swahtz rather than changed.


harrism:

Revised since the first push, after a Codex review of the full diff (which should have preceded opening this — it didn't; corrected now). Two of its confirmed findings were real and are folded in:

* **`BuildPrunedGrid.cu` — `PruneGrid` was never given a stream.** Its constructor defaults to legacy stream 0, so its mask reads and its two internal `cudaStreamSynchronize`s were not ordered after the fill kernel whenever torch's current stream is non-blocking. Pre-existing, but the new comment on that block asserted same-stream ordering, so it is fixed rather than reworded: `pruneOp(grid, leafMask, stream.stream())`.
* **`LoadNanovdb.cpp` — `copyIndexGridToHandle` header now fully describes its buffer.** Beyond `mGridIndex`, it copies the grid only up to the first blind-metadata record yet left `mBlindMetadataCount`/`Offset` claiming it, and every header write left the file's checksum stale. Both cleared; checksum marked disabled per the `MakeContiguous`/`ConcatenateGrids` convention.

Two further pre-existing items from the review are **not** fixed here, deliberately, and are in the body under *Review*: stale checksums in exported standard-type files (`patchGridWithBlindShape` and the CUDA export path both rewrite checksum-covered fields after computing it), and `TorchDeviceBuffer::create`'s null-stream-means-current-stream ambiguity, which is the design documented in commit 2 — @swahtz, flagging that one to you rather than changing it.

Re-verified at the new head: `-Werror` clean, 47/47 gtests, 1270 passed / 1 skipped across the eight suites, CI-equivalent clang-format clean on all 344 `src/` files.

src/fvdb/TorchDeviceBuffer.cpp
```diff
@@ -204,15 +211,17 @@ TorchDeviceBuffer::clear() {
...
 TorchDeviceBuffer
 TorchDeviceBuffer::create(uint64_t size, const TorchDeviceBuffer *proto, int device, void *stream) {
+    // The stream a nanovdb builder passes here is the one it orders its writes into the buffer
+    // on; forward it so the allocation is associated with that stream (see the constructor).
     if (proto) {
         // This is a hack to pass in the device index when creating grids from nanovdb. Since we
         // can't pass arguments through nanovdb creation functions, we use a prototype grid to pass
         // in the device index.
-        return TorchDeviceBuffer(size, proto->device());
+        return TorchDeviceBuffer(size, proto->device(), stream);
```
swahtz, line 291: **Regression, blocking.** Forwarding `stream` here without qualification keys the allocation to whatever the caller hands in. Through `DistributedPointsToGrid` that is `mDeviceMesh[0].stream` (`HandleStorage.h` `createDeviceStorage` → `BufferT::create(bytes, proto, device, stream)`), and `BuildGridFromIjk.cu:159` / `BuildDenseGrid.cu:233` construct a loop-local `DeviceMesh` whose dtor `cudaStreamDestroy`s that stream before the GridBatch is freed. For batch size 1 this buffer *is* the returned grid storage.

c10's native allocator pools freed blocks by exact `cudaStream_t`, so:
- every freed grid block is unreusable by any torch stream → `torch.cuda.memory_reserved()` grows one segment per `gridbatch_from_ijk` / `from_dense` in a training loop until an OOM-triggered release;
- with `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`, `empty_cache()` / OOM-retry hits `ExpandableSegment::unmapHandles` → `cudaStreamSynchronize` on the destroyed handle → `cudaErrorInvalidResourceHandle`;
- `backend:cudaMallocAsync` does `cudaFreeAsync` on the dead stream.

The old pin could not hit this because `DistributedPointsToGrid.cuh:912` passed no stream and the old `create()` dropped it.

Neither call site can supply a torch stream (`DeviceMesh` has no stream-injecting ctor), so the fix belongs at this seam: resolve `stream` to the device's current torch stream unless it is one fvdb owns / torch knows about. The same question applies to the `TorchDeviceResource` design; raised on #770.

src/fvdb/detail/io/SaveNanoVDB.cu
```diff
@@ -336,6 +337,12 @@ indexToGridHost(const nanovdb::NanoGrid<nanovdb::ValueOnIndex> *srcGrid,
     *dstGrid->data()   = *srcGrid->data();
     dstGrid->mGridType = nanovdb::toGridType<DstBuildT>();
     dstGrid->mData1    = 0u;
+    // The source header describes its place in the batch and the size of the index grid. This
+    // buffer holds exactly one typed grid, and GridHandle validates that at construction
+    // (matches what tools::cuda::indexToGrid does to its output).
```
swahtz: This parenthetical is not true at `31d1e596`: `IndexToGrid.cuh:164-168` copies the whole source header and resets only `mGridType` / `mData1` / `mGridIndex` / `mGridCount`, never `mGridSize`. The host and CUDA paths therefore diverge on an empty grid in the batch (which the export loop explicitly allows): this path succeeds, the CUDA path launches `processRootTilesKernel<<<0,1>>>` unguarded and `cudaCheckError` `exit()`s the process, and even with that guarded the stale `mGridSize` would fail the new `GridHandle` size check. Non-empty grids are fine because `patchGridWithBlindShape` rewrites `mGridSize` later.

Suggest rewording to what this code actually guarantees (single-grid header, size = this buffer) and dropping the claim about upstream; the empty-grid handling and the upstream `mGridSize` reset are noted on #770 for step 4.

src/fvdb/BuilderResource.h
```diff
@@ -21,9 +21,11 @@ namespace fvdb {
 ///        build-time switch guarding the TorchResource include — instead of
 ///        touching every op.
 ///
-///        The alias covers the builders' scratch only. Buffer allocations that
-///        are torch tensors by design (TorchDeviceBuffer, the SaveNanoVDB
-///        staging buffers) name their types directly.
+///        The alias covers the builders' scratch and the device-only staging
+///        and scratch buffers feeding them (nanovdb::cuda::Buffer<...,
+///        BuilderResource> in SaveNanoVDB, ReinitializeSdf and BuildPrunedGrid).
```
swahtz: `nanovdb::cuda::Buffer<Mask<3>, BuilderResource>` is now spelled at six call sites (here, `SaveNanoVDB.cu:632/664/679/699`, `ReinitializeSdf.cu:32`). A `template <class T> using BuilderBuffer = nanovdb::cuda::Buffer<T, BuilderResource>;` beside the alias would keep the header's stated purpose ("retarget here instead of touching every op") true for the buffer type as well. Fine to defer to step 6 if you'd rather not widen this PR.

Separately, the list of files in this comment already omits the ~11 other `Build*.cu` ops that bind `BuilderResource` and will rot with the next adopter; suggest describing the category rather than enumerating.

src/fvdb/TorchDeviceBuffer.h
```diff
@@ -29,7 +29,14 @@ class TorchDeviceBuffer {
     /// nanovdb::HostBuffer
     /// @param size The size (in bytes to allocate for this buffer)
     /// @param device Specifies the device to use for the buffer
-    TorchDeviceBuffer(uint64_t size = 0, const torch::Device &device = torch::kCPU);
+    /// @param stream For a CUDA device, the stream the allocation is associated with in torch's
+    /// caching allocator (raw_alloc_with_stream): torch defers reusing the block until work queued
+    /// on that stream at free time completes, so pass the stream the buffer is used on. Null (the
+    /// default, indistinguishable from the legacy default stream) associates the allocation with
```
swahtz: Two nits on the wording, worth getting right because `TorchDeviceResource` will inherit this text:

- "torch defers reusing the block until work queued on that stream at free time completes" describes the `cudaMallocAsync` backend. The native allocator pools freed blocks per stream and hands them straight back to same-stream allocations (`free_block`, `CUDACachingAllocator.cpp:1999`); only `recordStream`-tagged blocks get events. Suggest "reuse is restricted to allocations on the same stream".
- The "indistinguishable from the legacy default stream" caveat disappears if the `.cpp` resolves the stream once (`stream ? stream : getCurrentCUDAStream(idx).stream()`) and always calls `raw_alloc_with_stream`; `raw_alloc(size)` is exactly that on all three c10 backends.

swahtz (commented):
Review of steps 1–2 against the pinned headers at `31d1e596` and the local c10 source. One regression that should be fixed before merge (inline on `TorchDeviceBuffer.cpp`), one comment that mis-states an upstream invariant, and cleanups. Plan-level items that belong to later steps are on #770 rather than here.

Two notes on lines outside the diff hunks:

- **`BuildPrunedGrid.cu:103`** — `nanovdb::cuda::mergeGridHandles(handles, &guide)` still runs on legacy stream 0 while this PR is making the rest of the file stream-correct (the new comment on the `PruneGrid` call describes exactly this read-before-write class). Suggest passing `stream.stream()` here since the file is already open; the other eight `mergeGridHandles` sites and three `voxelsToGrid` sites are step 5's.
- **`SaveNanoVDB.cu:11,16`** — `<nanovdb/cuda/DeviceBuffer.h>` and `<c10/cuda/CUDACachingAllocator.h>` have no remaining uses in the file; the diff added `<nanovdb/cuda/Buffer.h>` next to `DeviceBuffer.h` rather than replacing it.

Not raised: the unreachable H2D fallback in `fvdbToNanovdbGridWithValues` (rewritten in step 4) and the `raw_alloc` ternary (class is deprecated in step 6).

harrism:

CI is 41/41 on `d6e9300`. For the record, the pip CUDA 13.2 workflow shows three attempts: attempt 1 failed only `test_sample.py::test_trilinear_dense_vs_pytorch_05_cuda` (fp16 × 16ch, `Max grad error 0.1875`) — a pre-existing flake unrelated to this PR (its backward accumulates with `gpuAtomicAddNoReturn`, so seeded inputs still see order-dependent fp16 rounding; it passed 8/8 locally on this exact code and on the previous head's identical job). Attempt 2 was a "re-run failed jobs" that can never start on our ephemeral runners (noted on #761); attempt 3 is the full rerun, all green.

swahtz:
> > _Analysis by Claude Code, run by @harrism._
> 
> CI is 41/41 on `d6e9300`. For the record, the pip CUDA 13.2 workflow shows three attempts: attempt 1 failed only `test_sample.py::test_trilinear_dense_vs_pytorch_05_cuda` (fp16 × 16ch, `Max grad error 0.1875`) — a pre-existing flake unrelated to this PR (its backward accumulates with `gpuAtomicAddNoReturn`, so seeded inputs still see order-dependent fp16 rounding; it passed 8/8 locally on this exact code and on the previous head's identical job). Attempt 2 was a "re-run failed jobs" that can never start on our ephemeral runners (noted on #761); attempt 3 is the full rerun, all green.

I also noted that flaky test.  I have a fix for it in the #753 PR because I hit it there too.

src/fvdb/TorchDeviceBuffer.cpp
harrism: Agreed on the mechanism, and thank you for tracing it through c10. Confirmed against the pinned headers and pytorch 2.13 source: at `31d1e596` the distributed builder creates its output storage via `createDeviceStorage` with `mDeviceMesh[0].stream`, `create()` forwarded that to `raw_alloc_with_stream`, and in the native allocator the allocation stream is the free-list key (`BlockComparatorSizeCounterAddress` compares it first), so a block keyed to a destroyed mesh stream is never reused, `ExpandableSegment::unmapHandles` would synchronize on the dead handle, and the cudaMallocAsync backend would `cudaFreeAsync` on it.

One scoping note: both mesh-per-item sites are the `kPrivateUse1` dispatches. The `kCUDA` dispatches use `voxelsToGrid` on the legacy default stream, which resolves to torch's current stream, and nothing in fvdb-core registers an allocator or hooks for the PrivateUse1 device, so the path is not reachable from this package today. A reserved-memory probe over repeated `gridbatch_from_ijk` builds shows zero growth at the previous head for that reason. Fixing it anyway, because the storage resource in #770 step 3 would inherit the same policy and the plan routes the distributed builder through it with mesh streams.

Fixed at the resource, not the buffer, in d4d4cbf: `TorchResource::allocate_async` now always allocates on the current torch stream and, when the caller's stream differs, records an event after the allocation and makes the caller's stream wait on it. That is the rule you proposed, with the ordering half that keeps the builder's writes behind the block's previous tenant. `TorchDeviceBuffer`'s CUDA paths allocate and free through `TorchResource` instead of calling c10 directly, so there is one allocation policy in the codebase and the step-3 resource inherits the correct one. This settles #770 item 1 toward the second option: torch's stream is the key, every other stream is an ordering hint. It does mean an explicitly passed foreign stream is not honoured as the allocation stream, which is deliberate given the c10 semantics and is now documented on both classes.

Re-verified: `-Werror` clean, 47/47 gtests, 1605 passed / 1 skipped across nine suites.

src/fvdb/detail/io/SaveNanoVDB.cu
harrism: Right, thank you. At `31d1e596` `IndexToGrid.cuh` resets only `mGridType`/`mData1`/`mGridIndex`/`mGridCount`; the parenthetical was wrong. Reworded in 6e8420c to what this path guarantees: one grid, header index/count/size match this buffer. The empty-grid divergence (unguarded `<<<0,1>>>` launch, stale `mGridSize`) stays on #770 for step 4 as you noted.

src/fvdb/BuilderResource.h
harrism: Both taken in 6e8420c. `template <class T> using BuilderBuffer = nanovdb::cuda::Buffer<T, BuilderResource>;` sits beside the alias, and the six spelled-out sites (SaveNanoVDB ×4, ReinitializeSdf, BuildPrunedGrid) use it, so retargeting the resource retargets the ops' own staging and scratch buffers as well. The header comment now describes the category (device-only allocations whose lifetime is an op) rather than listing files. Step 6 on #770 can drop its `BuilderBuffer` item.

src/fvdb/TorchDeviceBuffer.h
harrism: Agreed on both. The wording moved to `TorchResource::allocate_async` in d4d4cbf and now says what the native allocator does: the allocation stream is the free-list key, reuse is restricted to same-stream allocations, freeing does no synchronization. On the second point, 6e8420c goes one step further than resolving null to the current stream: null is treated as what it is, the legacy default stream, and the resource orders it after the allocation like any other write stream (a no-op when torch's current stream is also the default). So there is no longer an "unspecified" case to document, and the synchronous `cudaMemcpy` paths in `to()` become correctly ordered behind the allocation when a non-blocking torch stream is current, which `raw_alloc` alone did not give them.

swahtz (approved):
Nice, thanks for addressing my notes.  Looks good to go.

harrism:
Follow-up 6f605b9, after a codex review of d4d4cbf and 6e8420c that should have preceded pushing them. Its main finding was a race d4d4cbf introduced: keying *scratch* to torch's current stream is wrong, because `TorchResource::deallocate_async` does not order the free and the pinned `PointsToGrid` destroys its per-tile counts right behind an in-flight scan on legacy stream 0. The block has to be keyed to the stream its work runs on. So the two policies are now two resources:

- `TorchResource` (scratch, `BuilderResource`): back to `raw_alloc_with_stream(bytes, stream)`, with the comment stating why that is load-bearing and that the stream must outlive the block.
- `TorchStorageResource` (grid storage): allocate on the current torch stream, event-order the caller's writer stream after it, free the block if any event call fails. `TorchDeviceBuffer` uses this one. @swahtz, this is the body I'd suggest `TorchDeviceResource` in #770 step 3 adopts; it settles item 1 as before, just no longer at the expense of scratch.

Also in the commit, all surfaced by the same review: `TorchDeviceBuffer::to()`'s CUDA copies move off the legacy default stream onto the writing device's torch stream, are synchronized and error-checked, use `cudaMemcpyPeerAsync` across devices, and hold the destination in a scope guard so a failed copy neither leaks nor half-moves the buffer (all pre-existing). `BuilderResource.h` no longer claims to cover ops that still take CUB scratch from c10 directly. And a correction to d4d4cbf's message: `BuildDenseGrid`'s PrivateUse1 dispatch builds one mesh for the first item and copies for the rest; only `BuildGridFromIjk` builds one per item.

Five codex rounds to clean. Re-verified at 6f605b9: `-Werror` clean, 47/47 gtests, 1605 passed / 1 skipped across nine suites, clang-format clean.

swahtz:
[P3] Preserve zero-byte buffer transfers to CUDA — [src/fvdb/TorchDeviceBuffer.cpp:177](https://github.com/openvdb/fvdb-core/blob/6f605b9dd969130af9c01a30999fbdf81611b7a9/src/fvdb/TorchDeviceBuffer.cpp#L177)

The new allocation path changes the behavior of an empty C++ buffer:

```cpp
fvdb::TorchDeviceBuffer buf;
buf.to(torch::Device(torch::kCUDA, 0));
```

Previously this succeeded: the native allocator returns a null pointer for a zero-byte allocation, and the zero-byte H2D `cudaMemcpy` succeeds. Now `deviceAlloc(0)` reaches `TorchResource::allocate_async`, whose null-pointer check throws `fvdb: TorchResource::allocate_async failed`. The new CUDA-to-CUDA allocation path has the same issue for zero-sized buffers.

Please handle zero-sized transfers without allocating, while preserving the requested destination device, and add a focused buffer-level regression test. This concerns zero-byte `TorchDeviceBuffer` objects, not empty voxel grids that still have an allocated NanoVDB header.

Validation: confirmed the native allocator's null result and the old zero-byte H2D copy's success with a local CUDA probe; the new exception follows from the allocation call chain at `6f605b9`. I did not rebuild the updated PR.

harrism:
Confirmed, thank you: `to()` reached the storage resource's null check for a zero-byte request. Fixed in 8a77387: `to()` returns early for size 0, recording the requested device and nothing else, and that now precedes the same-device check, which also used to reject an empty buffer moved to the device it was already on.

New `TorchDeviceBufferTest` (four gtests): the empty moves, plus two that were worth writing while here. Each of those two was rebuilt until a mutation run proved it earns its keep: with the CPU-to-CUDA synchronization removed from `to()` and the allocation-to-writer event removed from `TorchStorageResource`, the two tests fail on 3 of 3 runs; with the source restored all four pass on 3 of 3, 48/48 gtests, 1605 pytest. The one residual is a host thread descheduled for longer than the 0.25 s spin at exactly the wrong moment, which turns a would-be failure into a pass, never the reverse; it is documented on the constant.

Issue #775, opened 2026-09-12
Batch the remaining per-member CUDA topology builders


swahtz:
#757 added a batched leaf-mask topology builder (`src/fvdb/detail/utils/nanovdb/BatchedTopologyBuilder.cuh`) and moved coarsen, refine, and the K==S conv / conv-transpose grid builds onto it; #753 added a `Prune` pass for `pruned_grid`. Each of those replaced a CUDA path that built one NanoVDB grid per batch member and then `mergeGridHandles`'d them. That pattern costs about 2 ms per member independent of grid size (several stream syncs per `getHandle`, one more per grid in the merge). For `pruned_grid` at batch 16 the change was 30.7 ms -> 0.7 ms, flat in batch size.

An audit of `src/fvdb/detail/ops/Build*.cu` finds seven CUDA builders still on the per-member loop. This issue tracks moving them over (items 1-7), plus two builder-level follow-ups that surfaced during implementation (items 8-9).

## Still per-member on CUDA

| # | op | file | today | batched approach |
|---|---|---|---|---|
| 1 | `dual_grid`, `padded_grid` (positive pads) | `BuildPaddedGrid.cu` | per-member pad pass per unit | existing BoxDilate pass, currently not wired up |
| 2 | `from_ijk` (+ `from_mesh`, nearest-voxels, all coordinate-list fallbacks; `from_points` had its own per-member `voxelsToGrid` loop and is folded in) | `BuildGridFromIjk.cu`, `BuildGridFromPoints.cu` | `joffsets().cpu()` then `PointsToGrid` per member | new `Coords` emission (slot = ijk, one-bit mask, OR dedup) |
| 3 | `dilated_grid` | `BuildDilatedGrid.cu` | `DilateGrid` x N per member | N chained BoxDilate(-1,+1) passes |
| 4 | `from_dense` | `BuildDenseGrid.cu` | one build, then a synchronous handle copy per member and a per-grid sync inside `mergeGridHandles` | replicate the buffer with log2(B) async copies and one header fix-up kernel |
| 5 | `merged_grid` | `BuildMergedGrids.cu` | `MergeGrids` per member | new `Union` emission from two source arrays, OR dedup |
| 6 | `from_nearest_voxels_to_points` (pad step) | `BuildGridFromNearestVoxelsToPoints.cu` | `PadGrid` per member after from_ijk | BoxDilate(0,+1) over the batched from_ijk result |
[...]

PR #779, opened 2026-09-12
Batched Erode passes for exclude_border / negative padded_grid on CUDA (issue 775 item 7)

swahtz:
## Summary

Stacked on PR #776 (item 1); only the second commit is new.

After PR #776, `dispatchBuildPaddedGrid<torch::kCUDA>` still looped over batch members for erosion (`dual_grid(exclude_border=True)`, `build_padded_grid` with `exclude_border`) and for negative padding: `erodeOncePass` computed a per-leaf keep-mask sidecar, tested it for emptiness with a `(keepTensor != 0).any().item<bool>()` host sync, ran NanoVDB `PruneGrid` per member, and `padOncePass` ran one `PadGrid` per member per unit of padding, all merged with `nanovdb::cuda::mergeGridHandles`. Wall clock grew linearly with batch size regardless of topology size.

The whole `[bmin, bmax]` family is now one chain of batched passes through `batched::batchedTopologyHandle`: `bmax` positive-octant passes followed by `-bmin` negative-octant passes, `BoxDilate` for plain padding (Minkowski sum) and the new `Erode` op for `exclude_border`. One output buffer, one stream synchronization per pass, no per-member builds, no handle merging, no host-side emptiness check: members eroded to nothing come out as valid empty grids inline. The identity branch (`bmin == bmax == 0`) is unchanged; CPU and PrivateUse1 paths are untouched.

## Approach

`TopologyPassSpec::Op::Erode` in `BatchedTopologyBuilder.cuh` carries the same `boxLo`/`boxHi` (components in {-1,0} / {0,1}) as `BoxDilate`, with the dual semantics: an output voxel is active iff every box offset of it is active in the source. Emission (`emitErodedLeaves`) is injective -- slot = source leaf, the eroded leaf is a subset of the source leaf -- so `combineDuplicateMasks` is skipped like for Refine. The AND over the box factorizes per axis, so the kernel erodes along z, then y, then x with the existing `shiftWordZ` / `shiftWordY` helpers and the x-stage word permutation from `emitBoxDilatedLeaves`, combining with AND from all-ones instead of OR from zero. Bits that cross a leaf border come from the neighbor leaf on that side (fetched through the source grid's root; a missing neighbor contributes zeros and erodes the border voxels away). Because the y-stage reads z-eroded words of the +-y neighbors and the x-stage reads (z,y)-eroded words of the +-x neighbors, the earlier stages run on every neighbor leaf the box reaches (up to 2^3 for a one-sided octant, 3^3 for the full box). Leaves eroded to nothing are dead slots.

One correction to the item-7 wording: negative padding *without* `exclude_border` is a Minkowski sum with `{-1,0}^3` (the old `padOncePass(positive=false)`, and the CPU `buildPaddedGridFromGridCPU`), not an erosion, so it is wired as `BoxDilate(-1, 0)` passes; only `exclude_border` uses `Erode`. The new tests pin both against the CPU path.

The dead `padOncePass` / `erodeOncePass` helpers and their includes (`PadGrid.cuh`, `PruneGrid.cuh`, `DeviceGridTraits.cuh`, `CreateEmptyGridHandle.h`, `BuilderResource.h`) are removed from `BuildPaddedGrid.cu`. `ErodeKeepMaskFunctor` in `PadGrid.cuh` is now unused but left in place (library helper, still documented alongside `PadLeafNodesFunctor`).

## Benchmark

~83k voxels per member (`from_ijk` of 100k random ints in [0,64)^3), 20 iterations after one warmup, `torch.cuda.synchronize()` around the loop, RTX PRO 6000 Blackwell. Before = the installed env (per-member erode / pad + merge); after = this branch.

| op | B | before (ms) | after (ms) |
|---|---|---|---|
| `dual_grid(exclude_border=True)` | 1 | 0.59 | 0.56 |
| | 4 | 2.45 | 0.60 |
| | 16 | 9.96 | 0.73 |
| | 64 | 38.55 | 1.20 |
| `build_padded_grid(-1, 0)` | 1 | 0.66 | 0.56 |
| | 4 | 2.80 | 0.60 |
| | 16 | 10.42 | 0.73 |
| | 64 | 39.34 | 1.34 |

## Tests

New in `tests/unit/test_batched_topology_builder.py`, over `_tricky_batches()` (empty members, all-empty, +-4096 tile boundaries, 16 members) plus new dense batches (`_dense_batches()`: near-solid blobs straddling leaf and +-4096 tile boundaries in both octants, thin slabs) so that interior voxels actually survive erosion:

- `test_dual_grid_exclude_border_matches_expected_and_cpu`: `dual_grid(exclude_border=True)` pinned elementwise (`ijk.jdata`, `num_voxels`, per-member bbox) against a `from_ijk` build of the expected erosion computed in torch (a voxel survives iff all its box neighbors are in the set, via `torch.isin` on packed coordinate keys), as coordinate sets against CPU, plus the dual-transform origin fix-up.
- `test_padded_grid_negative_bounds_match_expected_and_cpu`: `build_padded_grid` for `(bmin, bmax)` in `(-1,0), (-1,1), (-2,0)`, plain (expected Minkowski sum with `[bmin,bmax]^3`) and `exclude_border` (expected erosion), vs expected and vs CPU.
- `test_erode_shell_to_empty_and_solid_block_interior`: one-voxel-thick shells (including one straddling 4096) erode to empty members with `grid_count` preserved; solid blocks `[lo,hi)^3` straddling leaf and tile boundaries in both octants erode to `[lo, hi-1)`, `[lo+1, hi-1)`, `[lo+2, hi)` for the three boxes, with the bbox checked per member and an empty member alongside.

The torch reference itself was cross-checked against the (unchanged) CPU implementation for every batch and box before pinning the CUDA path to it.

Results with the wheel from this branch: `test_batched_topology_builder.py` 51 passed; `test_dual.py` + `test_batching.py` + `test_sliced_batch.py` 74 passed; `test_basic_ops.py` + `test_basic_ops_single.py` 430 passed, 1 skipped.

Part of #775 (item 7).

### DIFF
--- a/src/fvdb/detail/ops/BuildPaddedGrid.cu
+++ b/src/fvdb/detail/ops/BuildPaddedGrid.cu
@@ -1,7 +1,6 @@
 // Copyright Contributors to the OpenVDB Project
 // SPDX-License-Identifier: Apache-2.0
 //
-#include <fvdb/BuilderResource.h>
 #include <fvdb/GridBatchData.h>
 #include <fvdb/TorchDeviceBuffer.h>
 #include <fvdb/detail/GridBatchDataFactory.h>
@@ -15,14 +14,9 @@
 #include <fvdb/detail/utils/cuda/ForEachPrivateUse1.cuh>
 #include <fvdb/detail/utils/cuda/GridDim.h>
 #include <fvdb/detail/utils/nanovdb/BatchedTopologyBuilder.cuh>
-#include <fvdb/detail/utils/nanovdb/CreateEmptyGridHandle.h>
-#include <fvdb/detail/utils/nanovdb/PadGrid.cuh>
 
 #include <nanovdb/NanoVDB.h>
 #include <nanovdb/tools/CreateNanoGrid.h>
-#include <nanovdb/tools/cuda/PruneGrid.cuh>
-#include <nanovdb/util/cuda/DeviceGridTraits.cuh>
-#include <nanovdb/util/cuda/Util.h>
 
 #include <c10/cuda/CUDACachingAllocator.h>
 #include <c10/cuda/CUDAGuard.h>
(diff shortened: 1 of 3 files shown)


swahtz:

PR #780, opened 2026-09-12
Batched from_ijk / from_points via a Coords topology pass (issue 775 item 2)

swahtz:
## Summary

`GridBatch.from_ijk` on CUDA (and everything that routes through `_createNanoGridFromIJK`: `from_mesh`, `from_nearest_voxels_to_points`, non-power-of-two `conv_grid` / `coarsened_grid` fallbacks) copied `joffsets` to the host, ran `nanovdb::tools::cuda::voxelsToGrid` (PointsToGrid, several stream syncs) once per batch member, then `mergeGridHandles`'d the results (one more sync per grid). This PR builds the whole batch in one pass of the batched leaf-mask topology builder via a new `Coords` emission, and routes the CUDA `from_points` dispatch (its own per-member PointsToGrid loop over a transformed-point adaptor) through the same batched path. Time is now flat in batch size; the single-grid case got faster too.

## Approach

**`TopologyPassSpec::Op::Coords`** (`BatchedTopologyBuilder.cuh`): carries `const int32_t *ijk` (N x 3 contiguous) and `const int64_t *offsets` (the B+1 `joffsets`, device). Slot = coordinate. `emitCoordLeaves` writes the tile / node sort keys of the coordinate's 8-aligned leaf origin; instead of a 512-bit mask per slot it stores the full coordinate in the origin array, and `reduceLeafMasks` recovers the voxel bit (`LeafNode::CoordToOffset`: word = local x, byte = local y, bit = local z) and atomically ORs it into the leaf. Duplicate coordinates and coordinates sharing a leaf therefore combine downstream like Coarsen / BoxDilate (`passHasDuplicateProducers`).

**Caller-supplied slot counts** (for items 5 and 6 to build on): `BatchedTopologySource` gains `std::vector<int64_t> slotCounts`. When non-empty it replaces `leafCounts[g] * fanout` as the per-grid emission slot count and `leafCounts` is not consulted. `grids` may be empty for passes that do not read source grids (only Coords today); the batch size is then `slotCounts.size()`. `sourceFromCoordCounts(coordCounts, device)` builds that source; the counts come from the host `joffsets` copy the int32 candidate-count `TORCH_CHECK` already makes, so no new sync.

**Header seeding without a source grid:** `copyGridData` needs a source header per grid. With `grids` empty the driver runs `initDefaultGridData` instead: `GridData::init({IsBreadthFirst}, gridSize, Map(), toGridType<ValueOnIndex>(), GridClass::IndexGrid)` per grid, i.e. the header PointsToGrid's `BuildGridTreeRootFunctor` writes (identity map, voxel size 1, empty name, no blind data), before `initGridTreeRoot` / `finalizeGrids` set index/count/size/flags/`mData1`/bboxes as for every other pass. `makeGridBatchData` callers keep per-grid transforms in `GridBatchData` metadata, so the map is never read for these grids beyond `mVoxelSize`.

**Driver changes that apply to every pass (results unchanged, verified by the existing tests):**
- The two `cub::DeviceSegmentedRadixSort` sorts became three stable global `cub::DeviceRadixSort` sorts (node key, tile key, grid index; the last skipped for B = 1). The segmented sort assigns one thread block per segment, so a single 1M-coordinate grid sorted on one SM (16 ms; 62 ms at 4M). Global sorts make B = 1 flat in N (0.9 ms up to 1M coordinates).
- Scratch is released stage by stage, the per-node tables (`leafHeadSlot`, `leafParent`, ...) are allocated at unique-node size after the sizing readback instead of at slot count, and leaf masks are reduced into a per-leaf table (`reduceLeafMasks`, replacing `combineDuplicateMasks` + the sorted mask copy). Per-slot scratch drops from ~240 B to ~120 B for leaf-based passes; `from_ijk` on 2M coordinates peaks at 148 MiB of torch-visible scratch (PointsToGrid: 84 MiB), which keeps `test_nearest_voxels_to_points_peak_memory` under its 200 MiB bound (171 MiB, was 159 MiB).

`from_points` (CUDA): materializes the rounded index-space coordinates with the existing `ijkForPointsCallback` (12 B/point, as the PrivateUse1 path already does) and calls `_createNanoGridFromIJK`; the `TransformedPointPtr` adaptor is removed. CPU and PrivateUse1 dispatches of both ops are unchanged. The per-grid int32 candidate-count check is preserved (the Coords pass indexes slots with int32 as well).

## Benchmark (RTX PRO 6000 Blackwell, ~100k random int32 coords per member, 20 iterations after warmup, `torch.cuda.synchronize()` around the loop, GPU lock held)

| op | B | before (ms) | after (ms) |
|---|---|---|---|
| `from_ijk` | 1 | 1.22 | 0.91 |
| `from_ijk` | 4 | 3.97 | 0.90 |
| `from_ijk` | 16 | 14.61 | 1.07 |
| `from_ijk` | 64 | 56.06 | 2.83 |
| `from_points` | 16 | 14.31 | 1.12 |

Single grid, varying N (before -> after): 100k 1.20 -> 0.84 ms, 1M 1.28 -> 0.89 ms, 4M 1.70 -> 1.76 ms.

## Tests

`tests/unit/test_batched_topology_builder.py` (33 passed):
- `test_from_ijk_matches_cpu_per_member_and_canonical_order` over `_tricky_batches()` plus three new batches (heavy duplicate coordinates including a member of one coordinate repeated 1000x, coordinates straddling the +-4096 / +-8192 root-tile boundaries with negative components, int64 input): coordinate sets / counts / bboxes vs the CPU build, unique-count check, multi-member == single-member elementwise, idempotence (`from_ijk(result.ijk) == result`), and hand-written members pinned elementwise to the canonical order the old PointsToGrid path produced (`_FROM_IJK_CANONICAL_ORDER`, captured from the pre-change CUDA build).
- `test_from_points_multi_member_matches_per_member_and_cpu` (also `from_nearest_voxels_to_points`), `test_from_mesh_multi_member_matches_per_member_and_cpu` (bunny example mesh, skipped if unavailable), `test_conv_grid_k3s3_multi_member_matches_per_member_and_cpu` (coordinate-list fallback).
- All existing builder tests (refine / coarsen / conv / conv-transpose pinned against `from_ijk`) pass on the new build, which pins the new `from_ijk` ordering transitively.

Additionally, before switching, `ijk.jdata`, `joffsets`, `num_voxels`, `num_leaf_nodes`, per-member and total bboxes of the old CUDA `from_ijk` were dumped for all nine batches (multi-member and each member alone) and compared elementwise against the new build: 0 mismatches.

Existing suites run with the new build: `test_basic_ops.py`, `test_basic_ops_single.py`, `test_sliced_batch.py`, `test_batching.py`, `test_empty_grids.py`, `test_dense_interface.py`, `test_accessors.py`, `test_io.py`, `test_inject.py`, `test_jagged_tensor.py`, `test_nn_modules.py`: 1875 passed, 6 skipped.

Part of #775 (item 2).

### DIFF
--- a/src/fvdb/detail/ops/BuildGridFromIjk.cu
+++ b/src/fvdb/detail/ops/BuildGridFromIjk.cu
@@ -1,6 +1,7 @@
 // Copyright Contributors to the OpenVDB Project
 // SPDX-License-Identifier: Apache-2.0
 //
+#include <fvdb/BuilderResource.h>
 #include <fvdb/GridBatchData.h>
 #include <fvdb/detail/GridBatchDataFactory.h>
 #include <fvdb/detail/ops/BuildGridFromIjk.h>
@@ -34,7 +35,8 @@ nanovdb::GridHandle<TorchDeviceBuffer> dispatchCreateNanoGridFromIJK(const Jagge
 // NanoVDB's PointsToGrid / DistributedPointsToGrid radix sort casts the per-grid coordinate count
 // to int32 (PointsToGrid.cuh:645), which silently corrupts the grid above 2^31 candidates, and the
 // batched Coords pass indexes its emission slots (one per coordinate) with int32 as well. Reject
-// that here rather than return a garbage grid. Takes an already-on-host joffsets accessor so each
+// that here (on every CUDA path: single-member PointsToGrid and multi-member Coords) rather than
+// return a garbage grid. Takes an already-on-host joffsets accessor so each
 // dispatch reuses the host copy it makes anyway -- no extra device sync.
 static void
 checkCandidateCountsFitInt32(const torch::TensorAccessor<fvdb::JOffsetsType, 1> &joffsetsAcc) {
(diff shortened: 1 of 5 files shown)


swahtz:

PR #781, opened 2026-09-12
Batched merged_grid via a Union topology pass (issue 775 item 5)

swahtz:
## Summary

`GridBatch.merged_grid` on CUDA ran `nanovdb::tools::cuda::MergeGrids` once per batch member (several stream synchronizations per `getHandle`) and then `nanovdb::cuda::mergeGridHandles`, so its cost grew linearly in the batch size regardless of member size. It now builds the whole batch in one pass of the batched leaf-mask topology builder (`BatchedTopologyBuilder.cuh`) through a new `TopologyPassSpec::Op::Union` emission.

**Stacked on PR 780 at `edc1e5a0` (its current tip, including the single-member PointsToGrid fallback); only the last commit is new.** This branch is based on `batched/from-ijk` (item 2) because the Union pass uses that PR's caller-supplied per-grid slot counts (`BatchedTopologySource::slotCounts`); land PR 780 first and this PR's diff collapses to the last commit.

## Approach

- `BatchedTopologySource` gains an optional second batch (`gridsB` / `leafCountsB`); `sourceFromGridBatchPair(batchA, batchB)` fills it and sets `slotCounts[g] = leafCounts[g] + leafCountsB[g]`. `grids` stays batch A, so `copyGridData` seeds each output header from A (voxel sizes / origins are set from the first operand by `makeGridBatchData`, as before).
- New `emitUnionLeaves` kernel: slot with local index `i < leafCountsA[g]` reads leaf `i` of A[g], otherwise leaf `i - leafCountsA[g]` of B[g]; it copies the leaf origin and full 512-bit activity mask unchanged. Union is registered as non-injective in `passHasDuplicateProducers`, so a leaf present on both sides yields two slots with the same node key that `reduceLeafMasks` OR-combines; leaves unique to one side pass through.
- `dispatchMergeGrids<torch::kCUDA>` becomes one `runBatchedTopologyPass(sourceFromGridBatchPair(...), TopologyPassSpec::unionOf(), stream)`. Same-device / same-batch-size validation is kept. Members empty on one side pass the other side through; members empty on both sides become valid empty grids inline. One stream synchronization total (the sizing readback).
- CPU and PrivateUse1 paths unchanged. The Python binding's `batchSize() == 0` early return is unchanged.

## Benchmark

`a.merged_grid(b)`, ~100k voxels per member on each side (random `from_ijk` over a 128^3 box, B shifted so the union is ~194k voxels/member), RTX PRO 6000 Blackwell, 20 iterations after warmup, `torch.cuda.synchronize()` before/after, under `flock`:

| B | before (per-member MergeGrids) | after (batched Union) |
|---|---:|---:|
| 1 | 0.70 ms | 0.71 ms |
| 4 | 2.58 ms | 0.72 ms |
| 16 | 9.19 ms | 0.81 ms |
| 64 | 35.96 ms | 1.29 ms |

## Tests

Added to `tests/unit/test_batched_topology_builder.py`:

- `test_merged_grid_matches_expected_and_cpu` (16 parameterized pairs): every `_tricky_batches()` batch unioned with a shifted copy of itself (partial overlap) and with a permuted copy (identical sets, every leaf coincident), plus hand-built disjoint, identical, partially overlapping (including across the -4096 root-tile boundary) and one-side-empty pairs (A empty / B empty / both empty members). Each result is pinned elementwise (`_check_against_expected`) against `from_ijk` of `torch.unique(torch.cat([a, b]), dim=0)`, compared against the CPU `merged_grid` (`_check_against_cpu`), checked for symmetry (`b.merged_grid(a)` equal elementwise) and for voxel sizes / origins coming from the first operand.
- `test_merged_grid_on_sliced_views`: `merged_grid` on sliced / index-selected `GridBatch` views (non-contiguous member pointers) equals the contiguous build.

Results with the wheel from this branch: `test_batched_topology_builder.py` 50 passed; `test_basic_ops_single.py` (existing `merged_grid` tests) 154 passed; `test_basic_ops.py` (exercises `ops::mergeGrids` internally via `integrate_tsdf`) 276 passed, 1 skipped. `clang-format --dry-run --Werror` and `black --line-length 120` clean.

Part of #775 (item 5).

### DIFF
--- a/src/fvdb/detail/ops/BuildGridFromIjk.cu
+++ b/src/fvdb/detail/ops/BuildGridFromIjk.cu
@@ -1,6 +1,7 @@
 // Copyright Contributors to the OpenVDB Project
 // SPDX-License-Identifier: Apache-2.0
 //
+#include <fvdb/BuilderResource.h>
 #include <fvdb/GridBatchData.h>
 #include <fvdb/detail/GridBatchDataFactory.h>
 #include <fvdb/detail/ops/BuildGridFromIjk.h>
@@ -34,7 +35,8 @@ nanovdb::GridHandle<TorchDeviceBuffer> dispatchCreateNanoGridFromIJK(const Jagge
 // NanoVDB's PointsToGrid / DistributedPointsToGrid radix sort casts the per-grid coordinate count
 // to int32 (PointsToGrid.cuh:645), which silently corrupts the grid above 2^31 candidates, and the
 // batched Coords pass indexes its emission slots (one per coordinate) with int32 as well. Reject
-// that here rather than return a garbage grid. Takes an already-on-host joffsets accessor so each
+// that here (on every CUDA path: single-member PointsToGrid and multi-member Coords) rather than
+// return a garbage grid. Takes an already-on-host joffsets accessor so each
 // dispatch reuses the host copy it makes anyway -- no extra device sync.
 static void
 checkCandidateCountsFitInt32(const torch::TensorAccessor<fvdb::JOffsetsType, 1> &joffsetsAcc) {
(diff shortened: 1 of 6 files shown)


swahtz:

PR #782, opened 2026-09-12
Batched {0,1}^3 pad for from_nearest_voxels_to_points via a chained BoxDilate pass (issue 775 item 6)

swahtz:
Stacked on PR 780 (`batched/from-ijk` at edc1e5a0, item 2); only the last commit is new.

## Summary

`GridBatch.from_nearest_voxels_to_points` on CUDA built the floored-ijk base grid batch (one voxel per point) and then padded it by the positive octant {0,1}^3 -- the 8 voxels nearest to each point -- with a per-member loop: copy `points.joffsets()` to the host, run `morphology::PadGrid` once per member (several stream syncs per `getHandle`), create an empty grid handle for members without points, then `nanovdb::cuda::mergeGridHandles` (another sync per grid). With PR 780 the base grid is already a batched Coords pass result; this PR replaces the pad loop with one batched `BoxDilate(lo=0, hi=+1)` pass chained directly onto that result, so the whole op is two batched passes with one sizing synchronization each and no per-member work. The host copy of the jagged offsets in this file is gone (the Coords pass keeps its own single host copy for the int32 guard and slot counts).

## Approach

- `BuildGridFromNearestVoxelsToPoints.cu` (CUDA dispatch): `flooredIjkForPoints` -> `ops::batchedCoordsPassFromIJK` (PR 780: the batched Coords pass for B >= 2, or the single-member PointsToGrid handle wrapped by `batched::resultFromGridHandle` for B = 1) -> `runBatchedTopologyPass(sourceFromResult(base), TopologyPassSpec::boxDilate(Coord(0), Coord(1)))` -> `GridHandle`. Semantics match the CPU implementation in the same file (floor(transform(p)) + {0,1}^3 per point; the Minkowski sum distributes over the union, and `PadGrid(positiveOctant=true)` was exactly the {0,1}^3 pad). Members without points are empty base grids and stay empty through the dilate pass (zero emission slots). CPU and PrivateUse1 paths untouched; `PadGrid.cuh`, `CreateEmptyGridHandle.h`, `PointsToGrid.cuh` and `BuilderResource.h` are no longer included here.
- No builder or from_ijk changes in this PR; `batchedCoordsPassFromIJK` and `resultFromGridHandle` come from PR 780. This PR is the first runtime exercise of that B = 1 wrapping branch (nothing on PR 780 chains a pass onto a single-member result); the single-member batches below (1500 points, one point, no points) cover it.

## Benchmark

RTX PRO 6000 Blackwell, 100k random float32 points per member, `voxel_sizes=0.1`, 20 iterations after one warmup, `torch.cuda.synchronize()` around the loop, GPU lock held. "main" is the installed env (per-member PointsToGrid + per-member PadGrid); "PR 780" is this branch's base commit alone (batched from_ijk, per-member PadGrid); "this PR" adds the batched pad.

Points in a 40^3 world box (sparse: ~800k voxels per member spread over ~69k base leaves / ~86k output leaves, so the dilate pass emits ~550k slots per member and at large B the time is real per-leaf work, ~0.45 ms per member):

| B | main (ms) | PR 780 (ms) | this PR (ms) |
|---|---|---|---|
| 1 | 2.16 | 2.05 | 1.89 |
| 4 | 7.47 | 5.31 | 1.92 |
| 16 | 28.79 | 18.39 | 7.22 |
| 64 | 115.85 | 71.01 | 27.78 |

Points in an 8^3 world box (~410k voxels per member in 1000 base leaves / ~1.3k output leaves, the regime where the per-member fixed cost dominated):

| B | main (ms) | PR 780 (ms) | this PR (ms) |
|---|---|---|---|
| 1 | 2.60 | 1.88 | 1.81 |
| 4 | 7.83 | 4.62 | 1.27 |
| 16 | 27.31 | 15.49 | 1.43 |
| 64 | 106.31 | 62.11 | 3.26 |

At B = 1 the base grid comes from PR 780's PointsToGrid branch (its B = 1 time is unchanged by design); this PR removes the PadGrid's own syncs there (2.05 -> 1.89 ms). Before PR 780's single-member fallback, the same code measured 1.26 ms at B = 1 through the Coords pass, at a higher peak.

`test_nearest_voxels_to_points_peak_memory` (2M points, B = 1): torch-visible peak 104.2 MiB (main: 158.9 MiB; PR 780 alone: 170.8 MiB before its single-member fallback), well under the 200 MiB bound. The batched pad's scratch is per leaf, not per point, so it is a few MiB here, and PointsToGrid's B = 1 scratch is smaller than the old PadGrid TopologyBuilder scratch it replaces in the peak.

## Tests

`tests/unit/test_batched_topology_builder.py` (44 passed):
- New `test_from_nearest_voxels_to_points_matches_expected_and_cpu`, parameterized over point batches covering mixed member sizes, empty members (first/middle/last), an all-empty batch, points whose floored / padded coordinates straddle the +-4096 and -8192 root-tile boundaries with negative components (including points exactly on voxel boundaries), single members (1500 points, one point, no points -- the B = 1 PointsToGrid + `resultFromGridHandle` path) and a 16-member batch. Each result is pinned elementwise (`_check_against_expected`: `ijk.jdata`, `num_voxels`, per-member bbox) against a `from_ijk` build of expected coordinates computed in torch the way the op's `VoxelCoordTransform` does (`p * (1/voxel_size) + (-origin/voxel_size)`, floor, Minkowski sum with {0,1}^3, `unique`; power-of-two voxel size so the float32 arithmetic is exact on both sides), checked multi-member == per-member builds, and compared against the CPU implementation (`_check_against_cpu`).
- New `test_from_nearest_voxels_to_points_per_member_transforms`: per-member voxel sizes and origins (including an empty member), vs expected and vs CPU.
- All PR 780 tests unchanged and passing.

Existing suites on the new build: `test_basic_ops.py`, `test_basic_ops_single.py`, `test_empty_grids.py`, `test_sliced_batch.py`, `test_batching.py`: 524 passed, 1 skipped (includes `test_nearest_voxels_to_points_peak_memory`, `test_build_from_pointcloud_nearest_voxels` on every device/dtype combination, and the `from_nearest_voxels_to_points` empty-grid and >2^31-candidate guard tests).

Part of #775 (item 6).

### DIFF
--- a/src/fvdb/detail/ops/BuildGridFromIjk.cu
+++ b/src/fvdb/detail/ops/BuildGridFromIjk.cu
@@ -1,6 +1,7 @@
 // Copyright Contributors to the OpenVDB Project
 // SPDX-License-Identifier: Apache-2.0
 //
+#include <fvdb/BuilderResource.h>
 #include <fvdb/GridBatchData.h>
 #include <fvdb/detail/GridBatchDataFactory.h>
 #include <fvdb/detail/ops/BuildGridFromIjk.h>
@@ -34,7 +35,8 @@ nanovdb::GridHandle<TorchDeviceBuffer> dispatchCreateNanoGridFromIJK(const Jagge
 // NanoVDB's PointsToGrid / DistributedPointsToGrid radix sort casts the per-grid coordinate count
 // to int32 (PointsToGrid.cuh:645), which silently corrupts the grid above 2^31 candidates, and the
 // batched Coords pass indexes its emission slots (one per coordinate) with int32 as well. Reject
-// that here rather than return a garbage grid. Takes an already-on-host joffsets accessor so each
+// that here (on every CUDA path: single-member PointsToGrid and multi-member Coords) rather than
+// return a garbage grid. Takes an already-on-host joffsets accessor so each
 // dispatch reuses the host copy it makes anyway -- no extra device sync.
 static void
 checkCandidateCountsFitInt32(const torch::TensorAccessor<fvdb::JOffsetsType, 1> &joffsetsAcc) {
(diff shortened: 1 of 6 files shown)


swahtz:

PR #783, opened 2026-09-14
Overlap destination prefetch with gradient computation and NCCL reduction

matthewdcong:
When computing mGPU gradients, we often compute the local contribution in a device-local buffer and then use NCCL to reduce these device-local buffers. This is followed by a copy to managed memory. This eliminates the need for system atomics and cross-GPU paging which can become slow.

This PR builds upon that work by overlapping the prefetch of the managed memory destination with the device-local buffer initialization, gradient computation, and the NCCL reduction. This overlapping leads to a 12.5% end-to-end improvement in reconstruction time. In addition, we introduce the `makeLocalGradient` helper function to make allocation/deallocation of these device-local gradient buffers more concise.


src/fvdb/detail/utils/cuda/LocalGradient.cpp
```diff
@@ -0,0 +1,36 @@
...
+namespace fvdb::detail {
+
+torch::Tensor
+makeLocalGradient(const torch::Tensor &tensor, c10::DeviceIndex deviceId, cudaStream_t stream) {
+    const c10::Device device(c10::kCUDA, deviceId);
+    const auto options = tensor.options().device(device);
+
+    void *ptr = nullptr;
+    if (tensor.numel() != 0) {
+        cudaMallocAsync(&ptr, tensor.numel() * tensor.element_size(), stream);
```
swahtz: **☑️ todo:** ‏I would suggest we wrap these in `C10_CUDA_CHECK` like we did in the equivalent calls that this replaced:

```
  const auto allocateLocalGradient = [&](const torch::Tensor &like, float *&ptr) {
      const size_t numBytes = like.numel() * like.element_size();
      C10_CUDA_CHECK(cudaMallocAsync(&ptr, numBytes, stream));
      C10_CUDA_CHECK(cudaMemsetAsync(ptr, 0, numBytes, stream));
      return at::from_blob(ptr, like.sizes(), localTensorOptions);
  };
```

swahtz (changes requested):
One minor comment but I'd also ask we add some tests.  Something like:

- Give makeLocalGradient a huge logical shape backed by an expanded single CPU scalar. Expect an exception.
-  Give each GPU distinct, known gradient values; run reduction, prefetch, and copy; immediately queue a consumer. Check the sum. Include evenly divided sizes, uneven sizes, and fewer elements than GPUs.

src/fvdb/detail/utils/cuda/LocalGradient.cpp
matthewdcong: :+1: , wrapped with `C10_CUDA_CHECK`

matthewdcong:
> One minor comment but I'd also ask we add some tests. Something like:
> 
>     * Give makeLocalGradient a huge logical shape backed by an expanded single CPU scalar. Expect an exception.
> 
>     * Give each GPU distinct, known gradient values; run reduction, prefetch, and copy; immediately queue a consumer. Check the sum. Include evenly divided sizes, uneven sizes, and fewer elements than GPUs.

Added a test for `makeLocalGradient`. We discussed offline and the second test isn't possible to add right now, so we'll skip it for now.

swahtz (approved):
Nice, thanks for addressing the notes and for the optimization

PR #784, opened 2026-09-15
docs: complete Python API coverage and enforce it in CI

swahtz:
## Summary

Public Python APIs were missing from the reference, and CI could pass without noticing new undocumented exports. Complete the reference and enforce docstring and API-reference coverage on pull requests.

- Add 46 top-level API entries and eight functional entries, plus missing visualization and U-Net descriptions.
- Publish neural-network methods and properties automatically with `autoclass :members:`, excluding `extra_repr`. Render members inherited from private in-package bases (`SparseConv3d.reset_parameters`, the shared widget-view members) with a scoped `:inherited-members:`.
- Add the pinned official Ruff Action to `codestyle.yml` for class, method, function, and package docstrings and missing parameter descriptions in Google-style `Args` sections. Resolve existing violations.
- Add `sphinx.ext.coverage` to the docs workflow with an adapter that counts `__all__` re-exports plus methods, properties, nested classes, and members inherited from in-package base classes. Import under autodoc's mock so the result does not depend on a prior autodoc pass or on a compiled `_fvdb_cpp`. Fail when a compiled (mocked) class has no documented members. Enforce a configurable **100%** coverage threshold and upload text/JSON reports on failure or success.
- Document that jagged reductions must keep the leading dimension (`dim=None` is Tensor-only), and describe `point_size` in pixels.
- Document the check commands, coverage scope, threshold, and intentional exclusions in the README.

Coverage is currently **597/597 public API entries**. The check fails on empty inventories and invalid exports. Members inherited from outside `fvdb` (for example `torch.nn.Module`) and ordinary data attributes are outside this coverage measure; members of compiled exports cannot be enumerated under the mock and are documented manually. Ruff does not verify the behavioral accuracy of prose.

## Validation

Passed locally with Sphinx 9.1 and Ruff 0.15.7:

- `ruff check fvdb`.
- `sphinx-build -W --keep-going -a docs build/sphinx`.
- `sphinx-build -W --keep-going -b coverage docs build/sphinx-coverage`, run twice against the same output directory (100% both times; the warm-doctree case previously failed).
- Removing the seven `NanoVDBGridMetadata` attribute entries now drops coverage to 596/597 with an explicit warning.
- From `tests/`: `python -m pytest ../.github/scripts/test_public_api_coverage.py -q` (27 passed). Tests cover re-exports, missing functions/classes/methods/properties, inherited/nested/cached members, mocked imports without a prior autodoc pass, mocked classes without documented members, threshold boundaries and command-line overrides, explicit exclusions, empty or invalid inventories, and exclusion of Sphinx extensions from pytest collection.
- Black on changed Python files with `--target-version=py311 --line-length=120`, plus `git diff --check`.


docs/_ext/public_api_coverage.py
```diff
@@ -0,0 +1,107 @@
...
+    The stock builder skips objects whose defining module differs from their
+    public import path, and does not include properties in its denominator.
+    Replacing only Python inventory collection retains its report generation.
+    """
+
+    def build_py_coverage(self):
+        """Compare package exports with objects registered in Sphinx's Python domain."""
+        seen = self.env.domaindata["py"]["objects"]
+        for module_name in self.config.coverage_public_modules:
+            module = import_module(module_name)
```
harrism: `import_module` runs outside autodoc's `mock()` context. It works in CI only because autodoc already imported `fvdb` under mock during the read phase of the same process and left it in `sys.modules`.

Repro: run the README command twice without a compiled fvdb:
```
sphinx-build -W --keep-going -b coverage docs build/sphinx-coverage
sphinx-build -W --keep-going -b coverage docs build/sphinx-coverage
```
The second run (warm doctree, autodoc skipped) fails with `ModuleNotFoundError: No module named 'fvdb._fvdb_cpp'` raised from `fvdb/__init__.py:60` via this line. CI is unaffected only because each job uses a fresh outdir. In an environment that does have a built fvdb, the second run silently imports the real binary and measures a different inventory than CI (e.g. `NanoVDBGridMetadata` becomes a real class with seven properties).

Suggest wrapping the import in `sphinx.ext.autodoc.mock(self.config.autodoc_mock_imports)`, or adding `-E` to the documented/CI command.

docs/_ext/public_api_coverage.py
```diff
@@ -0,0 +1,107 @@
...
+        yield full_name
+        if not inspect.isclass(obj):
+            continue
+        for attr_name, attr in vars(obj).items():
+            if attr_name.startswith("_"):
+                continue
+            if isinstance(attr, (staticmethod, classmethod)):
+                attr = attr.__func__
+            if inspect.isfunction(attr) or isinstance(attr, property):
+                yield f"{full_name}.{attr_name}"
```
harrism: With `autodoc_mock_imports`, the C++-bound exports (`fvdb.NanoVDBGridMetadata`, `fvdb.morton`, `fvdb.hilbert`, `fvdb.config`) are `_MockObject` instances, so `inspect.isclass` is False and no members are ever enumerated. Concretely: deleting all seven `py:attribute` entries under `NanoVDBGridMetadata` in `operations.rst` leaves coverage at 100%. Any future C++-bound class added to `__all__` passes with a bare `py:class` line.

At minimum, list this in the README's "outside this coverage measure" section; better, keep an explicit member list for mocked exports or fail when an `__all__` export is a mock object with no documented children.

docs/_ext/public_api_coverage.py
```diff
@@ -0,0 +1,107 @@
...
+        full_name = f"{module.__name__}.{name}"
+        yield full_name
+        if not inspect.isclass(obj):
+            continue
+        for attr_name, attr in vars(obj).items():
+            if attr_name.startswith("_"):
+                continue
+            if isinstance(attr, (staticmethod, classmethod)):
+                attr = attr.__func__
+            if inspect.isfunction(attr) or isinstance(attr, property):
```
harrism: Only plain functions, static/classmethods and `property` pass this filter, and only from `vars(obj)`. Nested classes (`Widget.Nested`, `Widget.Nested.method`), `cached_property`, and inherited members are all skipped. Concrete case in this PR: `reset_parameters` on `SparseConv3d` / `SparseConvTranspose3d` lives on `_SparseConv3dBase` (`fvdb/nn/modules.py:310`), so it is neither counted here nor rendered by `:members:` (no `:inherited-members:` in `nn.rst`).

Either recurse into `inspect.isclass(attr)` members / handle inherited members, or document the gap alongside the existing "inherited methods and ordinary data attributes" note.

docs/api/operations.rst
```diff
@@ -0,0 +1,131 @@
...
+
+.. autofunction:: fvdb.eq
+.. autofunction:: fvdb.ne
+.. autofunction:: fvdb.lt
+.. autofunction:: fvdb.le
+.. autofunction:: fvdb.gt
+.. autofunction:: fvdb.ge
+.. autofunction:: fvdb.where
+
+Reductions
```
harrism, line 60: The docstrings published here for `fvdb.sum/mean/amax/amin/argmax/argmin/all/any/norm/var/std` (`fvdb/torch_jagged.py:811-820`) say "If None, reduces all dimensions ... Returns the same type as the input." On a `JaggedTensor`, `dim=None` raises `RuntimeError: torch.sum would change the primary jagged dimension (expected leading dim N, got ())` from `jagged_tensor.py:1663`; reproduced for sum/mean/amax/all/norm. The docstrings are pre-existing, but this page is what puts them in the reference. Please state that jagged reductions must preserve the leading dimension (`dim=None` is Tensor-only) and show a trailing-dimension example.

fvdb/viz/_point_cloud_view.py
```diff
@@ -10,6 +10,12 @@
 
 
 class PointCloudView:
+    """A handle for a colored point cloud displayed in a scene.
+
+    Obtain this handle from :meth:`fvdb.viz.Scene.add_point_cloud` and use
+    :attr:`point_size` to adjust the world-space radius of the displayed points.
```
harrism: "world-space radius" is incorrect: `point_size` is documented as a size in pixels (getter/setter below) and maps to the screen-space `eps_2d` as a diameter (`view.eps_2d = size / 2.0`). Suggest "to adjust the on-screen size (in pixels) of the displayed points".

harrism (changes requested):
This PR adds `docs/api/operations.rst` (concatenation, attention/volume rendering, the 44 torch-compatible ops, manual `py:function` entries for `morton`/`hilbert`, a manual `py:class` for `NanoVDBGridMetadata`), eight missing `fvdb.functional` autofunctions, `:members:` on every `fvdb.nn` autoclass, and docstrings across `fvdb/`, plus `functools.wraps` on the profiler decorator so autodoc sees the real `forward` signatures. Enforcement is (1) a new `test-python-docstrings` job running Ruff 0.15.7 with `D101-D104,D417` (google convention) on `fvdb/`, and (2) `docs/_ext/public_api_coverage.py`, a `CoverageBuilder` subclass that imports each module in `coverage_public_modules`, walks `__all__` plus `vars(cls)` for public functions/static/classmethods/properties, intersects with the py-domain inventory, and raises `SphinxError` below `coverage_min_percentage` (100). It runs in `docs-build-test.yml` against the source tree with `_fvdb_cpp` mocked, so no CUDA build is needed.

Verification: a fresh coverage build reports 577/577 with Sphinx 9.1; `ruff check fvdb` with 0.15.7 is clean; the 19 extension tests pass. The non-docs changes are safe: `functools.wraps` only adds `__wrapped__`/name/doc metadata (behaviour, hooks, gradients unchanged); the two `Args` fixes in `fvdb/__init__.py` and `jagged_tensor.py` correct real parameter-name mismatches; note there are no `__all__` edits despite the description. `nitpicky` is off, so `:class:`torch.Tensor`` on the new page unresolves silently (pre-existing pattern). The manual `morton`/`hilbert` entries (int32 in, int64 out) and the `NanoVDBGridMetadata` attribute names/types match `Bindings.cpp`.

Requesting changes for: the un-mocked `import_module` in the coverage builder (inline 1), and the two prose inaccuracies (jagged `dim=None` reductions, `point_size` units). The mocked-class-member and nested/inherited-member blind spots (inline 2-3) should at least be listed in the README's "outside this coverage measure" section before the 100% figure is advertised.

docs/_ext/public_api_coverage.py
swahtz: Confirmed; the second build against a warm doctree failed with `ModuleNotFoundError: fvdb._fvdb_cpp`. Fixed in e32c9be8: the builder now imports each coverage module inside `sphinx.ext.autodoc.mock(autodoc_mock_imports)`, so it neither depends on a prior autodoc pass nor silently picks up a compiled fvdb. Two consecutive builds against the same outdir now both report 597/597, and a new test (`test_mocked_imports_are_measured_without_autodoc_directives`) builds twice with a mocked import and no autodoc directives.

docs/_ext/public_api_coverage.py
swahtz: Confirmed. In e32c9be8 the builder now fails when an `__all__` export that is a mock object is documented as `py:class` but has no documented children in the py-domain inventory (warning plus the class counted as missing). Deleting the seven `NanoVDBGridMetadata` attribute entries now drops coverage to 596/597. Mocked functions (`morton`, `hilbert`) are unaffected since they are registered as `py:function`. The README now states that compiled exports cannot be enumerated and must be documented manually against `Bindings.cpp`, and that a stale or missing attribute still cannot be detected.

docs/_ext/public_api_coverage.py
swahtz: Confirmed, and it was wider than `reset_parameters`: the four `fvdb.viz` widget views each inherit `name`, `scene_name`, `on_update`, `remove_on_update` from `_WidgetView`. In e32c9be8 the inventory walks the MRO for bases defined inside the `fvdb` package (so `torch.nn.Module` members stay out), recurses into nested classes, and accepts `functools.cached_property`. `nn.rst` uses `:inherited-members: Module` on `SparseConv3d`/`SparseConvTranspose3d` and `viz.rst` gets a Widgets section with `:inherited-members:` so the counted members actually render. Coverage went from 579 to 597 objects; tests added for inherited, nested, and cached members. Members inherited from outside `fvdb` remain documented as out of scope in the README.

docs/api/operations.rst
swahtz: Confirmed via the leading-dimension guard in `jagged_tensor.py`. e32c9be8 rewrites the `dim` description in all eleven reductions (`torch.Tensor` only for `dim=None`; a JaggedTensor must keep its leading dimension, so pass `dim=-1` or another trailing dimension) and adds a short note with a `dim=-1` example and the failing `dim=None` case at the top of the Reductions section in `operations.rst`.

fvdb/viz/_point_cloud_view.py
swahtz: Fixed in e32c9be8: now reads "to adjust the on-screen size (in pixels) of the displayed points".

docs/_ext/public_api_coverage.py
```diff
@@ -0,0 +1,157 @@
...
+            if attr_name.startswith("_"):
+                continue
+            if isinstance(attr, (staticmethod, classmethod)):
+                attr = attr.__func__
+            member = f"{full_name}.{attr_name}"
+            if inspect.isfunction(attr) or isinstance(attr, (property, functools.cached_property)):
+                yield member
+            elif inspect.isclass(attr) and _in_package(attr, package):
+                yield member
+                yield from _class_members(member, attr, package)
```
harrism, line 46: Non-blocking. The nested-class branch follows any class-valued attribute whose `__module__` is inside the package, without checking that it is actually defined inside `klass`. A class-level alias such as `class Scene: View = CamerasView` would be enumerated as `fvdb.viz.Scene.View` plus every member of `CamerasView` under that path, forcing bogus duplicate reference entries; a self-referential alias (`Foo.Self = Foo`) would recurse without bound. No such attribute exists in the current tree (verified by walking every exported class), so this has no live trigger. Suggested guard: `elif inspect.isclass(attr) and attr.__qualname__ == f"{klass.__qualname__}.{attr_name}":`.

docs/_ext/public_api_coverage.py
```diff
@@ -0,0 +1,157 @@
...
+            self.py_undocumented[module_name] = missing
+            # Sphinx's text report groups functions separately from class methods.
+            funcs = []
+            classes = {}
+            for name in sorted(missing):
+                relative = name[len(module_name) + 1 :]
+                if "." in relative:
+                    cls_name, method = relative.split(".", 1)
+                    classes.setdefault(cls_name, []).append(method)
+                elif inspect.isclass(getattr(module, relative)):
```
harrism: Non-blocking. An opaque mocked class returned by `_opaque_classes` reaches this branch with `inspect.isclass(...)` false (it is a `_MockObject` instance), so it is appended to `funcs`. Reproduced: with the seven `NanoVDBGridMetadata` attribute entries removed, `python.txt` lists `NanoVDBGridMetadata` under "Functions:". `coverage.json` and the failure message are correct; only the text report is misfiled. Suggest treating `_is_mocked(obj) and seen.get(name) is not None and seen[name].objtype == "class"` (or simply the members of the opaque set) as classes here.

docs/api/operations.rst
```diff
@@ -0,0 +1,139 @@
...
+      Voxel size in world units, as a tuple of three floats.
+
+   .. py:attribute:: index_bbox_min
+
+      Inclusive minimum of the index-space bounding box, as a tuple of three integers.
+
+   .. py:attribute:: index_bbox_max
+
+      Inclusive maximum of the index-space bounding box, as a tuple of three integers.
+
```
harrism: Non-blocking. `git diff --check fcd3d4f e32c9be8` reports `docs/api/operations.rst:139: new blank line at EOF`. The `trailingspaces` CI job only greps for trailing spaces, so this does not fail CI; drop the extra blank line so the file ends with a single newline.

harrism (approved):
Re-review of e32c9be8 against the request-changes review on 8b90c3eb. The first three commits are a content-identical rebase onto fcd3d4f (range-diff shows only hunks that upstream main already contains: the `functools.wraps` profiler change and the `Prune` autoclass); all substantive changes are in e32c9be8.

**Previous findings**

1. Un-mocked `import_module` in the coverage builder - **resolved.** `build_py_coverage` now imports and enumerates inside `sphinx.ext.autodoc.mock(autodoc_mock_imports)`. Verified with Sphinx 8.2.3 (the `docs/requirements.txt` lower bound): two consecutive `sphinx-build -W --keep-going -b coverage` runs against the same outdir both report 597/597 (the warm-doctree run previously raised `ModuleNotFoundError`). `test_mocked_imports_are_measured_without_autodoc_directives` builds twice with a mocked import and no autodoc directives.
2. Mocked compiled classes never enumerated - **resolved.** `_opaque_classes` counts a mocked `__all__` export documented as `py:class` with no `Name.*` inventory entries as missing and emits a warning. Reproduced the original scenario: deleting the seven `NanoVDBGridMetadata` attribute entries now yields `WARNING: fvdb.NanoVDBGridMetadata is a compiled export whose members cannot be enumerated ...`, 596/597, and a failed build. The README states the residual gap (stale or missing individual attributes are undetectable).
3. Inherited, nested, and `cached_property` members skipped - **resolved.** `_class_members` walks the MRO restricted to in-package bases, recurses into nested classes, and accepts `cached_property`. Enumeration of the real tree now includes `SparseConv3d/SparseConvTranspose3d.reset_parameters` and the four widget views' `name`, `scene_name`, `on_update`, `remove_on_update`; a text render confirms they appear in the reference and that no `torch.nn.Module` members leak through `:inherited-members: Module`. Tests cover inherited, overridden, nested, and cached members.
4. Jagged `dim=None` reduction docstrings - **resolved.** All eleven reductions in `fvdb/torch_jagged.py` state that `dim=None` is `torch.Tensor`-only, and `operations.rst` adds a note with a `dim=-1` example (shape `(6,)` is correct for the example inputs) and the failing `dim=None` case.
5. `point_size` described as a world-space radius - **resolved.** Now "on-screen size (in pixels)".

**Verification of the new commit**: 27/27 extension tests pass under Sphinx 8.2.3 and 9.1.0; coverage build is clean at 597/597 (cold and warm); Black is clean on the changed Python files. The `Duplicate explicit target name` warning in `docs/installation.rst` is pre-existing on main and unrelated.

The PR description's bullet about adding `functools.wraps` to the profiling decorator is stale: `fvdb/nn/modules.py` on the base commit already carries it, and the rebased commit correctly no longer touches it.

**New observations** (non-blocking nits, inline): the nested-class recursion does not distinguish true nested classes from class-valued aliases; opaque mocked classes are listed under "Functions:" in `python.txt`; `operations.rst` has an extra blank line at EOF.

Verdict: all five previous findings are resolved and no new defect was found. Approve.

docs/_ext/public_api_coverage.py
swahtz: Good catch. Applied your suggested guard in c22a3ff9 (`attr.__qualname__ == f"{klass.__qualname__}.{attr_name}"`). The inherited-members test fixture now includes a class-valued alias and a self-referential alias, and asserts neither is enumerated.

docs/_ext/public_api_coverage.py
swahtz: Fixed in c22a3ff9: `build_py_coverage` keeps the opaque set and files those names under `classes` when building `py_undoc`. With the seven attribute entries removed, `python.txt` now lists `NanoVDBGridMetadata` under "Classes:"; the mocked-class test asserts the grouping.

docs/api/operations.rst
swahtz: Dropped in c22a3ff9; `git diff --check` against main is clean.

PR #786, opened 2026-09-15
Add GridStorage, TorchDeviceResource and device-handle utilities (#770 step 3)

harrism:
## Summary

Step 3 of #770: the primitives the grid-storage flip is built on. **Nothing uses them yet.** `GridBatchData` still holds `GridHandle<TorchDeviceBuffer>`; steps 4 and 5 move storage and the builders onto these. Reviewing this on its own keeps the flip PR to mechanical call-site changes.

## What is added

| piece | role |
|---|---|
| `TorchDeviceResource.h` | Storage resource bound to one torch device (`CUDA:i` or PrivateUse1). NanoVDB picks host vs. device storage per buffer *type*, so the device travels in the resource: a prototype buffer's `resource()` is what `createDeviceStorage` and `cuda::copyTo` allocate from. The block is keyed to the buffer's **retained** stream (the `TorchResource` policy): a `cuda::Buffer` orders its writes and its free on that stream, so torch's reuse rule and NanoVDB's ordering model agree with no events. This is deliberately *not* `TorchStorageResource`'s current-stream policy from #773, which is right for the dual-space `TorchDeviceBuffer` (no retained stream, unordered free) and wrong here; see the header comment. Obligations: the retained stream must outlive the block and must not be changed after allocation (`set_stream`/`resize` to another stream); `GridStorage` copies instead. `DeviceGridBuffer = cuda::Buffer<std::byte, TorchDeviceResource>`. |
| `GridStorage.{h,cu}` | `variant<GridHandle<HostBuffer>, GridHandle<DeviceGridBuffer>>` + `torch::Device`. Count, sizes, offsets, types and typed grid pointers come from adopted metadata; no device reads. `to(device, stream)` is `cuda::copyTo` with explicit ordering (below). `empty(device)`, `deviceProto(device, stream)` (replaces the zero-size `TorchDeviceBuffer` guide). Host-includable. |
| `GridHeaderUtils.h` | `normalizeStandaloneGridHeader`: the one spelling of the header writes every hand-wrapped grid needs now that `GridHandle` validates the chain (index 0 of 1, size, no blind data with offset = size, checksum disabled). The six existing sites migrate in step 4. |
| `DeviceGridHandleUtils.cuh` | `updateGridCountAndZeroChecksum` hoisted from `MakeContiguous`/`ConcatenateGrids`; `makeDeviceHandleFromLayout`; `mergeDeviceGridHandles`, the single-space replacement for `nanovdb::cuda::mergeGridHandles`: one allocation, one memcpy + one header kernel per grid, host-assembled metadata, no per-grid synchronization. |
| `GridStorageTest.cu` | Twelve gtests, below. |

## Stream contract of `GridStorage::to`

- The copy runs on the caller's `stream`. If the source is device storage, `stream` first waits on the source's retained stream, so the source's writers are complete.
- For a device destination the source's retained stream then waits on the copy. Destroying the source right after `to()` returns is therefore safe even when torch's caching allocator hands its block straight to the next allocation on that stream. This is the failure the #770 design review reproduced with a probe; `SourceMayBeDestroyedRightAfterAsyncCopy` reproduces it deterministically (the copy stream is held open by a spin kernel, the source is destroyed, its block is re-taken and overwritten on the source stream) and passes.
- A host destination is complete on return (`copyTo` synchronizes), so no post-copy ordering is needed there.
- Across CUDA devices the copy is `cudaMemcpyPeerAsync` (torch's expandable-segment and cudaMallocAsync backends do not support a plain memcpy between devices without peer access). Events are recorded under their stream's device.
- Empty storage moves without allocating and lands on the requested device. `copyTo` alone returns a default-constructed handle, whose resource is the *current* device (design review item 2).
- An explicit legacy stream 0 is honoured as the retained stream; an omitted stream means the destination device's current torch stream (design review item 3).

## Tests

`HostAccessorsAnswerFromMetadata`, `EmptyStorageKeepsItsDevice`, `DeviceProtoCarriesDeviceAndStream`, `RoundTripThroughCudaPreservesBytesAndMetadata`, `ExplicitStreamZeroWhileNonDefaultStreamIsCurrent`, `SourceMayBeDestroyedRightAfterAsyncCopy`, `MergeDeviceGridHandlesMatchesHostMerge` (byte-for-byte against `nanovdb::mergeGrids`, including an empty source in the middle, then re-parsed from raw bytes), `MakeDeviceHandleFromLayoutChecksExtent`, `NormalizeStandaloneGridHeaderMakesAGridWrappable` (the un-normalized copy is rejected by `GridHandle`; the normalized one wraps), `StorageBlockReturnsToItsRetainedStream`, `SynchronousResourceApiWorksUnderAnyCurrentDevice`. The lifetime test uses a source retained on a non-current stream, the case that distinguishes the two keying policies.

## Verification

`./build.sh install gtests` with `-Werror=all-warnings`: clean. ctest 49/49 (the new suite 12/12 on three consecutive runs). The nine pytest suites touched by the epic: unchanged. clang-format: clean. Codex-reviewed before opening; three passes found twelve real defects, all folded in and listed in the commit message.

## For #770

Step 5's `DistributedPointsToGrid` path creates storage on a `DeviceMesh` stream that dies with the mesh. Under retained-stream keying that is an orphaned block; it has to be solved at that builder (a long-lived mesh, or re-homing the storage after the builder's own synchronization), not in the resource.


src/fvdb/GridStorage.cu
```diff
@@ -0,0 +1,270 @@
...
+        const auto &src = std::get<0>(mHandle);
+        if (device.is_cpu()) {
+            return GridStorage(src.copy<nanovdb::HostBuffer>());
+        }
+        // Host to device: the pageable source makes the copy synchronous on the host side, and
+        // the destination retains `stream`; nothing to order.
+        c10::OptionalDeviceGuard deviceGuard(device.is_cuda() ? std::optional<torch::Device>(device)
+                                                              : std::nullopt);
+        const DeviceGridBuffer proto = deviceProto(device, stream);
+        return GridStorage(nanovdb::cuda::copyTo<DeviceGridBuffer>(src, stream, &proto), device);
```
swahtz: **Host to PrivateUse1 needs a sync after the copy.** The comment above covers the source side only. `copyTo<DeviceGridBuffer>` issues `cudaMemcpyAsync(pageable -> managed, cudaMemcpyDefault, stream)` and synchronizes only when `!dstDev || BufferHasHostSingle`; `TorchDeviceResource` does not declare `HOST_ACCESSIBLE`, so for a PrivateUse1 destination it returns as soon as the pageable source is staged, not when the DMA lands. The returned storage then hands out the unified pointer through `hostBytes()` / `hostGridAt()` immediately, so

```cpp
GridStorage(hostHandle).to(torch::kPrivateUse1).hostGridAt<float>(0)->tree().getValue(c)
```

can read pre-copy bytes. The device-source branch below is safe only because its trailing `orderStreamAfter` degenerates into a host sync for PrivateUse1; this branch has no equivalent. The path this replaces in `TorchDeviceBuffer.cpp` used a blocking `cudaMemcpy` here.

Suggest `if (device.is_privateuseone()) detail::synchronizeStream(stream, device);` after the copy, and narrowing the comment to the CUDA case.

src/fvdb/GridStorage.cu
```diff
@@ -0,0 +1,270 @@
...
+GridStorage::GridStorage(DeviceHandle &&handle)
+    : mHandle(std::in_place_index<1>, std::move(handle)),
+      mDevice(std::get<1>(mHandle).buffer().resource().device()) {}
+
+GridStorage::GridStorage(DeviceHandle &&handle, const torch::Device &device)
+    : mHandle(std::in_place_index<1>, std::move(handle)), mDevice(device) {
+    checkStorageDevice(device);
+    TORCH_CHECK(isDeviceKind(device), "GridStorage: a device handle cannot live on ", device);
+    const auto &h = std::get<1>(mHandle);
+    TORCH_CHECK(h.isEmpty() || h.buffer().resource().device() == device,
```
swahtz: **An empty handle bound to a different device slips through here.** `empty()` and `to()` both re-home an empty handle onto the requested device, but this overload keeps whatever resource the empty handle carries. A default `DeviceHandle{}` binds `TorchDeviceResource()` to the *current* device, and `cuda::copyTo` returns exactly such a handle for zero-byte sources, which is the case `GridStorage.h:50-52` steers into this constructor.

With cuda:1 current, `GridStorage s(GridStorage::DeviceHandle{}, cuda:0)` passes the check, `s.device()` says cuda:0, and `s.deviceHandle().buffer().resource().device()` says cuda:1. Anything that uses the buffer as a proto (`createDeviceStorage(bytes, &s.deviceHandle().buffer(), ...)`) then allocates on cuda:1 and is wrapped as cuda:0 storage, and `GridStorage(std::move(s.deviceHandle())).device()` reports cuda:1. The header and the tests treat the two as always agreeing.

Suggest applying the same rule as `to()`: when `h.isEmpty()` and the resource device differs, replace it with `emptyDeviceHandle(device, h.buffer().stream())`.

src/fvdb/detail/utils/nanovdb/DeviceGridHandleUtils.cuh
```diff
@@ -0,0 +1,130 @@
...
+///        are when copyTo adopts metadata.
+inline GridStorage::DeviceHandle
+makeDeviceHandleFromLayout(DeviceGridBuffer &&buffer,
+                           std::vector<nanovdb::GridHandleMetaData> meta) {
+    const uint64_t extent = buffer.size_bytes();
+    uint64_t end          = 0;
+    for (const auto &m: meta) {
+        // Checked without forming offset + size, which can wrap.
+        TORCH_CHECK(m.offset >= end && m.offset <= extent && m.size >= sizeof(nanovdb::GridData) &&
+                        m.size <= extent - m.offset,
```
swahtz: Nit, hardening only since today's callers lay out packed aligned grids. This check admits three layouts that produce handles the rest of the stack cannot use: a gap between grids (`m.offset > end`), an offset that is not a multiple of `NANOVDB_DATA_ALIGNMENT`, and empty `meta` over a non-empty buffer. With `meta={{0,s0},{s0+8,s1}}`, `deviceGridAt<float>(1)` is a 24-byte-misaligned `GridData*` and `updateGridCountAndZeroChecksum` faults with a misaligned address; even an aligned gap yields a handle that `DeviceHandle(buffer.copy(stream))` later rejects, since `parseGridChainKernel` advances by `mGridSize`.

Tightening to `m.offset == end`, `m.offset % NANOVDB_DATA_ALIGNMENT == 0`, and `meta.empty() == (extent == 0)` makes the helper match what it promises.

src/fvdb/detail/utils/nanovdb/DeviceGridHandleUtils.cuh
```diff
@@ -0,0 +1,130 @@
...
+    return nanovdb::cuda::detail::HandleFactory::make(std::move(buffer), std::move(meta));
+}
+
+/// @brief Lays the grids of @p handles end to end in one allocation from @p proto's resource,
+///        on @p stream, fixing each header's index/count: the single-space replacement for
+///        nanovdb::cuda::mergeGridHandles. No parse kernel and no per-grid synchronization; the
+///        metadata is assembled on the host from the sources'. All sources must live on the
+///        prototype's device. Each source's retained stream is ordered before its copy and after
+///        it, so sources may be destroyed as soon as this returns. Checksums are disabled, as the
+///        other fvdb merge paths leave them.
```
swahtz: Doc nit. `nanovdb::cuda::mergeGridHandles`, which this replaces for the CUDA builders, does not leave checksums disabled; it recomputes an existing one (`CheckMode::Partial`) after `updateGridCount`. The `mergeGrids` host path and `ConcatenateGrids` / `MakeContiguous` do disable them, so the sentence is true of those. Worth saying which, since every step-5 CUDA builder calls `setChecksum` before merging and returns early at `handles.size()==1`. Which way step 5 goes (recompute here, or drop those `setChecksum` calls) is raised on #770; this comment just should not claim parity with the replaced path.

swahtz (commented):
Reviewed at f28cfb38 against the #770 plan and its comment thread. Two items look like they should land before step 4 builds on these primitives (the PrivateUse1 host-source sync and the 2-arg constructor). The rest are nits or optional cleanups, marked as such. Points that belong to later steps rather than this PR (checksum policy for the merge helper, cross-stream readers, PrivateUse1 ordering cost, the long-name flag in the normalizer) are on #770 instead.

src/tests/GridStorageTest.cu
```diff
@@ -0,0 +1,428 @@
...
+                               TorchDeviceResource(kCuda0),
+                               nbytes,
+                               nanovdb::cuda::noInit);
+    DeviceGridBuffer onSide(
+        side.stream(), TorchDeviceResource(kCuda0), nbytes, nanovdb::cuda::noInit);
+    if (onSide.data() != first && onCurrent.data() != first) {
+        GTEST_SKIP() << "allocator did not reuse the block at all (caching disabled?)";
+    }
+    EXPECT_EQ(onSide.data(), first) << "the freed block should be reusable on its retained stream";
+    EXPECT_NE(onCurrent.data(), first) << "and not on an unrelated stream";
```
swahtz, line 315: Nit. This asserts a native-caching-allocator detail that `TorchResource.h` documents as backend-dependent. Under `PYTORCH_CUDA_ALLOC_CONF=backend:cudaMallocAsync`, `raw_delete` becomes `cudaFreeAsync(ptr, side)` with nothing queued on `side`, and torch enables opportunistic pool reuse, so the next allocation on the current stream can receive `first`; `onCurrent.data()==first` skips the `GTEST_SKIP` guard above and both expectations fail. Only `PYTORCH_NO_CUDA_MEMORY_CACHING` is guarded. `TorchDeviceBufferTest.cu:144-147` already skips for this backend; mirroring that here keeps the test honest about what it can observe.

src/fvdb/detail/utils/cuda/StreamOrdering.h
```diff
@@ -0,0 +1,80 @@
...
+    // PrivateUse1 first: two null handles on PrivateUse1 need not be the same GPU's default
+    // stream, so the equality shortcut below must not apply to them.
+    if (!signalerDevice.is_cuda() || !waiterDevice.is_cuda()) {
+        synchronizeStream(signaler, signalerDevice);
+        return;
+    }
+    if (waiter == signaler && waiterDevice == signalerDevice) {
+        return;
+    }
+    cudaEvent_t ev = nullptr;
```
swahtz, line 61: Optional, non-blocking. `at::cuda::CUDAEvent` already does this sequence with the right device guards (`record()` guards the recording stream's device, `block()` the waiting stream's device, the destructor destroys under the event's device), so after the early-outs the body reduces to

```cpp
at::cuda::CUDAEvent ev(cudaEventDisableTiming);
ev.record(at::cuda::getStreamFromExternal(signaler, sigIdx));   // getDefaultCUDAStream(idx) for a null handle
ev.block(at::cuda::getStreamFromExternal(waiter, waitIdx));
```

Same per-call create/destroy cost either way. `TorchStorageResource::allocate_async` carries a second copy of the same dance, but that one serves only the deprecated `TorchDeviceBuffer` and goes at the next pin bump, so this is the copy that will live on. Take or leave.

src/fvdb/GridStorage.h
```diff
@@ -0,0 +1,187 @@
...
+    DeviceHandle &deviceHandle();
+    //@}
+
+    /// @brief The adopted per-grid metadata, in physical order.
+    const std::vector<nanovdb::GridHandleMetaData> &metadata() const;
+
+  private:
+    template <typename ValueT>
+    const nanovdb::NanoGrid<ValueT> *
+    gridAt(const void *base, uint32_t i) const {
```
swahtz: Optional, non-blocking. `gridAt<ValueT>` re-derives what `GridHandle` already exposes as `grid<ValueT>(n)` for `HostBuffer` and `deviceGrid<ValueT>(n)` for a single-space device buffer, at the cost of three `metadata()` visits per lookup (`gridCount()`, `gridType(i)`, `gridOffset(i)`) plus two `TORCH_CHECK`s that cannot fire. Delegating by variant arm (`grid<ValueT>(i)` on the host arm, `deviceGrid<ValueT>(i)` on the device arm, with the PrivateUse1 unified case picking the device arm for `hostGridAt`) drops the duplication; a private `metaAt(i)` would likewise collapse the triplicated check in `gridSize` / `gridOffset` / `gridType`.

src/fvdb/GridStorage.cu
harrism: Right: copyTo returns once the pageable source is staged, and PrivateUse1 hands its unified pointer to the host at once. Fixed in 920f50f: `synchronizeStream(stream, device)` after the copy for a PrivateUse1 destination, comment narrowed to the CUDA case.

src/fvdb/GridStorage.cu
harrism: Agreed, the header promised the two agree and this constructor let them differ. Fixed in 920f50f: an empty handle whose resource device differs is re-homed with `emptyDeviceHandle(device, h.buffer().stream())`, the same rule `empty()` and `to()` apply; the retained stream is kept.

src/fvdb/detail/utils/nanovdb/DeviceGridHandleUtils.cuh
harrism: Taken in 920f50f: `offset == end`, `offset % NANOVDB_DATA_ALIGNMENT == 0`, and `meta.empty() == (extent == 0)`. The test adds the gap and the metadata-less-bytes cases. Trailing padding after the last grid stays allowed, as the handle's `isPadded()` expects.

src/fvdb/detail/utils/nanovdb/DeviceGridHandleUtils.cuh
harrism: Corrected in 920f50f, and made more precise after a second look: only the CUDA MakeContiguous/ConcatenateGrids kernel (the one hoisted into this header) disables checksums; `nanovdb::mergeGrids` and the CPU sides of those two ops go through `tools::updateGridCount`, which updates an existing checksum, as the replaced `nanovdb::cuda::mergeGridHandles` did. The comment now says exactly that and no longer claims parity. Which way step 5 goes is on #770.

src/tests/GridStorageTest.cu
harrism: Taken in 920f50f: the test skips unless `CUDACachingAllocator::name() == "native"`, alongside the existing no-caching skip. Verified the suite under `backend:cudaMallocAsync` and under `PYTORCH_NO_CUDA_MEMORY_CACHING=1`: 10 pass, this one skips, in both.

src/fvdb/detail/utils/cuda/StreamOrdering.h
harrism: Tried it, and it is the version I would rather have, but it does not survive the legacy sentinel: `getStreamFromExternal(cudaStreamLegacy, idx)` hands c10 stream id 1, which it rejects ("invalid StreamId1") when converting back, whereas raw `cudaEventRecord` accepts the handle. The raw body stays, with the three guards `CUDAEvent` would have applied (record and destroy under the signaler's device, wait under the waiter's) and a comment saying why. If fvdb ever forbids the sentinels at this seam, this collapses to your three lines.

src/fvdb/GridStorage.h
harrism: Taken in 920f50f: `hostGridAt`/`deviceGridAt` delegate to `grid<T>(i)` on the host arm and `deviceGrid<T>(i)` on the device arm (PrivateUse1 host access goes through the device arm), and one `metaAt(i)` replaces the three duplicated checks. Same nullptr behaviour for out-of-range and wrong type.

swahtz (approved):
Looks good, thanks for addressing my comments

PR #787, opened 2026-09-16
Hold grid storage in GridStorage instead of GridHandle<TorchDeviceBuffer> (#770 step 4)

harrism:
## Summary

Step 4 of #770: `GridBatchData` holds a `GridStorage` (#786) instead of `GridHandle<TorchDeviceBuffer>`. `nanoGridHandle()` is gone. Every path that assembles grids by hand (compaction, concatenation, clone, serialize, load, save, empty grids) produces `GridStorage` directly; the builders are untouched and still return the legacy handle type, which the factory adopts through a temporary copying seam.

## What changes for callers

| before | after |
|---|---|
| `nanoGridHandle().buffer().data() / deviceData() / size()` | `gridStorage().hostBytes() / deviceBytes() / bufferSize()` |
| `nanoGridHandle().grid<T>(i)` / `deviceGrid<T>(i)` (physical index) | `hostGridPtrAt(i)` / `deviceGridPtrAt(i)` (logical, view-safe) |
| `nanoGridHandle().gridCount()` for a reserve | `batchSize()` |
| `contiguousGridHandle` for fvdb-internal use | `contiguousGridStorage` |

## Correctness fixes folded in

`SerializeGrid` on a non-contiguous CPU batch exported every physical grid and read past the view's metadata; it now compacts first. An empty batch serialized by copying one metadata record from a null pointer; it now round-trips as zero grids. `SaveNanoVDB`'s blind-data assembler patched host memory right behind each `cudaMemcpyAsync` into pageable memory, which CUDA only *may* complete synchronously; it now copies everything, synchronizes once, then patches, and renumbers each logically selected grid's header for the file it is assembling.

These sites indexed the shared storage *physically* and read the wrong grid for a sliced batch (`gridCount() > batchSize()`): `Inject` (both paths), `BuildMergedGrids` (CPU and CUDA), `PredGatherIGemm`, `ReinitializeSdf`, and `SaveNanoVDB`'s blind-data assembler and `getIndexGrid`. All now go through the logical accessors or compact first.

## Kernel-free assembly

`contiguousGridStorage`, `concatenateGrids` and the empty-merge path know their layout, so the device handle adopts it (`makeDeviceHandleFromLayout`): no parse kernel, no per-grid synchronization. The duplicated `updateGridCountAndZeroChecksum` kernels in `MakeContiguous` and `ConcatenateGrids` are replaced by the one hoisted in #786.

## Stream contract for copies out of a batch

Compaction, concatenation and the legacy handle paths order their copy stream after the source storage's retained stream (its writers) and that stream after the copies (so releasing the source right away cannot let torch reuse its block under the copy), via the #786 helper. PrivateUse1 bytes are read by host code immediately, so there the copy stream is synchronized instead. The legacy adapter synchronizes unconditionally: the dual-space buffer does not record the stream its block is keyed to.

## The temporary seam

`makeGridBatchData(GridHandle<TorchDeviceBuffer>&&)` remains, implemented by `adoptLegacyGridHandle` (`LegacyGridHandle.h`), which copies the bytes into `GridStorage` on the device's current torch stream and synchronizes (see above). **Until step 5 converts the 26 builder sites this costs one extra copy of the grid bytes per build.** `contiguousGridHandle`, `cloneGridHandleAt` and `createEmptyGridHandle` keep their legacy return type for the builders that merge per-item handles; step 5 removes all of it. I chose the copying seam over a third variant arm in `GridStorage`: it is ~30 lines that disappear, versus threading a legacy arm through every accessor of a class that was just reviewed.

## Verification

`./build.sh install gtests` with `-Werror=all-warnings`: clean. ctest 50/50. Full `tests/unit`: 2993 passed / 33 skipped (one known fp16 flake deselected). Memory-sensitive files (`test_basic_ops`, `test_dual`, `test_basic_ops_single`, `test_io`, `test_sdf`) rerun with `PYTORCH_NO_CUDA_MEMORY_CACHING=1` as CI does: 1089 passed. Targeted probes of the fixes above: pickle round trip of a CUDA view, a CPU view and an empty CUDA batch; `save_nanovdb`/`from_nanovdb` of a CUDA view, a CPU view and a reordered two-grid view, all reloading with the right grid count, voxel count and features. clang-format clean.

Codex-reviewed before opening: three passes, seven real findings, all folded in and listed in the commit message.


swahtz (commented):
Reviewed against the #770 thread as well as the diff. Six inline comments below; the first four are correctness, the last two are small.

Three items the thread assigned to step 4 in `SaveNanoVDB.cu` are not in this PR, and step 5 (builders) and step 6 (deprecation) do not revisit this file:

1. **`patchGridWithBlindShape` checksum.** The function still rewrites `mGridSize`, index/count, blind fields, `mVoxelSize`, `mMap` and `mGridName` with no `mChecksum.disable()`. The raw-bytes assembler now disables per grid (nice), but the typed-values export calls this function unchanged, so batch-size-1 exports of partial-checksum grids still fail `nanovdb_validate`.
2. **Empty grid in the CUDA typed export.** The loop at ~L756 calls `indexToGrid` for `numVoxelsBi == 0`, which upstream launches as an unguarded `<<<0, 1>>>` and `cudaCheckError` exits.
3. **`HasLongGridName` in `normalizeStandaloneGridHeader`.** Untouched, and `patchGridWithBlindShape` still hand-writes `mBlindMetadataCount = 1`, the pattern the thread flagged as not a drop-in for the normalizer.

Either they land here, or the issue needs a named home for them. Findings that step 5 already owns (the identity paths' double copy through the seam, the three copies of the copy-and-fix loop, PU1 sync counts, the `createEmptyGridStorage` duplicate) I've put on #770 rather than here.

src/fvdb/detail/io/SaveNanoVDB.cu
```diff
@@ -201,68 +202,85 @@ assembleOnIndexBlindBuffer(const GridBatchData &gridBatchData,
...
     for (int64_t batchIdx = 0; batchIdx < gridBatchData.batchSize(); ++batchIdx) {
-        // Copy this batch's index grid to the buffer. D2H copies into pageable host memory are
-        // synchronous w.r.t. the host, so the immediate host-side patches below are safe; the
-        // final stream sync is belt-and-suspenders.
-        const size_t gridBytes = nanoGridHdl.gridSize(batchIdx);
+        const size_t gridBytes  = gridBatchData.numBytesAt(batchIdx);
+        const uint8_t *readHead = srcBase + gridBatchData.cumBytesAt(batchIdx);
         if (isCuda) {
             c10::cuda::CUDAGuard deviceGuard(gridBatchData.device());
             at::cuda::CUDAStream stream =
```
swahtz, line 254: `assembleOnIndexBlindBuffer` here, and `fvdbToNanovdbGridWithValues` at the `dSrcBufferStart + cumBytesAt(bi)` read (~L737), read `gridStorage().deviceBytes()` on the current torch stream without ordering it after `storage.stream()`. Everywhere else this PR reads storage it brackets with `orderStreamAfter` (`GridStorage.h:152-158`, `MakeContiguous.cu:46-56`, and `getIndexGrid` goes through `contiguousGridStorage`).

Repro shape: build, `contiguous()`, or `to(device)` a CUDA batch under `with torch.cuda.stream(s1):` so the buffer retains s1 with header-fix kernels or copies still queued, then `save_nanovdb`/`to_nanovdb` with a JaggedTensor on the default stream. The D2H memcpys (or the `indexToGrid` kernel) run alongside s1's writers.

Pre-existing hazard, but both hunks were rewritten here. One `detail::orderStreamAfter(stream, device, storage.stream(), device)` before the first read in each function closes it, or route both through `contiguousGridStorage(...).to(kCPU)` like `getIndexGrid`.

src/fvdb/detail/io/SaveNanoVDB.cu
```diff
@@ -873,26 +891,14 @@ maybeSaveStandardNanovdbGrid(const std::string &path,
...
-
-    if (isCuda) {
-        c10::cuda::CUDAGuard deviceGuard(gridBatchData.device());
-        cudaCheck(cudaMemcpy(writeHead, readHead, sourceGridByteSize, cudaMemcpyDeviceToHost));
-    } else {
-        std::memcpy(writeHead, readHead, sourceGridByteSize);
+    // The batch's *logical* grids, compacted and brought to the host: exactly the grids a sliced
+    // batch selects, each header already renumbered, so the patch loop below can index the
+    // handle by batch position.
+    GridStorage hostStorage = ops::contiguousGridStorage(gridBatchData);
```
swahtz: `batchSize() == 0` now throws from `nanovdb::io::writeGrid` ("Segment contains no grids") after the file has been opened and truncated. `contiguousGridStorage` over zero logical grids yields empty storage, `to(kCPU)` returns `GridStorage()`, `Segment::add` adds nothing, `Segment::write` throws (pinned `io/IO.h:386-389`). The base wrote the one-grid sentinel and `fvdb.load` read it back as one empty grid.

```python
gb = fvdb.GridBatch(device="cuda")
gb.save_nanovdb("e.nvdb")   # base: ok; head: RuntimeError, e.nvdb left at 0 bytes
```

`toNVDB()` without data silently returns an empty handle on the same path. Either `TORCH_CHECK_VALUE(batchSize() > 0, ...)` at the three entry points or an explicit sentinel, plus a test. `test_io.py` has no empty-batch case.

src/fvdb/detail/ops/CloneGrid.cu
```diff
@@ -15,21 +15,19 @@ cloneGrid(const GridBatchData &grid, const torch::Device &device, bool blocking)
...
-        // Requested a different target device: the handle is now contiguous, so a whole-handle copy
-        // to `device` moves exactly the selected grids.
-        TorchDeviceBuffer guide(0, device);
-        clonedHdl = clonedHdl.copy<TorchDeviceBuffer>(guide);
+    GridStorage cloned = contiguousGridStorage(grid);
+    if (cloned.device() != device) {
+        // A different target device: the storage is now contiguous, so moving it moves exactly
+        // the selected grids. Complete on return for a host destination; for a device destination
+        // ordered on its current torch stream, which is where the batch's later work runs.
+        cloned = cloned.to(device);
```
swahtz, line 30: Non-empty moves now go through `GridStorage::to`, whose `checkStorageDevice` (`GridStorage.cu:33`) rejects an index-less CUDA device. The base `TorchDeviceBuffer::copy(guide)` path accepted `cuda:-1` as the current device.

```python
gb = fvdb.GridBatch.from_dense(1, [4, 4, 4], device="cuda:0")
gb.to(torch.device("cuda"))
# RuntimeError: GridStorage: CUDA device must carry an index, got cuda
```

`grid_batch.py:1574` only normalizes strings, so a `torch.device` object arrives here as `cuda:-1`. `cloned.device() != device` is then true even for a `cuda:0` batch and `to(cuda:-1)` throws. Same for `Grid.to` and C++ `fvdb::load(path, torch::kCUDA)` (`LoadNanovdb.cpp:626-629`). #770 says the resource defaults to the current device, so normalizing here or in `GridStorage::to` matches the design:

```cpp
if (device.is_cuda() && !device.has_index()) device = {torch::kCUDA, c10::cuda::current_device()};
```

src/fvdb/GridBatchData.cu
```diff
@@ -8,42 +8,43 @@
...
+    return *mStorage;
 }
 
 const c10::Device
 GridBatchData::device() const {
-    return mGridHdl->buffer().device();
+    return mStorage->device();
 }
 
 nanovdb::OnIndexGrid *
```
swahtz, line 26: When the storage is on the other side, `<side>Bytes()` is null and these accessors return `static_cast<uint8_t*>(nullptr) + cumBytesAt(bi)`, a non-null garbage pointer. Views keep the physical `mCumBytes` (`IndexGrid.cu:53-55`), so even `bi == 0` on a sliced view is non-zero. That deadens the `TORCH_CHECK(grid != nullptr)` guards at the four sites this PR moved off the null-safe `deviceGrid<T>(i)`: `BuildMergedGrids.cu:46-48`, `Inject.cu:237-238`, `ReinitializeSdf.cu:399`, `PredGatherIGemm.cu:1075-1076`.

Three of those have an earlier device-equality check. `predGatherIGemmSparseConv` (~L1120-1170, bound as `pred_gather_igemm_conv`) checks only that features and weights are CUDA, so `pred_gather_igemm_conv(features_cuda, weights_cuda, cpu_grid[1:2], cpu_out_grid[1:2], ...)` used to raise "Failed to get device input grid" and now hands a bogus device address to the CUTLASS kernel.

Returning `nullptr` when the base pointer is null makes every existing guard live again. Step 5 puts 26 more builder sites on these accessors, so worth settling now. A grid/feature device check in `predGatherIGemmSparseConv` would be belt and braces.

src/fvdb/detail/ops/MakeContiguous.cu
```diff
@@ -35,36 +28,122 @@ copyAndFixGrid(nanovdb::GridData *dst,
...
+        if (mDevice.is_cuda()) {
+            fvdb::detail::orderStreamAfter(mStream, mDevice, mSrcStream, mDevice);
+        } else if (mDevice.is_privateuseone()) {
+            fvdb::detail::synchronizeStream(mSrcStream, mDevice);
+        }
+    }
+    void
+    finish() {
+        if (mDevice.is_cuda()) {
+            fvdb::detail::orderStreamAfter(mSrcStream, mDevice, mStream, mDevice);
```
swahtz, line 75: `finish()` is a manual call, so a throw inside the copy loop (a non-sticky `cudaErrorInvalidValue` from `cudaMemcpyAsync`, or `finish()`'s own `cudaEventCreate` failing) leaves the source's retained stream unordered after copies were already enqueued on another stream. The caller releases the source, torch returns its block to stream A's pool, and A's next allocation can overwrite it under B's in-flight reads.

Low realism (distinct streams plus a non-sticky error plus immediate release), and the destination side is safe since both allocators free on the copy stream. But the struct exists to prevent exactly this, so a destructor with `try`/`catch` around `orderStreamAfter` (it can throw via `C10_CUDA_CHECK`) would close it at no cost.

src/tests/LoadNanovdbTest.cpp
```diff
@@ -103,7 +103,8 @@ makeTensorGridBlindDataHandle(const fvdb::GridBatchData &gridBatchData,
 
     torch::Tensor contiguousData = sourceData.contiguous();
 
-    const nanovdb::GridData *sourceGridData = gridBatchData.nanoGridHandle().gridData(0);
+    const nanovdb::GridData *sourceGridData =
+        gridBatchData.gridStorage().hostGridAt<nanovdb::ValueOnIndex>(0)->data();
```
swahtz: `hostGridAt<ValueOnIndex>(0)->data()` dereferences before the `TORCH_CHECK(sourceGridData != nullptr)` on the next line. The replaced `nanoGridHandle().gridData(0)` was null-safe, so the check is now dead, and a member call on null is UB the optimizer may fold. Unreachable in today's fixture (L94-95 assert CPU and batch size 1), but:

```cpp
const auto *g = gridBatchData.hostGridPtrAt(0);
TORCH_CHECK(g != nullptr, ...);
sourceGridData = g->data();
```

also avoids spelling a physical `GridStorage` index on a `GridBatchData`.

src/fvdb/detail/io/SaveNanoVDB.cu
harrism: Both now order the copy stream after `storage.stream()` before the first read, in 2c16b8d: the assembler right before its copy loop (under the batch's device guard) and the typed export right after it takes `deviceBytes()`. Same helper as the other storage reads.

src/fvdb/detail/io/SaveNanoVDB.cu
harrism: Fixed in 2c16b8d and tested (`test_save_and_load_empty_gridbatch`, CPU and CUDA). The index-grid export of an empty batch writes the storage's one voxel-less sentinel grid, as the base did, and loads back as one empty grid; exporting an empty batch *with* data is a `ValueError` at both funnels rather than a truncated file. That test also exposed that `fvdb/functional/_io.py` passed `None` for an empty `JaggedTensor` (falsy), so the data path was never reached; fixed to test for `None`. While here: a contiguous batch now moves in one pass (`gridStorage().to(kCPU)`) instead of compacting first, per your #770 note, so `getIndexGrid` on CUDA is back to one D2H copy.

src/fvdb/detail/ops/CloneGrid.cu
harrism: Fixed in 2c16b8d: `GridStorage::resolveDevice` maps an index-less CUDA device to the current one, applied at every `GridStorage` entry point that takes a device and at the top of `cloneGrid` before its comparison, so `gb.to(torch.device("cuda"))` on a `cuda:0` batch is a same-device clone again. Test: `test_to_indexless_cuda_device`. `cloneGrid` also takes the one-pass path for contiguous batches now.

src/fvdb/GridBatchData.cu
harrism: Agreed, and worth settling before step 5 puts 26 more sites on these. Both accessors return `nullptr` when the storage has no bytes on that side (2c16b8d), so the four moved null checks are live again; `predGatherIGemmSparseConv` additionally checks both grids' devices against the features' up front, so the mismatch is a clear error rather than a device address handed to the kernel.

src/fvdb/detail/ops/MakeContiguous.cu
harrism: Taken in 2c16b8d: `SourceStreamOrder` now orders in its destructor when `finish()` was not reached, swallowing a secondary CUDA error so the original propagates. Kept the explicit `finish()` for the normal path so a failure of the ordering call itself is reported rather than silenced; if you would rather it were destructor-only, say so and I will flip it.

src/tests/LoadNanovdbTest.cpp
harrism: Taken in 2c16b8d: `hostGridPtrAt(0)` with the null check before the dereference, and no physical `GridStorage` index spelled on a `GridBatchData`.

harrism:
The three SaveNanoVDB items from the #770 thread are in 2c16b8d as well: `patchGridWithBlindShape` goes through `normalizeStandaloneGridHeader`, which now takes the blind layout, disables the checksum the header writes invalidate, and clears `HasLongGridName` when no blind record remains (`indexToGridHost` stops computing a checksum that was always invalidated afterwards); the CUDA typed export converts an empty member on the host instead of through upstream `indexToGrid`'s unguarded launch, with `test_save_and_load_with_empty_grid_in_batch` covering it on CPU and CUDA. Also folded in from your #770 step-5 note where it touched code this PR rewrote: contiguous batches move in one pass in `getIndexGrid` and `cloneGrid`, `serializeGrid` uses `totalBytes()`, and the unreachable `has_index` check is gone. Re-verified: `-Werror` clean, ctest 50/50, full `tests/unit` 2998 passed, no-caching rerun clean, codex clean after two passes (its one finding was the wrapper's falsy-empty-tensor bug, fixed).

src/fvdb/detail/ops/MakeContiguous.cu
swahtz: No I think that's fine

swahtz (approved):
Great, thanks for addressing my notes, looks good to go.

PR #788, opened 2026-09-16
Builders produce GridStorage; one grid assembler (#770 step 5)

harrism:
## Summary

Step 5 of #770: the grid builders produce `GridStorage` directly, and every TorchDeviceBuffer seam that steps 3 and 4 left in place is removed. Two commits.

**Commit 1: one grid assembler.** The copy-and-fix loop (bracket the source stream, copy each grid, stamp its new index/count, disable its checksum, adopt the layout) was spelled three times, in `contiguousGridStorage`, `concatenateGrids` and `mergeDeviceGridHandles`, and the three disagreed on PrivateUse1 ordering. `assembleGridStorage(sources, device, stream)` (`GridAssembly.cu`) is now the one spelling. On it: `contiguousGridStorage(input, device)` compacts onto a destination device in one pass (so `cloneGrid` across devices and `getIndexGrid` of a sliced batch are single copies), `cloneGridStorageAt`, `concatenateGrids`, and `serializeGrid`, which no longer builds a throwaway `GridBatchData` to read its bytes. Also advances the NanoVDB pin to 0ab0a81e (openvdb #2322); fvdb builds unchanged against it.

**Commit 2: the builders.** Each builder's per-item grids are built straight into single-space `DeviceGridBuffer` storage through `GridStorage::deviceProto(device, stream)`, which carries the device and the storage stream into every allocation the NanoVDB builder makes, or into `HostBuffer` storage on the CPU, and are laid end to end by `mergeGridStorages`. `BatchedTopologyBuilder` allocates its output as a `DeviceGridBuffer` and the final pass's buffer becomes the batch's storage, adopting the host-computed layout instead of parsing it back. Deleted: the zero-size "guide" buffers, `createEmptyGridHandle`, `contiguousGridHandle`, `cloneGridHandleAt`, `nanovdb::cuda::mergeGridHandles` usage, `adoptLegacyGridHandle` / `LegacyGridHandle.h`, the `makeGridBatchData(GridHandle<TorchDeviceBuffer>&&)` overload and `mergeDeviceGridHandles`. No builder output is copied into `GridStorage` any more: the extra copy per build that #787 introduced is gone.

## Stream contract

CUDA builders run their kernels and allocate their output on the device's current torch stream (`storageStream`), which the storage retains; previously `voxelsToGrid` ran on the legacy default stream. PrivateUse1 builders over `DistributedPointsToGrid` retarget the result's retained stream onto the storage stream before the `DeviceMesh` whose stream `getHandle` allocated on is destroyed, so no buffer names a dead stream (safe because the PrivateUse1 allocator ignores the stream; noted in `TorchDeviceResource.h`). For a device destination the assembler orders `stream` after each distinct source stream and each CUDA source stream after `stream`, so a source may be released as soon as it returns; host and PrivateUse1 sources or destinations synchronize.

## Latent bugs exposed and fixed

Serializing a CUDA batch's own metadata records exposed that `PopulateGridMetadata`'s CUDA path never set the `version` fields: the device records are raw allocations, the kernel writes only the fields it computes, and the host records are overwritten from the device. Pickling a CUDA grid failed once the serializer stopped cloning to CPU first. The device records are now seeded from the constructed host ones. The readback after that kernel was also a synchronous `cudaMemcpy` on the legacy default stream, which torch's non-blocking streams do not synchronize with; it is now stream-ordered.

## Policy decisions

Checksums: NanoVDB's builders default to none, assembled storage disables them, and nothing in fvdb reads them, so the per-item `setChecksum(CheckMode::Default)` calls, which only ever survived for single-grid batches, are dropped. `TorchDeviceBuffer.cpp` becomes `.cu`: the dual-space `GridHandle` constructor is only defined for CUDA translation units and the builders were the last CUDA units instantiating it. `TorchDeviceBuffer` itself is now unused by the library; step 6 deprecates it.

## Verification

`./build.sh install gtests`: clean. ctest 50/50 (the merge gtest now covers `mergeGridStorages`: device parts with an empty part mixed in, host parts onto the host, the single-part pass-through, and the all-empty case keeping device and stream). Full `tests/unit`: 3016 passed / 33 skipped (the known fp16 flake deselected). `PYTORCH_NO_CUDA_MEMORY_CACHING=1` rerun of `test_basic_ops`, `test_basic_ops_single`, `test_io`, `test_sliced_batch`, `test_conv_default`, `test_conv_transpose_default`: 1124 passed. clang-format clean.

Codex-reviewed before opening: four passes on the assembler (six findings, all folded in and listed in the commit message), two on the builders (clean), plus a self-review audit whose findings (include self-containment, comment drift, naming) are in the second commit.


src/fvdb/detail/ops/SerializeGrid.cu
```diff
@@ -55,11 +55,11 @@ serializeGrid(const GridBatchData &grid) {
     memcpy(retPtr, &header, sizeof(V01Header));
     retPtr += sizeof(V01Header);
 
-    memcpy(retPtr, &self->mBatchMetadata, sizeof(GridBatchData::GridBatchMetadata));
+    memcpy(retPtr, &grid.mBatchMetadata, sizeof(GridBatchData::GridBatchMetadata));
     retPtr += sizeof(GridBatchData::GridBatchMetadata);
 
     if (numGrids > 0) {
-        memcpy(retPtr, self->mHostGridMetadata, numGrids * sizeof(GridBatchData::GridMetadata));
+        memcpy(retPtr, grid.mHostGridMetadata, numGrids * sizeof(GridBatchData::GridMetadata));
```
swahtz, line 64: `serializeGrid` now writes the batch's own per-grid records, which is right for CPU and CUDA but breaks PrivateUse1. There the records come from `allocateUnifiedMemoryGridMetadata` (`GridBatchDataFactory.cu:70-75`), a `raw_allocate` with no constructor run, and `hostMeta = deviceMeta` aliases them (`:163`). `populateGridMetadataKernel` writes every field but `version`, so the record's `version` is whatever the allocator returned. Before this PR the `cloneGrid(grid, kCPU)` step produced constructed host records (`version = 1`); now `deserializeGrid` fails its `TORCH_CHECK(gridMetadata[i].version == 1, ...)` at `:103`, or passes by luck. Same for `IndexGrid` views of a PrivateUse1 batch, which copy the parent's records.

The `PopulateGridMetadata.cu:156-170` seeding fixes this for CUDA only, and it depends on every caller handing in constructed host records. Stamping `version = 1` in the kernel (one store per record, next to `setTransform` at `:52`, and one for the batch record) fixes CPU, CUDA and PrivateUse1 uniformly and lets the two seeding `cudaMemcpyAsync` calls go.

src/fvdb/detail/ops/SerializeGrid.cu
```diff
@@ -26,21 +28,19 @@ torch::Tensor
...
-        // and a view shares storage holding more grids than it selects.
-        cpuClone = cloneGrid(grid, torch::kCPU, true);
-        self     = cpuClone.get();
+    // The bytes to write are the batch's logical grids on the host. A contiguous CPU batch has
+    // them already; anything else is compacted straight onto the host in one pass. The per-grid
+    // metadata written below is the batch's own: only its voxel sizes and origins are read back,
+    // and deserialization recomputes the rest from the grids.
+    std::optional<GridStorage> compacted;
+    if (!(grid.device().is_cpu() && grid.isContiguous()) && grid.batchSize() > 0) {
+        compacted = contiguousGridStorage(grid, torch::kCPU);
```
swahtz: A contiguous CUDA batch still takes the per-grid assembler here: B D2H copies, B header rewrites (with the values the headers already hold), and a source-stream sync. `getIndexGrid` (`SaveNanoVDB.cu`) and `cloneGrid` (`CloneGrid.cu`) both split on `isContiguous()` and take `gridStorage().to(kCPU)` for the contiguous case since #787; reusing that split makes pickling a contiguous CUDA batch one copy:

```cpp
if (grid.batchSize() > 0 && !(grid.device().is_cpu() && grid.isContiguous())) {
    compacted = grid.isContiguous() ? grid.gridStorage().to(torch::kCPU)
                                    : contiguousGridStorage(grid, torch::kCPU);
}
```

src/CMakeLists.txt
```diff
@@ -52,7 +52,7 @@ set(FVDB_CPP_FILES
     fvdb/detail/utils/cuda/LocalGradient.cpp
     fvdb/detail/utils/cuda/Prefetch.cpp
     fvdb/detail/ops/jagged/JaggedReductions.cpp
-    fvdb/TorchDeviceBuffer.cpp
+    fvdb/TorchDeviceBuffer.cu
```
swahtz: `TorchDeviceBuffer.cu` is left in `FVDB_CPP_FILES` after the rename. CMake compiles it with nvcc by extension so nothing breaks, but the two lists are where the language-specific options below (L188-241) are reasoned about by membership; it belongs in `FVDB_CU_FILES` next to `GridAssembly.cu`.

swahtz (commented):
Reviewed against the #770 thread as well as the diff. Three inline comments below; the first is a correctness regression on the PrivateUse1 path, the other two are small. Findings that read as step-5/step-6 decisions or as performance follow-ups (the topology builder's adopted layout, replicated and empty grid copies, the PrivateUse1 IndexGrid metadata leak) are on #770 rather than here.

The CI failures on the first attempt are infrastructure: g6 had no capacity in any zone, the workflow fell back to g5 (A10G, SM 8.6), and the wheel was built for `8.9+PTX` only, so every kernel fails to load, including `DispatchScanLibCudaTest`, which this PR does not touch. Separate workflow fix (add 8.6 to the test build's arch list, or drop the g5 fallback).

src/fvdb/detail/ops/SerializeGrid.cu
harrism: Fixed as suggested. `populateGridMetadataKernel` now stamps `version` on the batch record and on every per-grid record (both loop body and the last-grid tail), so CPU, CUDA and PrivateUse1 get it from the one place that fills the records; the two seeding `cudaMemcpyAsync` calls are gone (the stream-ordered D2H readback stays). The literal moved into `GridMetadata::kVersion` / `GridBatchMetadata::kVersion`, which the default member initializers and `deserializeGrid`'s checks now use too. (4664a42)

src/fvdb/detail/ops/SerializeGrid.cu
harrism: Done: a contiguous device batch takes `gridStorage().to(torch::kCPU)`, the non-contiguous case keeps the one-pass compaction. (4664a42)

src/CMakeLists.txt
harrism: Moved to `FVDB_CU_FILES`. (4664a42)

swahtz (approved):
Looks good, thanks for addressing the comments

PR #789, opened 2026-09-17
CI: build the conda test wheel for the arch list in versions.json

harrism:
## Summary

#763 switched the pip CI workflows to build PR wheels for the `arch_list_pr` in `.github/versions.json` (now `8.6+PTX`) and gave the GPU test runners a `g5.xlarge` (A10G, SM 8.6) fallback pool. The conda test workflow (`tests.yml`) got the fallback runner but kept its hardcoded `--cuda-arch-list '8.9+PTX'`.

The first time the fallback fired, on #788's run, all three conda test jobs failed with `cudaErrorNoKernelImageForDevice` on every CUDA kernel, starting with `src/dispatch/examples/relu.cu`: sm_89 code cannot run on an sm_86 device, and compute_89 PTX cannot be compiled down to it. The test job's own log shows the mismatch: `Detected CUDA architectures: 8.6+PTX` against a wheel built for 8.9.

This makes `tests.yml` read the same versions output `cu130.yml`, `cu132.yml`, `cu130-nightly.yml` and `weekly-compat.yml` already use. No other workflow hardcodes an arch list.

## Verification

Workflow-only change; the `fvdb-build` job already lists `versions` in `needs`, and the `env:` block mirrors `cu130.yml` line 160. This PR's own conda test run exercises the new line.


swahtz (approved):
Woops!  Thanks for the fix

PR #790, opened 2026-09-17
Grid storage is private to GridBatchData; TorchDeviceBuffer is deprecated (#770 step 6)

harrism:
## Summary

Step 6 of #770, the last: grid storage becomes private to `GridBatchData`, and `TorchDeviceBuffer` is deprecated.

**Storage is private.** `mStorage` and `gridStorage()` are no longer reachable from outside the class. Readers use the logical accessors `hostGridPtrAt` / `deviceGridPtrAt` (correct for sliced views, whose storage holds more grids than they select), order themselves after the new `storageStream()`, and copy a whole batch with `detail::ops::contiguousGridStorage`, the one friend. That function now owns the "is it already contiguous" question: a contiguous or empty batch's storage is copied whole, anything else is compacted in one pass. `SerializeGrid`, `CloneGrid` and `SaveNanoVDB::getIndexGrid` call it unconditionally; the blind-data assembler, the host typed export, `ConcatenateGrids` and `MakeContiguous` read grids through the accessors; the CUDA typed export of a CPU batch stages its logical grids onto the device in one pass instead of uploading the whole storage; `IndexGrid` builds views through a new constructor that shares the source batch's storage.

**Deprecation.** `TorchDeviceBuffer`, and `TorchStorageResource` with it (moved into the shim's header, its only user), are `[[deprecated]]` and unused by the library. They stay for one release for downstream code that took them from the public headers (#632) and go with the NanoVDB pin bump past upstream's removal of dual-space buffers (openvdb #2232). The shim's own translation unit and test silence the warning; every other use of the header now fails `-Werror`.

**Grep gate.** A new codestyle job greps `src/` for `TorchDeviceBuffer`, `TorchStorageResource`, and NanoVDB's dual-space buffer names and include paths, allowing only the shim's files and their build entries. `PadGrid::getHandle` loses its `DualDeviceBuffer` default template argument, which nothing used.

**CHANGES.md** gets the Breaking entry for the whole migration (#773, #786, #787, #788, this PR).

## Behaviour

No functional change is intended beyond the CUDA typed-export staging path, which now copies only the logical grids (one pass, `TorchDeviceResource`) instead of the whole storage buffer. `contiguousGridStorage`'s hoisted split gives `cloneGrid`, `serializeGrid` and `getIndexGrid` exactly the branches they had inline.

## Verification

`./build.sh install gtests`: clean under `-Werror`. ctest 50/50. Full `tests/unit`: 3016 passed / 33 skipped (known fp16 flake deselected). `PYTORCH_NO_CUDA_MEMORY_CACHING=1` rerun of the memory-sensitive files: 1059 passed. Explicit probe: pickling an empty slice and a one-empty-grid batch round-trips on CPU and CUDA (codex caught the empty-CPU case throwing in the first draft; fixed before commit). The grep gate returns clean on this tree. clang-format clean.

Codex-reviewed before opening (one finding, folded in) plus a self-review audit (comment drift, a missing include and a lost `static_assert` in the moved shim resource, changelog accuracy; all folded in).


swahtz (commented):
Reviewed against the #770 thread as well as the diff. No functional regression found: the hoisted `isContiguous() || batchSize() == 0` split inside `contiguousGridStorage` gives `cloneGrid`, `serializeGrid` and `getIndexGrid` the branches they had inline, an empty batch always owns exactly its sentinel storage (zero-length views short-circuit to a fresh batch), the grep gate returns clean on this tree, and the CHANGES.md entry matches what the five PRs did. Seven small items below, all inline except one.

Not inline because the file is outside this diff: `BuildPaddedGrid.cu:347-348` still says "unlike `gridStorage().deviceGridAt(i)` which indexes physically", steering readers to the accessor this PR makes private. The equivalent sentence in `GridBatchData.h` was rewritten; this one was missed. Reword to "the storage's own `deviceGridAt` indexes physically".

src/fvdb/detail/io/SaveNanoVDB.cu
```diff
@@ -239,26 +234,26 @@ assembleOnIndexBlindBuffer(const GridBatchData &gridBatchData,
...
-            C10_CUDA_CHECK(cudaMemcpyAsync((void *)gridDst[batchIdx],
-                                           (const void *)readHead,
-                                           gridBytes,
-                                           cudaMemcpyDeviceToHost,
-                                           stream.stream()));
+            C10_CUDA_CHECK(cudaMemcpyAsync(
+                gridDst[batchIdx], readHead, gridBytes, cudaMemcpyDeviceToHost, stream.stream()));
         } else {
-            std::memcpy((void *)gridDst[batchIdx], (const void *)readHead, gridBytes);
+            std::memcpy(gridDst[batchIdx], readHead, gridBytes);
```
swahtz, line 259: For a PrivateUse1 batch `isCuda` is false, so the ordering block at L232-238 is skipped and this `memcpy` reads the unified pointer (`hostBytes()` returns it for PrivateUse1, per `GridStorage.h:104-105`) without ordering after `storageStream()`. Same at L572, where `fvdbToNanovdbGridWithValuesHost` walks `hostGridPtrAt(bi)` on the host.

Every PrivateUse1 producer synchronizes before handing storage out (`assembleGridStorage` for a PrivateUse1 destination, `GridStorage::to`, the distributed builder's `getHandle`), so this cannot race today. It is a contract the reader does not enforce, where `GridStorage::to` and the assembler both do before touching a PrivateUse1 pointer from the host. Pre-existing, and both readers were rewritten here; one `if (gridBatchData.device().is_privateuseone()) synchronizeStream(gridBatchData.storageStream(), gridBatchData.device());` before the first host read in each closes it.

src/fvdb/detail/io/SaveNanoVDB.cu
```diff
@@ -576,8 +569,7 @@ fvdbToNanovdbGridWithValuesHost(const GridBatchData &gridBatchData,
         TORCH_CHECK_VALUE(
             numVoxelsBi >= 0, "Invalid number of voxels at grid index ", bi, ": ", numVoxelsBi);
 
-        const auto *hSrcGrid = reinterpret_cast<const nanovdb::NanoGrid<nanovdb::ValueOnIndex> *>(
-            hSrcBufferStart + gridBatchData.cumBytesAt(bi));
+        const nanovdb::NanoGrid<nanovdb::ValueOnIndex> *hSrcGrid = gridBatchData.hostGridPtrAt(bi);
```
swahtz, line 581: Same as the assembler comment above: a PrivateUse1 batch reaches this host path (L646 routes every non-CUDA batch here) and reads unified memory through `hostGridPtrAt` with no ordering after `storageStream()`.

src/fvdb/detail/io/SaveNanoVDB.cu
```diff
@@ -685,34 +677,31 @@ fvdbToNanovdbGridWithValues(const GridBatchData &gridBatchData,
...
                                          : torch::Device(torch::kCUDA, c10::cuda::current_device());
     c10::cuda::CUDAGuard deviceGuard(cudaDevice);
     const at::cuda::CUDAStream stream = at::cuda::getCurrentCUDAStream(cudaDevice.index());
 
-    const uint8_t *dSrcBufferStart = nullptr;
+    // The source grids as device pointers: a CUDA batch's own storage, with this stream ordered
+    // after its writers; a CPU batch's logical grids brought onto this device in one pass
+    // (contiguousGridStorage runs on, and the staged storage retains, this device's current
+    // stream, which is `stream`).
+    std::optional<GridStorage> staged;
```
swahtz: L646 returns through the host path for every non-CUDA batch, so this staging branch is unreachable, and the comment at L680-682 says as much. The PR rewrote it into `std::optional<GridStorage> staged`, the `cudaDevice` fallback at L684-686 and the `dSrcGridAt` lambda rather than deleting it. Deleting the `else` lets `cudaDevice` become `gridDevice`, removes `staged` and the lambda, and the loop reads `gridBatchData.deviceGridPtrAt(bi)` directly. Nothing in the 3016-test run can enter it, so it only accrues maintenance.

src/fvdb/detail/ops/ConcatenateGrids.cu
```diff
@@ -29,24 +29,23 @@ concatenateGrids(const std::vector<c10::intrusive_ptr<GridBatchData>> &elements)
         if (element->batchSize() == 0) {
             continue;
         }
-        const GridStorage &storage = element->gridStorage();
-        const auto *base = static_cast<const uint8_t *>(device.is_cpu() ? storage.hostBytes()
-                                                                        : storage.deviceBytes());
-        GridSpanSource source{device, storage.stream(), {}};
+        GridSpanSource source{device, element->storageStream(), {}};
```
swahtz: This loop (L32-41) re-implements `logicalGridSpans()` from `MakeContiguous.cu` line for line: same `GridSpanSource{device, storageStream(), {}}`, same `is_cpu() ? hostGridPtrAt : deviceGridPtrAt` selection, same `GridSpan{ptr, numBytesAt(j), OnIndex}` push. Two copies of the "logical grids as spans" rule in two files means the next change to one (a PrivateUse1 side selection, a grid-type parameter) misses the other. Move `logicalGridSpans(const GridBatchData&, int64_t first, int64_t count)` out of the anonymous namespace into `MakeContiguous.h` and call it here.

src/fvdb/GridBatchData.h
```diff
@@ -79,21 +85,33 @@ struct GridBatchData : public torch::CustomClassHolder {
...
 
+  private:
+    // The grids' bytes and the metadata locating them; shared with views over this batch. Grid
+    // indices on it are *physical*: a view over a subset of the batch maps logical items onto them
+    // by byte offset (cumBytesAt), which is what deviceGridPtrAt / hostGridPtrAt do. Nothing
+    // outside this class reads the storage directly; ops go through those accessors,
+    // storageStream(), and detail::ops::contiguousGridStorage, which copies a whole batch.
+    std::shared_ptr<GridStorage> mStorage;
+    const GridStorage &gridStorage() const;
+    friend GridStorage detail::ops::contiguousGridStorage(const GridBatchData &input,
```
swahtz: Making storage private through a friend that lives in the ops layer puts an ops symbol in the core header: `GridBatchData.h:25-28` forward-declares `detail::ops::contiguousGridStorage`, and its signature is now split across two headers (no default argument here, `= std::nullopt` in `MakeContiguous.h`). Legal, but a TU that includes only this header sees a different callable. The private `gridStorage()` accessor exists solely to serve that one friend.

Alternative, if you want it: one public primitive on the class, e.g.

```cpp
/// The storage moved to @p device on @p stream, whole; the caller decides whether that is the batch.
GridStorage copyStorage(const torch::Device &device, cudaStream_t stream) const { return mStorage->to(device, stream); }
```

keeps `mStorage` private, removes the friend, the forward declaration and `gridStorage()`, and lets `contiguousGridStorage` stay an ordinary function that calls `copyStorage` for the contiguous case and the accessors otherwise. A design preference, not a defect; fine to leave as is.

src/fvdb/detail/ops/CloneGrid.cu
```diff
@@ -17,14 +17,10 @@ cloneGrid(const GridBatchData &grid, const torch::Device &requested, bool blocki
...
-    // voxelSizes/gridCount mismatch) for an indexed batch, where gridCount() > batchSize().
-    // One pass either way. A contiguous batch's storage already holds exactly the logical grids,
-    // so GridStorage::to (a deep copy on the same device, a move otherwise) is the clone; a view
-    // is compacted straight onto the target device.
-    GridStorage cloned =
-        grid.isContiguous() ? grid.gridStorage().to(device) : contiguousGridStorage(grid, device);
+    // The selected grids as fresh storage on the target device, one pass either way: a
+    // contiguous batch's storage is copied whole, a sliced or non-contiguous view (whose storage
+    // holds more physical grids than the batch selects) is compacted straight onto the device.
+    GridStorage cloned = contiguousGridStorage(grid, device);
```
swahtz, line 23: `blocking` (defaulted to `false` in `CloneGrid.h`) is never read in the body, so a C++ caller asking for a non-blocking clone gets the synchronous `contiguousGridStorage` → `GridStorage::to` path and no error. Pre-existing, but this PR rewrote the only line that could have honored it. Either thread it through or drop the parameter and its default.

src/fvdb/TorchDeviceBuffer.cu
```diff
@@ -1,6 +1,10 @@
 // Copyright Contributors to the OpenVDB Project
 // SPDX-License-Identifier: Apache-2.0
 //
+// This translation unit is the deprecated shim itself (or its test); it is the one place the
+// deprecation must not fire.
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
swahtz, line 6: The GCC pragma governs the host-compiler pass; nvcc's own front end emits separate deprecation diagnostics (`#1215-D` / `#1444-D`) that `-Werror=all-warnings` (`src/CMakeLists.txt:209`, `src/tests/CMakeLists.txt:90`) promotes to errors. CI builds pass under `-Werror`, so the current toolkit honors the GCC pragma in EDG mode; a toolkit that does not would fail both shim TUs. `#pragma nv_diag_suppress 1215 1444` under `#ifdef __CUDACC__` next to this line (here and in `TorchDeviceBufferTest.cu:6`) makes the suppression independent of that behavior.

src/fvdb/detail/io/SaveNanoVDB.cu
harrism: Added: a PrivateUse1 batch now waits on its storage stream (`synchronizeStream(storageStream(), device)`) before the first host read, in the `else` of the ordering block. (28670d5)

src/fvdb/detail/io/SaveNanoVDB.cu
harrism: Same fix here, ahead of the per-grid loop. (28670d5)

src/fvdb/detail/io/SaveNanoVDB.cu
harrism: Deleted. The function now checks `gridDevice.is_cuda()`, guards and streams on the batch's own device, orders after `storageStream()`, and the loop reads `deviceGridPtrAt(bi)` directly; `staged`, the lambda and `cudaDevice` are gone. (28670d5)

src/fvdb/detail/ops/ConcatenateGrids.cu
harrism: Moved `logicalGridSpans(batch, first, count)` next to `GridSpanSource` in `DeviceGridHandleUtils.cuh` (MakeContiguous.h is included from a .cpp, so it cannot carry the .cuh type); `MakeContiguous.cu` and `ConcatenateGrids.cu` both call it. (28670d5)

src/fvdb/GridBatchData.h
harrism: Took the alternative: `copyStorage(device, stream)` is the one public primitive, and the friend, the forward declaration, the private `gridStorage()` and the `<optional>` include are gone; `contiguousGridStorage` is an ordinary function calling it for the contiguous/empty case. (28670d5)

src/fvdb/detail/ops/CloneGrid.cu
harrism: Dropped the parameter (no caller passed it: the Python binding, `BuildDilatedGrid`, `LoadNanovdb` all call the two-argument form). (28670d5)

src/fvdb/TorchDeviceBuffer.cu
harrism: Added `#pragma nv_diag_suppress 1215 1444` under `#ifdef __CUDACC__` next to the GCC pragma in both TUs. (28670d5)

harrism:
The one item not inline (`BuildPaddedGrid.cu` still steering readers to `gridStorage().deviceGridAt`) is reworded in 28670d5 as well.

swahtz (approved):
Looking good! 🚀

